PR #30, opened 2022-02-19
Import sage-angular lib

tschaffter:
## Changelogs

- [x] Add shared library `assets` that includes images for apps and other libs
- [x] Add shared library `style` that includes shared style variables
- [x] Add SCSS theme support to the web-app 
- [x] Add web library `ui` that includes UI components used by the web-app
- [x] Add web library `data-access` that includes data access components/services used by the web-app


tschaffter:
## Importing styles and theme from Angular library

This is currently an issue. Some of the links below may help.

- https://github.com/nrwl/nx/issues/1542#issuecomment-952984016

PR #52, opened 2022-03-02
Use . ./dev-env.sh to support both bash and zsh

### LINKED ISSUE #50: dot command does not work for zsh shell
Check you shell:
```
echo $SHELL
```
If the shell is `/bin/zsh`, running `. dev-env.sh` will not work. It will raise below error:
```
.: no such file or directory: dev-env.sh
```

Two solutions:
1. using `source` instead
2. according to this [post](https://github.com/ohmyzsh/ohmyzsh/issues/8100#issuecomment-524344725), (adding `slash`) run `. ./dev-env.sh` can solve this issue as well

tschaffter:
Fixes #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Clone this repository.
 
 Source `dev-env.sh`.
 
-    . dev-env.sh
+    . ./dev-env.sh
 
 Prepare the development environment.
 


rrchai (approved):
Nice!

PR #56, opened 2022-03-05
Add VS Code dev container

tschaffter:
https://code.visualstudio.com/docs/remote/containers

## 2022-03-05

I created a dev container that includes Python 3.9, Node 16, yarn and docker. VS Code provide some support to run docker inside the dev Docker container, which would be needed if we want for example to run `yarn docker` to build all the docker images defined in the Nx workspace inside the dev container.

Getting Docker to run inside a docker container is tricky because of the Docker socket permission. This is well explained [here](https://github.com/microsoft/vscode-dev-containers/tree/main/containers/docker-from-docker).


thomasyu888:
I would avoid installing a docker daemon inside your container. At least it was a security concern previously: https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/

Although what we do in challenges is we mount the host docker daemon.

tschaffter:
@thomasyu888 We are on the same wavelength. Have you ever installed the Docker CLI without the engine? I checked if it was possible but stopped after finding https://github.com/docker/cli/issues/2281.

`Docker-in-docker` is actually a feature offered by VS Code for dev containers (see `devcontainer.json`).

> Create child containers inside a container, independent from the host's docker instance. Installs Docker extension in the container along with needed CLIs.

thomasyu888:
Interesting, is the use case to support a mono repo with multiple docker images? I have not tried, but thanks for letting me know you can't install docker-ce-cli without the daemon!

I did see this:  https://github.com/microsoft/vscode-dev-containers/blob/main/containers/docker-in-docker/README.md.  my slight concern is more geared towards vscode even proposing this as a standard due to the issues in the link in my first comment.  That being said, maybe it's not that big of a concern anymore.

tschaffter:
The stack relies on a dockerized MongoDB instance. The development workflow should also enable to build docker images for the api and web app.

VS Code offers another dev container feature called `docker-from-docker` where the container connects to the Docker engine of the host. This comes with some benefit like being able to reuse the cache of the host engine and be able to orchestrate all the images and containers running on the host from the dev container. Yet `docker-in-docker` provides an isolation from the host Docker engine that may be preferable for reproducibility. :)

thomasyu888:
@tschaffter got it.  So if I'm understanding correctly - you are going to be hosting 3 docker containers within the dev container (API, web app, mongodb) and the dev container will only be used for development. In actual practice, will you will still host the 3 containers independently (not within a docker container).  I ask because I'm almost certain there will be institutions against docker-in-docker

It's an interesting point you bring up about reproducibility - I wonder if it is possible that you run into an issue that something works within the dev container but not when you actually host the applications on aws due to the differences in operating system.

tschaffter:
@thomasyu888 Actually to run the stack in development mode, only one docker container is required (MongoDB). The Python Flask app is run with python and the web app is run with Node.js in watch mode.

VS Code has three options for running dev containers: 1) by specifying the name of an existing image, 2) by specifying a Dockerfile and build options (what I do), 3) similar to 2) but using a `docker-compose.yml`. With the third option, I could create a Docker service for MongoDB so that I would not need to deploy it from within the dev container. However, option 2) is more convenient as the existing development workflow + documentation can be applied as-is inside the dev container. Docker inside the dev container is also useful to check that the Docker images of the project build successfully.

docs/dev-container.md
```diff
@@ -0,0 +1,61 @@
...
+- [Docker]
+- [VS Code] and the extension [Remote - Containers]
+
+## Quick Start
+
+Open the workspace folder in VS Code after cloning it. VS Code should invite you
+to open the folder in the dev container specified by this repository.
+Alternatively:
+
+- Press `F1` or click on `Help` > `Show All Commands`
```
rrchai, line 23: I don't have "F1" on my external keyboard 😄. We could also add shortcuts for vscode to open "All Commands". But it is not necessary change and you can make it call.

<kbd>Command</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd> (macOS) or <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd> (Windows)

```suggestion
- Press `F1` or <kbd>Command</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd> (macOS) or <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd> (Windows) or click on `Help` > `Show All Commands`
```

rrchai (approved):
LGTM! 🚀

PR #85, opened 2022-03-12
Add pyenv to development requirements

### LINKED ISSUE #86: Fix unspecified Python dependencies in README
@vpchung reported the following issue:

Might I suggest some additional content to the README? I don’t know if maybe my case is just special and [@Rong (he/him)](https://sagebionetworks.slack.com/team/U01DV4J0J4B) did not experience the same errors as me, but following “Running with Docker” verbatim was met with some errors, e.g. challenge-registry-prepare requires Python 3.9 (never mentioned so our users wouldn’t know) and pipenv already installed (not mentioned until the “Development” section).
We can provide a “shortcut” to users with:

```console
conda create -n rocc nodejs python=3.9
conda activate rocc
conda install -c conda-forge yarn
pip install pipenv
```

tschaffter:
Fixes #86 

When developing on the host, `pipenv` is specified as a requirements for Python projects. This applies to the project `apps/api` and in the future probably to the Challenge Registry API client library for Python. However, `pipenv` requires `pyenv` in order to install the version of Python specified in the files `Pipfile` and `Pipfile.lock`.

The first change made by this PR is to add `pyenv` to the list of development requirements in the README.

The first way deployment method described in the README requires to run the command `challenge-registry-prepare`. For the project `apps/api`, this will create the configuration file `.env` and install the python environment with `pipenv`. Note that installing this virtualenv is actually not required for the sake of building the docker image with `yarn docker` as described in the README. To make it so that building - and later running - the docker images do not required `pipenv` and `pyenv`, I propose to decouple the creation of the `.env` file (required to run docker container) and the creation of the virtualenv (not required to build and run the containers). See changed files.

```console
$ yarn python
yarn run v1.22.17
$ bash -c '. dev-env.sh && challenge-registry-python'

 >  NX   Running target python for 1 project(s):

    - api

 ————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

> nx run api:python

Installing dependencies from Pipfile.lock (523cb0)...
Loading .env environment variables...
Collecting click==8.0.3
  Using cached click-8.0.3-py3-none-any.whl (97 kB)
Installing collected packages: click
  Attempting uninstall: click
    Found existing installation: click 7.1.2
    Uninstalling click-7.1.2:
      Successfully uninstalled click-7.1.2
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
flask 1.1.4 requires click<8.0,>=5.1, but you have click 8.0.3 which is incompatible.
Successfully installed click-8.0.3

 ————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 >  NX   Successfully ran target python for 1 projects


   See Nx Cloud run details at https://nx.app/runs/KcKJwMMenV8

Done in 4.06s.
```

## Changelogs

- In the README, move the preparation instructions common to the sections "Running with Docker" and "Development" in their own section named "Preparing the workspace".
- Extract `yarn --install` from `challenge-registry-prepare`.
  - The motivation is to now make use of Node.js dependencies cache in the CI workflow AND use in the CI workflow the same set of instruction as documented in the README.
- Extract the creation of Python virtualenvs from the API task `prepare`. The API task `prepare` now only creates the configuration file (`.env`).
  - The motivation is to not require the developer to have `pyenv` and `pipenv` installed when they want to run the stack with Docker.


tschaffter:
## TODO

- [ ] Update CI workflow to call `yarn python`

vpchung (approved):
As someone who tends to copy-paste first, read later, the updated instructions are great!

The only thing I would mention is that `yarn seed-db` would only work IF the app has already started.  It's been hinted at by how the instructions are written but perhaps we can make it even more explicitly clear by mentioning something like, open a new terminal and run `yarn seed-db`.

tschaffter:
> The only thing I would mention is that yarn seed-db would only work IF the app has already started.

Only the MongoDB instance needs to be up as the CLI talks directly to it.

vpchung:
> Only the MongoDB instance needs to be up as the CLI talks directly to it.

Which, based on the README, is started up by `yarn start`.  (Sorry, that's what I meant by "if the app has already started"!)

PR #152, opened 2022-04-12
Add --runInBand to jest testing on large components

rrchai:
This will significantly speed up the testing. Also, add "text" in report format to show coverage in cli. Take `web-ui` as an example:

---
Screen Shot 2022-04-12 at 12 02 39 PM

### DIFF
--- a/jest.preset.js
+++ b/jest.preset.js
@@ -3,7 +3,7 @@ const nxPreset = require('@nrwl/jest/preset');
 module.exports = {
   ...nxPreset,
   collectCoverage: true,
-  coverageReporters: ['html','json','lcov'],
+  coverageReporters: ['text', 'html','json','lcov'],
   coverageThreshold: {
     "global": {
       "branches": 0,
--- a/jest.preset.js
+++ b/jest.preset.js
@@ -2,6 +2,7 @@ const nxPreset = require('@nrwl/jest/preset');
 
 module.exports = {
   ...nxPreset,
+  runInBand: true,
   collectCoverage: true,
   coverageReporters: ['text', 'html','json','lcov'],
   coverageThreshold: {
--- a/libs/web/challenge-search/project.json
+++ b/libs/web/challenge-search/project.json
@@ -9,8 +9,7 @@
       "outputs": ["coverage/libs/web/challenge-search"],
       "options": {
         "jestConfig": "libs/web/challenge-search/jest.config.js",
-        "passWithNoTests": true,
-        "runInBand": true
+        "passWithNoTests": true
       }
     },
     "lint": {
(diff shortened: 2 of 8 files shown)


tschaffter:
> This will significantly speed up the testing.

According to [the doc](https://jestjs.io/docs/cli#--runinband), using the option `--runInBand` runs test in series instead of parallel, thus increasing the time needed to complete the tests.

tschaffter:
@rrchai Please document in this discussion the additional info you shared with me via Zoom. I'll give this PR a spin and comment and the behavior of `--runInBand` I observe.

tschaffter:
Here are benchmarks with and without `--runInBand`.

With `--runInBand`

```console
$ npx hyperfine --runs 10 'nx test web-ui --skip-nx-cache'
Need to install the following packages:
  hyperfine
Ok to proceed? (y) y
Benchmark 1: nx test web-ui --skip-nx-cache
  Time (mean ± σ):     35.378 s ±  1.747 s    [User: 12.431 s, System: 5.021 s]
  Range (min … max):   33.311 s … 39.175 s    10 runs
```

Without:

```console
$ npx hyperfine --runs 10 'nx test web-ui --skip-nx-cache'
Benchmark 1: nx test web-ui --skip-nx-cache
  Time (mean ± σ):     41.374 s ±  2.354 s    [User: 26.405 s, System: 9.975 s]
  Range (min … max):   37.850 s … 46.306 s    10 runs
```

tschaffter:
According to this [SE thread](https://stackoverflow.com/questions/43864793/why-does-jest-runinband-speed-up-tests),  it seems that

> ...using the `--runInBand` helps in an environment with limited resources.

A few people have reported improved performance even on computer that they would not consider as having limited resources.

Another benefit of `--runInBand` is that the output is more readable because the output of the tests are ordered.

I would be interested in running the above benchmark on my PC which has 8 CPU cores and more memory. Would the performance with `--runInBand` improve or decrease?

I read at a couple of places that a good place to use `--runInBand` is in the CI/CD workflow where resources currently provided by GitHub are clearly limited.

tschaffter (changes requested):
The option `--runInBand` seems a good find that improves the performance when compute resources are limited. Instead of adding the option to each project - as we would forget to add it to new projects - I propose to move this option to `jest.config.js` (I placed the coverage options in `jest.preset.js` but maybe `jest.config.js` would be a better place for them).

rrchai:
I ran in the devcontainer with 4CPU.  Adding `--runInBand` (bottom) also speed up a bit.
Screen Shot 2022-04-18 at 11 42 56 PM

tschaffter (approved):

PR #162, opened 2022-04-25
Add GUI plugin in vscode for jest

### LINKED ISSUE #154: Add VS Code plugin that gives a GUI for Jest in watch mode
Add the following extension to the list of recommended VS Code plugins in both folders `.vscode` and `.devcontainer`.

https://marketplace.visualstudio.com/items?itemName=Orta.vscode-jest

rrchai:
- fixes #154 

---
❌  force the extension to run jest testing via nx 

```json
"jest.jestCommandLine": "bash -c '. ./dev-env.sh && challenge-registry-test'",
```
The test can run successfully, but the UI is not displayed properly: 1. pass/fail icons not showing anymore 2. folder is not able to expand 3. the test result (0/0 passed) is not right. Not sure how to make this extension fit our nx config yet. 

Screen Shot 2022-04-25 at 12 40 28 PM

I just use default settings for now, the caveat is that it will not benefit from the nx caches. The `--runInBand` is enabled by default setting of this extension, so the overall speed is not bad. 

Screen Shot 2022-04-25 at 1 55 43 PM

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -24,5 +24,11 @@
   "eslint.workingDirectories": [
     "./apps/db-cli",
     "."
-  ]
+  ],
+  "jest.jestCommandLine": "bash -c '. ./dev-env.sh && challenge-registry-test'",
+  "jest.autoRun": {
+    "watch": false,
+    "onSave": "test-file",
+    "onStartup": ["all-tests"]
+  }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -26,7 +26,8 @@
     "."
   ],
   "jest.autoRun": {
-    "watch": true,
-    "onStartup": ["all-tests"]
+    "watch": false,
+    "onSave": "test-file",
+    "onStartup": []
   }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -25,10 +25,8 @@
     "./apps/db-cli",
     "."
   ],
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -24,5 +24,9 @@
   "eslint.workingDirectories": [
     "./apps/db-cli",
     "."
-  ]
+  ],
+  "jest.autoRun": {
+    "watch": true,
+    "onStartup": ["all-tests"]
+  }
```
tschaffter, line 31:
```suggestion
  "jest.autoRun": {
    "watch": false,
    "onSave": "test-file",
    "onStartup": []
  }
```

Motivations:
- Do not run all the tests at startup because this may take too much time and not be required by the developer.
-`watch: true` will runs the tests each time a component or test file is saved. This is not efficient from a performance stand point. Instead, we should use `watch: false` and `onsave: "test-file"` to trigger tests only when saving test files.

tschaffter (changes requested):
Running tests when saving test files may save us time. This VS Code plugin could also help us to identify more quickly which tests have failed and navigate to their code. The plugin is expected to work well on a performance computer like my PC but not like my work MBP where tests take ages to run. This negative point should be solved by using suitable EC2 instances to run the monorepo dev container (https://github.com/Sage-Bionetworks/challenge-registry/issues/159).

tschaffter (approved):

PR #170, opened 2022-05-08
Add .code-workspace for API spec development

### LINKED ISSUE #169: Explore VS Code multi-root workspaces
I discovered this feature of VS Code when reading about [Jest testing workflow in monorepos](https://github.com/jest-community/vscode-jest#how-to-use-the-extension-with-monorepo-projects).

The goal of this ticket is to explore whether we can benefit from using VS Code multi-root workspaces.

tschaffter:
Fixes #169

## Changelog

- Define a multi-root VS Code workspace that includes only folders relevant to the development of the challenge registry API spec

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ nx-cloud.env
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
+!.vscode/*.code-workspace
 
 # misc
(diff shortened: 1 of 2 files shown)


tschaffter:
I'm shelving this PR because we currently don't have a use case for using multi-root VS Code workspaces.

PR #303, opened 2022-06-09
Add list of contributors to the README

tschaffter:


vpchung (approved):

tschaffter:
At some point I would like to list non-developer contributors too as well as include roles and responsibilities. I looked for standard and found the following resources:

- [all-contributors/all-contributors](https://github.com/all-contributors/all-contributors)
  - Promising but I'm not a fan of the design that may not be accessible (rely on tooltips)
- [contrib.rocks](https://contrib.rocks/preview)

image

- [CRediT](https://casrai.org/credit/)
- [data2health/contributor-role-ontology](https://github.com/data2health/contributor-role-ontology)

rrchai (approved):

PR #304, opened 2022-06-09
Add alias `challenge-docker-stop` to stop running containers

tschaffter:
The projects that compose the backend have now the task `serve-detach` that starts the project using docker. For example, running `challenge-registry-serve` automatically starts all the required backend services using docker.

I currently need to often stop all the containers using `docker stop $(docker ps -q)`. It's not a complicated command but I could use an alias, which is what this PR suggests. The alias is `challenge-docker-stop`.

### DIFF
--- a/dev-env.sh
+++ b/dev-env.sh
@@ -96,3 +96,7 @@ function challenge-welcome {
     challenge-nx-cloud-help
   fi
 }
+
+function challenge-docker-stop {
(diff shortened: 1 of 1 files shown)


tschaffter:
The alias is prefixed with `challenge-` to group it with other aliases defined by this project. I could be convinced to rename it `docker-stop`. :-)

vpchung (approved):

PR #359, opened 2022-06-23
Update Footer Component

rrchai:
Here are the changes made in this PR:

1. Assuming the palettes and fonts are used globally for `challenge-registry` project, so I move the themes from `libs/challenge-registry/ui/src/_lib-theme.scss` to `apps/challenge-registry/src/_app-theme.scss`. In this way, other libs outside of `ui` library, like homepage, can also use this theme. 
2. Updating the code to extract the fonts from the config in `_footer-theme.scss`, so we don't need to import fonts each component.
3. Moving all "fonts-*" styles into typography.


rrchai:
Re @vpchung 
> > In this way, other libs outside of ui library, like homepage, can also use this theme.
> 
> Interesting... I was able to get the [homepage component](https://github.com/Sage-Bionetworks/challenge-registry/pull/358#issuecomment-1164031128) to use the color and typography successfully while they were defined in `libs/challenge-registry/ui/src/_lib-theme.scss`. Granted, I did have to tweak `_lib-theme.scss` a bit, by defining the typography levels and passing that to `mat.core()`.

vpchung:
@rrchai ahhh okay. I think you are right - we should probably move the app-wide theme configs to `apps/challenge-registry/src/_app-theme.scss`.  I failed to notice this file, as I only worked inside `libs/challenge-registry`... my bad!

libs/challenge-registry/ui/src/lib/footer/_footer-theme.scss
```diff
@@ -23,13 +22,30 @@
 
 @mixin typography($theme) {
   $typography-config: mat.get-typography-config($theme);
+  $font-family: map.get($typography-config, 'font-family');
```
vpchung, line 25: Not sure if it's only on my end, but the `Poppins` font doesn't seem to reflect on the app:
Screen Shot 2022-06-24 at 12 29 15 AM

vpchung: P.S. I do see it being used in the navbar though! 🤣

vpchung (changes requested):

libs/challenge-registry/ui/src/lib/footer/_footer-theme.scss
rrchai: Good catch! I thought it worked when I see font-family was using `Poppins` on footer, but it should not contain quotes (either single or double quotes).
Screen Shot 2022-06-24 at 9 56 31 AM

> I do see it being used in the navbar though! 
I don't know why navbar is smarter than footer 🤣

rrchai: Both of below ways are working:
1. 
```scss
@mixin typography($theme) {
  $typography-config: mat.get-typography-config($theme);
  $font-family: map.get($typography-config, font-family);
  footer {
    font-family: $font-family;
  }
}
``` 
We would also need to replace`$font-family: "'Poppins', 'sans-serif'"` with `$font-family: 'Poppins, sans-serif'` in this way.

2. 
```scss
@mixin typography($theme) {
  $typography-config: mat.get-typography-config($theme);

  footer {
    font-family: mat.font-family($typography-config);
  }
```

I will update to use the second way since it is shorter.

tschaffter (approved):

PR #377, opened 2022-06-27
Enable user service to communicate with the core service

### LINKED ISSUE #375: Enable the User Service to communicate with the Challenge Core Service
Here I’m going to use Spring Cloud OpenFeign as the REST API client and configure it with spring boot based microservices.

## References

- [How to Use Feign Client in Spring Boot](https://javatodev.com/how-to-use-feign-client-in-spring-boot/)
- [Consuming REST API Using Feign Client in Spring Boot](https://javatodev.com/consuming-rest-api-using-feign-client-in-spring-boot/)
- [Microservices – Communication With Spring Cloud OpenFeign](https://javatodev.com/microservices-communication-with-spring-cloud-openfeign/)

tschaffter:
- Fixes #375 

## Changelog

- Enable the user service to retrieve information from the core service using [Spring Cloud OpenFeign](https://spring.io/projects/spring-cloud-openfeign).
  - The current approach enables the user service to communicate with the core service via the HTTP protocol.
  - In the future, using a messaging protocol like RabbitMQ is recommended for optimal performance.
- Add support for global exceptions found in the GH repository of this series of articles (I still need to read this article)

## Note

Creating a user via the user service requires that prior information about the user has been inserted in a table managed by the core service, namely the identification number of the user and their email address. This scenario may make sense for a banking system where a client would go to the bank where a clerk would insert information about the client in their database (here using the core service). Then, the client would be invited to create a login for the bank web app (here using the user service).

The above strategy is not viable for a user who registers to the challenge registry. I will update the strategy to meet our need in an upcoming PR.

I'm wondering if the existing strategy could be used for challenge organizations in a scenario where we would onboard organizations first similarly to how a client opens a bank account.


tschaffter:
## Registering users

The current implementation from the series of tutorial works as follow:

1. A row in the `challenge_core_user` table includes core information about the user such as first name, last_name, email and identification_number.
2. The object below can be sent to the user service to register a new user. The user service then contacts the core service to check if a user with the specified identification number exists. If yes, the user service creates an account for the user in Keycloak and creates a row in the table `challenge_user`.

Issue #155, opened 2022-04-18
Enable Search Engine Optimization (SEO)


tschaffter:
The goal is to enable Google to understand the content of the pages in the web client. This could be achieved using schema.org and/org JSON-LD. See https://developers.google.com/search/docs/advanced/structured-data/intro-structured-data.

> We will make our application search engine (SEO) friendly and compatible with social media crawlers (for example Facebook or Twitter).

[Source](https://blog.angular-university.io/angular-universal/)

rrchai:
quote from @@tschaffter 
> Let’s take an e-commerce site as an example. This e-commerce site has thousands of clients and thousands of products, and for each product there is a details page.
Because it’s an e-commerce site, it should be server-side rendered (probably using Angular Universal) in order for crawlers to read its content that is needed for SEO.
Source: https://itnext.io/incremental-static-regeneration-for-angular-42b0a8440e53

tschaffter:
Here are a few articles I found but haven't ready (yet).

- https://itnext.io/incremental-static-regeneration-for-angular-42b0a8440e53
  - https://angular.io/guide/universal
    - https://web.dev/rendering-on-the-web/

tschaffter:
Server-side rendering is made possible by [Angular Universal](https://angular.io/guide/universal).

tschaffter:
## Style not fully applied with SSR

- The reason is likely because the font we use is not found on the server.
- It is not possible to inspect the style of an SSR page using the browser developer tools (`element {}`)

### CSR

![image](https://user-images.githubusercontent.com/3056480/181516819-8743b267-9aa9-4351-becd-a9ed6f96dc53.png)

### SSR

![image](https://user-images.githubusercontent.com/3056480/181516893-7578e350-34bf-4cb5-9154-42bce7deb98a.png)

![image](https://user-images.githubusercontent.com/3056480/181516901-9b72be03-9a02-49a9-978f-d04918390a77.png)

PR #444, opened 2022-07-25
feat(app): Authenticate the user using Keycloak

### LINKED ISSUE #315: Enable the user to log wtih Keycloak

### LINKED ISSUE #316: Enable the user to get their profile using microservices
We now have an endpoint to the the profile of a user by id. The goal of this ticket is to use this endpoint to show the user information on their profile page.

Depends on #655

### LINKED ISSUE #317: Enable the user to logout

tschaffter:
- Fixes #315
- Fixes #317

## TODO

- [x] Can logout by clicking on the logout button.
- [x] Access Keycloak user information.
  - The web app know when the user is logged in or not.
  - I can get the KC user profile.
- [ ] ~Access User object from the challenge-user-service~
- [ ] ~Fix guard infinite loop issue when accessing a protected page without being logged in~
- [x] Can log in and be redirected to the previously visited page.
- [x] Remove `home/movie-list` folder.


tschaffter:
## Issue

I'm redirected to KC login page when trying to access a protected page. I can successfully login, that is, get an access token, BUT KC ends up in an infinite page loop and is not able to navigate back to CR. I bumped into a similar issue in the past but I thought I temporarily solved the issue by tweaking the auth guard.

tschaffter:
Need to better understand the [`state` query parameter](https://medium.com/keycloak/the-importance-of-the-state-parameter-in-oauth-5419c94bef4c).

tschaffter:
## Issue - not related to the auth guard

When logging in and after being redirected to the app, the url include query params that ideally shouldn't be displayed. For example, http://localhost:4200/#state=485d2875-7d41-49ac-8820-fedcde504af8&session_state=703036ef-f941-4ae3-b56a-1a5b774fdab0&code=2dea4630-436d-4f07-b644-8ffecbc01c16.703036ef-f941-4ae3-b56a-1a5b774fdab0.221d4622-5864-4eed-895e-5246058d817d

tschaffter:
- https://stackoverflow.com/a/66131477

tschaffter:
Create an interceptor:

```console
$ nx g @nrwl/angular:interceptor oidc-interceptor  --project=challenge-registry-auth

>  NX  Generating @nrwl/angular:interceptor

CREATE libs/challenge-registry/auth/src/lib/oidc-interceptor.interceptor.spec.ts
CREATE libs/challenge-registry/auth/src/lib/oidc-interceptor.interceptor.ts
```

tschaffter:
Another solution to remove the OIDC query parameters, but would need to be done on all pages, unless the login redirection lands always on the same page.

https://stackoverflow.com/questions/48552993/angular-5-remove-query-param

tschaffter:
Actually maybe the solution is at the level of the router: https://www.codegrepper.com/code-examples/typescript/angular+remove+query+params+from+url+without+reload

tschaffter:
This is a suggestion for cleaning the URL is the auth service:

https://github.com/manfredsteyer/angular-oauth2-oidc/issues/1071#issuecomment-833751535

tschaffter:
## Trying `useHash: false`

```console
@NgModule({
  imports: [
    RouterModule.forRoot(routes, {
      initialNavigation: 'enabledBlocking',
      useHash: false,
    }),
  ],
  declarations: [],
  providers: [],
  exports: [RouterModule],
})
export class AppRoutingModule {}
```

This change does not remove the fragments in the URL after the login redirections.

According to the article [Angular — The hash(#) trap](https://medium.com/@dao.houssene/angular-the-hash-trap-b2d415c2c241), disabling the hash can be done by using Angular Universal, which is the solution that enables server-side rendering. This is something that we already wanted to explore in #155 .

tschaffter:
## Cleaning fragment with a guard

- https://github.com/angular/angular/issues/22598

PR #550, opened 2022-08-15
feat: Notify the developer when a different dev container must be used

### LINKED ISSUE #543: Invite the developer to rebuild the devcontainer when devcontainer.json uses an image different from the one used
Errors may occur if the tools provided by the dev container and the tools expected by the code base (defined in `devcontainer.json`) are not in sync. We had a similar issue with node packages, which we solved by automatically running `yarn install --frozen-lockfile` when switching or rebasing  branches. Here we can't do that automatically for devcontainer as the developer needs to shutdown the devcontainer and rebuild it.

The idea is display a message in the terminal if the image version defined in `devcontainer.json` and the version of the current dev container are not the same.

cc: @rrchai (following our discussion on Friday)

tschaffter:
Fixes #543

### DIFF
--- a/.husky/post-merge
+++ b/.husky/post-merge
@@ -1,4 +1,8 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-rm -rf node_modules && yarn install --frozen-lockfile
+# rm -rf node_modules && yarn install --frozen-lockfile
+yarn install --frozen-lockfile
+
+# Notify the developer when a different dev container must be used.
+node tools/check-devcontainer-version.js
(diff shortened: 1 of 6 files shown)


.husky/post-merge
```diff
@@ -1,4 +1,8 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-rm -rf node_modules && yarn install --frozen-lockfile
```
tschaffter, line 4: Not sure why the article I used as reference is removing the folder in this hook but not in other hooks. I'm trying without removing it and will see.

PR #617, opened 2022-08-27
feat: Store large and binary file with DVC

### LINKED ISSUE #616: Explore DVC
GitHub is [not suitable to version control large files and binaries](https://youtu.be/RcqLV1lU408). Tracking such files will permanently slow down git commands. One solution is to rewrite the history to remove these files, thus breaking the rewritten past commits.

Another way is to track large and binary files outside of git, for example using git LFS or [DVC](https://dvc.org/). I have already explored GitHub LFS and have deemed it not suitable (see #584). This task will focus on exploring DVC.

Tasks:

- [x] Initialize DVC for this repo.
- [x] Configure a remote storage with AWS S3.
- [ ] Use case: move bitmap images tracked on GitHub to DVC storage.
- [x] Add DVC to the dev container.
- [ ] Document how to use DVC.
- [x] Evaluate VS Code extension for DVC.
    - This extension only benefits to projects that run ML experiements, which is not the case of our project.
- [x] Find a way to avoid `dvc unprotect` before updating files.
    - https://dvc.org/doc/user-guide/large-dataset-optimization
- [ ] Add git hooks
  - `dvc checkout` after git clone and checkout
  - https://dvc.org/doc/command-reference/checkout#example-automating-dvc-checkout
  - https://dvc.org/doc/command-reference/install
- [ ] Share DVC with Sagers - I can see this tool being useful to researcher and bioinformatics engineers/analysts.

tschaffter:
Fixes #616


tschaffter:
The primary motivation for exploring DVC was to store binary assets like images. We are now storing images on S3 so using DVC is no longer needed.

PR #626, opened 2022-08-30
feat(ui): Centralize styles and constants

### LINKED ISSUE #534: Centralize sytle used by the user profile page
The components of the user profile page use duplicated style. Some of this style should be placed in a shared style file.

### LINKED ISSUE #535: Centralize theme used by the user profile page
The components of the user profile page use duplicated theme. Some of this theme should be placed in a shared theme file.

rrchai:
- fixes #534
- fixes #535


libs/challenge-registry/styles/src/lib/_colors.scss
```diff
@@ -0,0 +1,32 @@
+// Figma color variables
```
vpchung, line 1: Are we using this instead of palettes?

vpchung (approved):

vpchung:
One clarifying comment.

apps/challenge-registry/src/_app-theme.scss
```diff
@@ -1,13 +1,12 @@
 @use '@angular/material' as mat;
-@use 'palette' as pal;
+@use 'libs/challenge-registry/themes/src/palettes' as pals;
```
tschaffter, line 2:
```suggestion
@use 'libs/challenge-registry/themes/src/palettes' as palettes;
```
Consider using more descriptive/standardized variable names. This suggestion may apply to other variables defined in this PR.

tschaffter (commented):
I'm looking forward to this PR being merged! Good job!

libs/challenge-registry/styles/src/lib/_colors.scss
rrchai: Sorry for late reply. Those color variables (`libs/challenge-registry/styles/src/lib/_colors.scss`) are not in the theme palettes (`libs/challenge-registry/themes/src/_palettes.scss`) yet. It's the location for now to store the global color variables that are not related with themes palettes.

vpchung: Oh interesting.  I created the palette with the colors from Figma in mind, so they should've ported over to teleportHQ.  

Let's try to use something similar?

rrchai: @vpchung  I am still not able the palettes in [_palettes.scss](https://github.com/Sage-Bionetworks/challenge-registry/blob/main/libs/challenge-registry/ui/src/lib/_palette.scss) for those variables. 

Taking `$dl-color-default-accent1: rgba(255, 210, 218, 1);` as an example, in the [figma](https://www.figma.com/file/yAeIrg6Whc8Bqj2EnG4MWP/Challenge-Registry), it shows the hex code of `accent1` is `#FFD2DA`. It is also the same hex code from [rgba-hex converted](https://rgbacolorpicker.com/rgba-to-hex) using `rgba(255, 210, 218, 1)`, but it's not found in the [_palettes.scss](https://github.com/Sage-Bionetworks/challenge-registry/blob/main/libs/challenge-registry/ui/src/lib/_palette.scss) file.

rrchai: Correct me if I am wrong. To clarify the usage difference between  `_colors.scss` and `_palettes.scss`  : 
  - The colors in the `_colors.scss` file are used for the global color variables; one variable name to one color. 
  - The colors in `_palettes.scss` are used to change/define the app's themes and they are usually the material palette object with different levels of hues and contrasts. In order to use the colors defined in this file, the colors need to be configured into the theme first, i.e: 
    ```scss
    ...
    $awsome-colors: mat.define-palette(palettes.$awesome-color, 400, 600, 800);
    $theme: mat.define-light-theme(
      (
        color: (
          ...
          my-colors: $awsome-colors,
        ),
    ```

As we discussed, there are two approaches:

1.  Move these colors (from figma) into [themes/src/_palettes.scss](https://github.com/rrchai/challenge-registry/blob/feat-fix-534/libs/challenge-registry/themes/src/_palettes.scss) >> Configure these colors into app's theme >> Extract the needed color in each component
3. Keep these colors (from figma) in the [styles/src/lib/_colors.scss](https://github.com/rrchai/challenge-registry/blob/feat-fix-534/libs/challenge-registry/styles/src/lib/_colors.scss) >> Extract the needed color in each component

@vpchung  @tschaffter  which approach would you prefer to go forward?

vpchung: > I am still not able to find these variables. in _palette.scss

I identified the color translations below:

```
$dl-color-default-navbar: ($dark-blue-palette, 600)
$dl-color-default-accent1: ($accent-pink-palette, 100)
$dl-color-default-accent2: ($accent-orange-palette, 100)
$dl-color-default-accent3: ($accent-green-palette, 100)
$dl-color-default-accent4: ($accent-purple-palette, 50)
$dl-color-default-primary1: ($dark-blue-palette, 400)
$dl-color-default-primary2: $dark-focused
$dl-color-default-navbardark: ($dark-blue-palette, 800)
$dl-color-default-secondary1: ($light-blue-palette, 300)
$dl-color-default-secondary2: ($accent-purple-palette,300)
$dl-color-default-darkaccent2: ($accent-orange-palette, 300)
$dl-color-default-darkaccent3: ($accent-green-palette, 600)
$dl-color-default-darkaccent4: ($accent-purple-palette, 200)
$dl-color-default-darkacccent1: ($accent-pink-palette, 200)
```

I think what happened is that Material's color tool (what I used to help create `_palette.scss`) sometimes alter what the "base" color is as I cycle through the levels, so I didn't notice that the original Hex code changed.  That being said, using the above levels should look about the same!

rrchai: Thanks for identifying these colors and that's really helpful! 

For [this commit](https://github.com/Sage-Bionetworks/challenge-registry/pull/626/commits/60e23fb2e992ba6b4595e1184d11b5ec8dda6a91), I moved the figma variables to `_palettes.scss`. 

---
If we decide to go with the option 1 ^, we would need to do further updates.

FYI: Based on my testing, the key names in `mat.define-light-theme` can only use material default names, i.e `primary`, `accent`, `warn`, but not accept custom key names, i.e `figma: $accent`:
```scss
$theme: mat.define-light-theme(
  (
    color: (
      primary: $primary,
      accent: $accent, // working
      figma : $accent, // not working
    ),
 ```

Comment (vpchung) [reply libs/challenge-registry/styles/src/lib/_colors.scss]: > we would need to do further updates
What would be the easier/more convenient approach?  At this point, I think you have the most experience out of all of us, so I'm more inclined to follow what you would rather do.

Comment (vpchung) [reply libs/challenge-registry/styles/src/lib/_colors.scss]: > If we decide to go with the approach 1 ^, we would need to do further updates.

Hm, I'm trying to think which option would be more the "Angular Material" standard...

But honestly, at this point, I think you have the most experience out of us 3, so I'm more inclined to follow what you think is best/easiest.

Comment (tschaffter) [reply libs/challenge-registry/styles/src/lib/_colors.scss]: My only recommendation is to place anything related to colors in the `themes` library because we have defined "themes" as color + typography and "styles" as layout.

Comment (rrchai) [reply libs/challenge-registry/styles/src/lib/_colors.scss]: Thanks for the suggestions, @Verena and @tschaffter.

I think the approach 1 is more standard way to call the colors from the defined themes, while approach 2 is easier since we don't need to change variable names each time we implement teleportHQ exported codes.

However, I managed to be able to add custom variable names to the theme object. Now, we can still call the figma colors from the theme using the same variable names exported from teleportHQ in each component. 

What I changed in this [commit](https://github.com/Sage-Bionetworks/challenge-registry/pull/626/commits/bf0781c):
1. Save all figma-teleportHQ exported color variables into `$figma-collection` object
2. Add `$figma-collection` into color object of defined themes using `mat.deep-merge` 
3. To call these color variables, we can directly extract the figma colors from the theme, i.e:
```scss
  $config: mat.get-color-config($theme);
  $figma: map.get($config, 'figma');

  .base-top-section {
    background-color: map.get($figma, dl-color-default-hover1);
  }
```
---
>My only recommendation is to place anything related to colors in the themes library because we have defined "themes" as color + typography and "styles" as layout.

Yes, we have placed all colors in the themes library

vpchung: Sounds good to me, thanks @rrchai !!

PR #628, opened 2022-08-30
feat(ui): Centralize fonts and clean footer themes

rrchai:
This PR depends on #626


vpchung (approved):

apps/challenge-registry/src/_app-theme.scss
```diff
@@ -1,24 +1,24 @@
 @use '@angular/material' as mat;
-@use 'palette' as pal;
+@use 'libs/challenge-registry/themes/src/lib/fonts' as fonts;
+@use 'libs/challenge-registry/themes/src/lib/palettes' as pals;
```
tschaffter, line 3:
```suggestion
@use 'libs/challenge-registry/themes/src/lib/palettes' as palettes;
```
Using `palettes` would be more descriptive, unless `pals` is commonly used somewhere else (e.g. Angular Material docs)

libs/challenge-registry/home/src/_lib-theme.scss
```diff
@@ -0,0 +1,5 @@
+@use './lib/featured-challenge-list/featured-challenge-list-theme' as featured;
```
tschaffter, line 1: Consider using a more descriptive name, e.g. `featuredChallengeList` or `featured-challenge-list`.

tschaffter (commented):
This looks great! I left two comments about using more descriptive/standardized variable names, which can maybe be extended to other variables.

libs/challenge-registry/home/src/_lib-theme.scss
rrchai: has changed to `featured-challenge-list`

rrchai:
TODO:
- [ ] (@vpchung suggests ) Try to use `typography-config` to see if the fonts will automatically configured for all pages. If so, we don't need to re-specify the typography styles in each component.

PR #636, opened 2022-09-02
feat(ui): Update challenge card in User Profile page

### LINKED ISSUE #536: Set the background color of the status chip

rrchai:
This pr will need to further update - depends on #626 
- fixes #536 
- fixes layout of challenge cards

---
Screen Shot 2022-09-02 at 11 12 14 AM


tschaffter:
@rrchai Should this PR be turned into a draft or is it ready to assign reviewers?

libs/challenge-registry/ui/src/lib/challenge-card/_challenge-card-theme.scss
```diff
@@ -98,7 +98,18 @@
     color: #000;
 
     &.active {
-      // wait until the colors are determined
+      border-color: #71c663 !important;
```
tschaffter, line 101: I read in the past that using `!important` should be avoided. I did a quick search to refresh my memory of [why](https://www.w3schools.com/css/css_important.asp).

> Tip: It is good to know about the `!important` rule, you might see it in some CSS source code. However, do not use it unless you absolutely have to.

I'd say that it's OK to use `!important` in the style or theme of a component like done here - though maybe this can be avoided using more specific CSS selectors? - but not in a reusable styles and themes like in the project `challenge-registry-styles` and `challenge-registry-themes`.

vpchung (approved):

libs/challenge-registry/user-profile/src/lib/user-profile.component.scss
```diff
@@ -198,8 +198,9 @@ $dl-radius-radius-round: 50%;
 .base-main-section {
   top: 729px;
   left: 495px;
-  width: 100%;
+  width: calc(100% - 495px);
```
tschaffter, line 201: Could `495px` be replaced by a variable? If it is meant to be equal to `left`, the layout could easily break if one of the two values is changed but not the other.

tschaffter (changes requested):

libs/challenge-registry/user-profile/src/lib/user-profile.component.scss
rrchai: hmm, it's actually an easy way to prevent from breaking the layout. If we stretch width, the right content take the rest of space. If we use variable, that means the width will be fixed on the right content. 

Also, our entire page is using the arbitrary px and absolute position from teleportHQ. I don't think we would change just couple of values, which will probably break the layout anyways :).

libs/challenge-registry/ui/src/lib/challenge-card/_challenge-card-theme.scss
tschaffter: If I remember correctly, the order of priority of CSS styles is defined by the order the .css files are imported in `index.html` (we have no control over this). I think that another factor is the order of the classes specified to an element in the .html file. Is that actually true? If yes, you could create a custom CSS class and list it first/last on the element.

>  I think it's fine to use !important on challenge-card ?

Yes!

libs/challenge-registry/user-profile/src/lib/user-profile.component.scss
tschaffter: I know you know that duplicated values increase the risk of breaking code. Maybe the way you generate/integrate the code reduces this risk. I leave it to up to you.

libs/challenge-registry/ui/src/lib/challenge-card/_challenge-card-theme.scss
rrchai: >  I think that another factor is the order of the classes specified to an element in the .html file. Is that actually true? If yes, you could create a custom CSS class and list it first/last on the element.
 
Yes, it is true. You remind me of the order and the issue is because the theme (`_challenge-card-theme.scss`) is loaded before the `challenge-card.component.scss` where it has:
```scss
.challenge-card-status-tag {
  border: 1px solid;
}
```
If `border` is used and no color provided, it will use the default color, which will overwrite the border color we define in the `_challenge-card-theme.scss` since it's called later. Splitting `border` into `border-width` and `border-style` can avoid overwrite the border color. I will make the changes.

tschaffter:
>**Note**
> The `main` branch is now using Yarn 2+.

rrchai:
> The main branch is now using Yarn 2+.

@tschaffter It looks like there is no need to update the branch. Should I still pull the updates from main and test this PR?

tschaffter:
> Should I still pull the updates from main and test this PR?

I can it now as I planned to do it for all the other open PRs.

tschaffter:
Done. The PR looks good on my end. On your end, you just need to remove your local `node_modules` and run `challenge-install` after pulling code from this feature branch or `main`.

tschaffter (approved):

PR #750, opened 2022-09-30
feat(ui): Add pagination in User Profile > starred

rrchai:
Recording 2022-09-30 at 20 00 07

The card looks like static at the bottom in gif, but I think it's because of scrolling too fast and slow recoding's fps with my internet :(  - please ignore that.


libs/challenge-registry/user-profile/src/lib/user-profile-stats/user-profile-stats.component.html
```diff
@@ -8,9 +8,9 @@ <h4>1</h4>
     <p class="mat-caption">stars</p>
   </div>
   <div class="stat-item action-btn" *ngIf="loggedIn">
-    <mat-icon aria-hidden="true" class="card-icon">edit</mat-icon> Edit Profile
+    <mat-icon aria-hidden="true" class="stat-card-icon">edit</mat-icon> Edit Profile
```
rrchai, line 11: @vpchung  just a note why it's been changed here: Both `challenges-card` and  `user-profile-starred` component have `card-icon` class, which results in the color of `card-icon` will be overwritten by what defined in `challenge-card` (It might be one caveat from using "less" descriptive class name).

Screen Shot 2022-09-30 at 8 06 24 PM

tschaffter (approved):
LGTM! I like the look and feel of PrimeNG.

libs/challenge-registry/user-profile/src/lib/user-profile-stats/user-profile-stats.component.html
vpchung: @rrchai yeah, I realize that when I made the first PR, hence why I think we should eventually extract common classes, then do specific defs!  But... one thing at a time 🙃

libs/challenge-registry/user-profile/src/lib/user-profile-starred/user-profile-starred.component.html
```diff
@@ -18,12 +18,22 @@ <h3>Starred</h3>
       </button>
     </div>
     <ng-container [ngSwitch]="activeTab">
-      <div *ngSwitchCase="tabs['challenges']" class="card-group">
-        <challenge-registry-challenge-card
-          *ngFor="let challenge of challenges"
-          [challenge]="challenge"
-        >
-        </challenge-registry-challenge-card>
+      <div *ngSwitchCase="tabs['challenges']">
```
vpchung, line 21: I think it might be better to place the paginator at the bottom?  That way, as users scroll through the list of cards, they can change the page once they reach the tail end rather than have to scroll back up.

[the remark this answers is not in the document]
vpchung: Alternatively, we could place the paginator at both the top and bottom, similar to how PubMed does it: https://pubmed.ncbi.nlm.nih.gov/?term=crispr&size=200

libs/challenge-registry/user-profile/src/lib/user-profile-starred/user-profile-starred.component.html
```diff
@@ -18,12 +18,22 @@ <h3>Starred</h3>
...
+        <p-paginator
+          [first]="0"
+          [pageLinkSize]="10"
+          [rows]="limit"
+          [totalRecords]="challengeResultsCount"
+          [rowsPerPageOptions]="[10, 20, 30]"
+          (onPageChange)="onPageChange($event)"
```
vpchung, line 28: Is there a way to display the total number of results? Kind of like what PubMed and Synapse does:

**PubMed**
Screen Shot 2022-10-03 at 11 07 56 AM

**Synapse**
Screen Shot 2022-10-03 at 11 08 05 AM

vpchung:
Overall, it looks great to me!  I also appreciate how fast it is.

I just had a few minor comments of features we could possibly add.

rrchai:
Screen Shot 2022-10-07 at 2 48 12 AM

tschaffter (approved):

PR #757, opened 2022-10-03
feat(devcontainer): remove serverless from workspace package.json

tschaffter:
The motivation is to make `package.json` leaner.

- [x] Remove `serverless` from the workspace `package.json`.
- [ ] Add `serverless` to the devcontainer.


tschaffter:
This is no longer required. Using yarn 2+ with PnP may make npm dependency management even faster in the future.

Issue #748, opened 2022-09-30
Upgrade the Challenge Search page to the ROCC Specification v3


tschaffter:
The goal of this Story is to upgrade the frontend and backend of the Challenge Search page to meet the ROCC Specification v3.

## ROCC Specifications

- v3
  - [UI v3 Specification](https://www.figma.com/file/yAeIrg6Whc8Bqj2EnG4MWP/Challenge-Registry)
- v2
  - Legacy code developed before we moved the project to this monorepo.

## Tasks

- [x] Create placeholder page for the challenge search page
- [x] #937
- [x] #938
- [x] #939
- [x] #940 
- [x] #941
- [x] #942
- [x] #943
- [x] #1160
- [x] #1041
- [x] #1176
- [x] #1110 
- [x] #1111
- [ ] #1035
- [ ] #1132

tschaffter:
Added to Sprint 22.10

tschaffter:
Hopefully the work on the challenge search page can start this sprint. Following our last team discussion, the goal is to have the challenge search page implemented by the end of the year (sprint 22.12).

tschaffter:
Added to Sprint 22.12

tschaffter:
I reopened this Story because we only completed the first step.

tschaffter:
@rrchai Let's break down this story into individual tasks. We should then address each of this task with a PR. Do you want to propose a list of task for this story (maybe one task per filter type)?

rrchai:
@tschaffter Sure. In terms of UI, here are the list of tasks:
- Create the placeholder that has been completed via #884 

Filters (using mock-up data):
- Add date picker filter - challenge year
- Add checkbox filter - status, difficulty, submission type, input data type, incentive type, platform
- Add dropdown filter - organization, organizer
- Add sort filter - start/end date, featured, # of stars, # of participants, # of views, alphabetical
- Enable search feature
- Add pagination
- Add list-view challenge card - optional

vpchung:
> Add list-view challenge card (as well as list/grid switch button) - optional

Something to consider before we put effort into the card vs list layouts: https://www.eleken.co/blog-posts/card-ui-examples-and-best-practices-for-product-owners#toc-2

Adam also didn't see a purpose for the view change, so we may want to further discuss this "feature".

rrchai:
[...]

PR #884, opened 2022-11-16
feat(ui): search page

vpchung:
Addresses #748:
search-page

Notes:
* Breadcrumbs still need some alignment adjustments but I figured we had enough to get the idea for now.
* The "Add Challenge" button is not added yet; will incorporate in 2023 Q1 when the service is ready.
* The "Sort by" is not added; I couldn't remember if this was an option with the pagination component already?
* For facets, I only listed 3 for now and get those working first.
* Not sure what's going on but the `mat-select-panel` is transparent and I wasn't able to adjust the background color:

    Screen Shot 2022-11-17 at 2 45 05 PM

    I found a related [SO post here](https://stackoverflow.com/questions/52988519/mat-select-options-became-transparent-after-updating-to-angular-material-v7) -- maybe this is our issue too?
* Couldn't get the `mat-checkbox` to respond..... 

TODOs (will finish in this PR):
- [x] Fix breadcrumb text and icon alignment
- ~[ ] Get `mat-checkbox` to be clickable~

### DIFF
--- a/libs/challenge-registry/challenge-search/src/_lib-theme.scss
+++ b/libs/challenge-registry/challenge-search/src/_lib-theme.scss
@@ -1,0 +1,5 @@
+@use './lib/challenge-search-theme' as challenge-search;
+
+@mixin theme($theme) {
+  @include challenge-search.theme($theme);
+}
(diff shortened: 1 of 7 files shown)


vpchung:
@rrchai @tschaffter I have a couple of remaining TODOs for this PR, but for the most part, everything is ready!  I think what's currently implemented can give us the general idea of what the page will look like.

vpchung:
@rrchai @tschaffter After some exploration, I believe that it's the _animation_ that's not working:

checkbox

Maybe we're missing a JS file?

vpchung:
Okay! This PR is now ready ✨ Final look:

search-final

In the facets column, I'm showcasing how we can have some panels open by default and some closed by default (to mock what is seen in our Figma).

PR #1000, opened 2022-12-14
feat(ui): Update challenge search page

### LINKED ISSUE #937: Add date picker filter - challenge year to the Challenge Search page

rrchai:
- fixes #937 

When I used `ngAfterViewInit`, I got the below error. But replacing by `ngAfterContentInit`, the error is solved.
Screen Shot 2022-12-14 at 11 14 03 AM

Preview
---
Recording 2022-12-14 at 14 58 07


tschaffter:
@rrchai The conflict comes from the upgrade to Angular v15. The v14 Angular components are now named "Legacy". It's OK to continue to use the "Legacy" components for now.

tschaffter (approved):
The component behave as expected.

vpchung (approved):
I'm not a big fan of the box-in-a-box look, but overall, it works great!

tschaffter:
> I'm not a big fan of the box-in-a-box look, but overall, it works great!

I would also agree with this observation. Maybe we could remove the outer box or keep from it only its right vertical border to separate the filters from the rest of the page.

rrchai:
Thanks @vpchung @tschaffter  yes, that was default styling from primeng and I have changed to no border, like below:
Screen Shot 2022-12-15 at 10 03 09 AM

We could adjust how it fits with other filters when I add them back.

PR #1118, opened 2023-01-04
docs: update instructions to reflect workspace changes

vpchung:
One of the changes that came with Angular v15 is that the global `workspace.json` is no longer used (and subsequently, `project.json` files are no longer updated when creating a new library). The docs has been updated to reflect this.


rrchai (approved):
LGTM

tschaffter (approved):
Thanks!

PR #1121, opened 2023-01-04
feat: Add Challenge headline and description

### LINKED ISSUE #1119: Add Challenge tagline
The Challenge card includes a tagline in the challenge registry mockup. There are two strategy to capture this information:

- Add a dedicated property, e.g. `Challenge.tagline`
- Use the property `Challenge.description`

For context, I'm currently working on text search with Hibernate Search 6 and Elasticsearch. This work reminds me of the importance of accepting large descriptions (`Challenge.description`) in order to increase the relevance of text search results.

But there is still value in having a shorter tagline that fits into the challenge card. Hence this proposal:

1. If `Challenge.tagline` is specified, we use it in the Challenge card,
2. otherwise, we display the first N characters from `Challenge.description`.

That way, we don't absolutely need to come up with a tagline for all challenges, though this a tagline is probably a plus to capture the attention of the users.

How does this proposal sound? @vpchung @rrchai

### LINKED ISSUE #1120: Add property `Challenge.description`
Need at this point to test text search using multiple fields.

tschaffter:
- Fixes #1119 
- Fixes #1120 

## Changelog

- Add the property `Challenge.headline`
  - The recommended length in this article is 60 chars. Here I give a bit more room and set the max length to 80 chars.
- Add the property `Challenge.description`
  - The recommended length in this article is between 150 and 160 characters. Here I set it to 280 (tweet length).
  - 280 chars is a compromise between a suitable description for Google SEO and a longer description that contribute to the relevance of results when using text search.
  - Happy to revisit this decision if additional considerations are available.

### DIFF
--- a/libs/challenge-registry/api-description/src/components/schemas/ChallengeCreateRequest.yaml
+++ b/libs/challenge-registry/api-description/src/components/schemas/ChallengeCreateRequest.yaml
@@ -10,11 +10,15 @@ properties:
   #     type: string
   #     minLength: 3
   #     maxLength: 60
-  #   description:
-  #     description: A short description of the challenge
-  #     type: string
-  #     nullable: true
-  #     maxLength: 280
+  headline:
+    description: The headline of the challenge
+    type: string
+    nullable: true
+    maxLength: 80
+  description:
+    description: The description of the challenge
+    type: string
+    maxLength: 200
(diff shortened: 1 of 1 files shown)


tschaffter:
I should not have worked on this feature inside my main branch. This feature is replaced by

PR #1128, opened 2023-01-06
feat(ui): Enable Search Term with mock up service

### LINKED ISSUE #1013: Enable the Search filter
be able to filter the challenges by user input

rrchai:
- fixes #1013 

Preview
---
Recording 2023-01-06 at 11 53 53


libs/challenge-registry/challenge-search/src/lib/challenge-search.component.ts
```diff
@@ -281,4 +324,21 @@ export class ChallengeSearchComponent implements OnInit, AfterContentInit {
...
     }
   }
+
+  // mock up the service to filter challenge by search term
+  private searchChallenges(
+    challenges: Challenge[],
+    searchTerm: string
+  ): Challenge[] {
+    return challenges.filter((challenge) =>
+      (Object.keys(challenge) as [keyof Challenge]).some((k) =>
```
vpchung, line 334: Not sure if this is something we should address here or in a separate ticket, but --

When I search with `dream` (no space), no filtering seems to be done and all 4 cards are still displayed:

Screen Shot 2023-01-06 at 2 31 13 PM

After some sleuthing, I see that this is due to "dream" is part of the [`fullName` attribute](https://github.com/rrchai/sage-monorepo/blob/32ea7d5e057b36fafec2ef0c2c7658928aaf4e16/libs/challenge-registry/ui/src/lib/challenge-card/mock-challenges.ts#L19) for the 4 mock challenges.

If I instead search with `dream ` (with a space) or use the DREAM facet, I get only one card back:

Screen Shot 2023-01-06 at 2 31 19 PM

Thinking as a user, getting the one card back will make the most sense to me.

---

Basically, my question is:
Are we still using the `fullName` attribute?  I don't see it listed anymore in [challenge.openapi.yaml](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/challenge-registry/api-description/build/challenge.openapi.yaml).

vpchung (commented):
Works great!!!

I had a quick(ish) question about the Challenge data model, but maybe we should address that elsewhere?

libs/challenge-registry/challenge-search/src/lib/challenge-search.component.ts
rrchai: > After some sleuthing, I see that this is due to "dream" is part of the [fullName attribute](https://github.com/rrchai/sage-monorepo/blob/32ea7d5e057b36fafec2ef0c2c7658928aaf4e16/libs/challenge-registry/ui/src/lib/challenge-card/mock-challenges.ts#L19) for the 4 mock challenges.

Yes, it's expected with this simple mock-up service, which tries to search for the input string in any properties of challenge. I think @tschaffter has merged the real better service to filter challenges by search term today. We will replace with real service later.

> Are we still using the fullName attribute? I don't see it listed anymore in [challenge.openapi.yaml](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/challenge-registry/api-description/build/challenge.openapi.yaml).

`fullname` was used to define the url path (like {org_name}/{challenge_name}) of the challenge page. I don't remember why we needed it in the legacy repo, but we probably don't not need the `fullname` property anymore. Maybe @tschaffter could answer this question.

tschaffter: > We will replace with real service later.

That's right, the challenge service now supports full-text search (#1104).

> Are we still using the fullName attribute? I don't see it listed anymore in [challenge.openapi.yaml](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/challenge-registry/api-description/build/challenge.openapi.yaml).

No, `fullName` is no longer used.

libs/challenge-registry/challenge-search/src/lib/challenge-search.component.ts
```diff
@@ -134,12 +151,28 @@ export class ChallengeSearchComponent implements OnInit, AfterContentInit {
   }
 
   ngAfterContentInit(): void {
+    this.searchTerms
+      .pipe(
+        debounceTime(1000), // wait for 1s
```
tschaffter, line 156: I did some research in the past and 400 or 450 ms were recommended for the debounce time to get a responsive behavior.

```suggestion
        debounceTime(400),
```

tschaffter (approved):
I approve this PR with only a small change.

rrchai:
After committed the suggested change - [8f60b6a](https://github.com/Sage-Bionetworks/sage-monorepo/pull/1128/commits/8f60b6ac5cef2cb07e9f8c0de1628f1aca78d74e), the lint is failing I couldn't figure out why. By running the lint locally, I didn't see errors. I also tried to update the branch  with the latest upstream changes, but still failed.

```console
nx run challenge-registry-app:lint --skip-nx-cache

> nx run challenge-registry-app:lint

Linting "challenge-registry-app"...
All files pass linting.
 >  NX   Successfully ran target lint for project challenge-registry-app (5s)
```
@tschaffter Any idea?

tschaffter:
@rrchai It looks like this is a formatting issue. I think that we are using spaces for tabs, but GH may have used tabs when you accepted the suggestion (or the other way around). Could it be that you have fetched the file locally, open and save it, which then triggered the format on save and fixed the issue locally? If yes, you should see local changes that you can commit and push.

rrchai:
@tschaffter Thanks for looking into it.

It's about the formatting.

PR #1150, opened 2023-01-09
style(ui): Global styles

vpchung:
This PR will add some common CSS definitions to `_general.scss`.  For now, I have extracted styles related to the grid layout, some text, and buttons.

Note that for the styles related to `mat-button`, `!important` was required so that our custom values will be applied (and not ignored).

Issue #1149, opened 2023-01-09
Fix the appearance of the footer


tschaffter:
The footer seems to have reverted to an old/default font in the latest version of the app.

![image](https://user-images.githubusercontent.com/3056480/211423310-cc4d43c5-0660-484d-bcd5-c1d9618b2bdd.png)

Here are other potential modifications:

- Use the new name of the project (OpenChallenges)
- Remove the white margin around the footer
- Increase the size of the smallest font

tschaffter:
Tentatively assigning to this sprint before the monthly sprint review (January 31st).

vpchung:
> Remove the white margin around the footer

This surrounds the entire app (that is, it surrounds the navbar and main content as well), and I am not sure where it's coming from.  I'll look into it in a separate ticket.

vpchung:
Ah okay, found the issue!  I can add the fix to this PR?

PR #1154, opened 2023-01-10
fix(ui): Fix footer

vpchung:
This PR will address #1149:

Screen Shot 2023-01-09 at 4 33 35 PM

Ideally, this should be merged after #1150, as some styles are dependent on those defined in `_general`.


tschaffter (approved):
LGTM

rrchai (approved):
Thanks! LGTM

PR #1164, opened 2023-01-11
style(ui): Cleanup homepage

vpchung:
This PR will clean up the homepage styles and themes so that it adheres to Angular Material standards.


vpchung:
Will reopen, with a cleaner slate of commits.

PR #1167, opened 2023-01-11
feat: enable to search challenges using `challengeService`

### LINKED ISSUE #1131: Enable to filter challenges by search term with real challenge service

rrchai:
- fixes #1131 

Preview
---
Recording 2023-01-11 at 11 19 04

Notes
---
1. All filters other than "search term" are temporarily commented out - will add back as implementing the real challengeService 
2. As discussion with @tschaffter, the "elasticsearch service" might not be activated immediately when the app launched. To not confuse the users, the error message to end user is needed (using [snackbar](https://material.angular.io/components/snack-bar/overview)), as well as is needed to catch error for unsuccessful query, like below (feel free to suggest a better message):
    
    Screen Shot 2023-01-11 at 11 28 11 AM
3. The challenge search page starts using the real `Challenge` object from `api-client-angular`, but `challenge-registry-challenge-card` that used in different pages still uses `Challenge` object from `api-client-angular-deprecated`. So there are some conflicts that needed to temporarily fixed. 
    Once `Challenge` has all required properties for different pages, we should change all `Challenge` objects to use `api-client-angular` - that's for future prs.
5. `challenge-registry-challenge-card` has some styles issues which will be saved for future pr as well.


tschaffter (approved):
This looks great! Thanks for documenting this PR.

> the "elasticsearch service" might not be activated immediately when the app launched. 

When we deploy in production, we can setup a mechanism that checks that the backend is fully ready before making the app available. Unfortunately, Nx does not provide a way to define such check.

tschaffter:
The snackbar you added should anyway remain in the code in production because we want to notify the user in case the backend is experiencing an issue.

PR #1187, opened 2023-01-12
style(ui): Fix navbar

vpchung:
Addresses #1186 :

Screen Shot 2023-01-12 at 10 52 22 AM

The most notable change is updating the brand to "OpenChallenges" in the navbar as well as the page title.

I was also thinking of changing the language for signing up to "Join" to make it more... personable?  But happy to revert back if desired.

### DIFF
--- a/apps/challenge-registry/app/src/app/app-sections.ts
+++ b/apps/challenge-registry/app/src/app/app-sections.ts
@@ -3,7 +3,7 @@ import { NavbarSection } from '@sagebionetworks/challenge-registry/ui';
 export const APP_SECTIONS: { [key: string]: NavbarSection } = {
   about: {
     name: 'About',
-    summary: 'About the Challenge Registry',
+    summary: 'About OpenChallenges',
   },
   challenges: {
     name: 'Challenges',
--- a/libs/challenge-registry/ui/src/lib/navbar/navbar.component.html
+++ b/libs/challenge-registry/ui/src/lib/navbar/navbar.component.html
@@ -1,7 +1,7 @@
 <!-- TODO: figure out if the <nav> should go inside of a <header> element. -->
 <nav class="sage-navbar-header" aria-label="Top Toolbar">
   <a mat-button class="home-btn" routerLink="/" aria-label="Sage Angular">
-    <img class="logo" src="assets/images/registry.png" alt="Challenge Registry" />
+    <img class="logo" src="assets/images/registry.png" alt="OpenChallenges" />
     <span>{{ title }}</span>
   </a>
 
--- a/libs/challenge-registry/ui/src/lib/navbar/navbar.component.html
+++ b/libs/challenge-registry/ui/src/lib/navbar/navbar.component.html
@@ -23,9 +23,6 @@
     aria-label="Log In"
     >Log In</a
   >
-  <a mat-button class="sage-button" *ngIf="!isLoggedIn" routerLink="/signup" aria-label="Sign Up"
-    >Join</a
-  >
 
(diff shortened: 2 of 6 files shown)


libs/challenge-registry/ui/src/lib/navbar/navbar.component.html
```diff
@@ -24,7 +24,7 @@
     >Log In</a
   >
   <a mat-button class="sage-button" *ngIf="!isLoggedIn" routerLink="/signup" aria-label="Sign Up"
-    >Sign Up</a
+    >Join OC</a
```
rrchai, line 27: "OC" looks good to me, as long as we will mention what stands for, somewhere in the home page maybe.

apps/challenge-registry/app/src/app/app-sections.ts
```diff
@@ -3,7 +3,7 @@ import { NavbarSection } from '@sagebionetworks/challenge-registry/ui';
 export const APP_SECTIONS: { [key: string]: NavbarSection } = {
   about: {
     name: 'About',
-    summary: 'About the Challenge Registry',
+    summary: 'About OpenChallenges',
```
rrchai, line 6: English question: Do we need article here?
General question: Is our app called "openChallenges" or "Open Challenges"? - that's confused me a bit, lol.

rrchai (approved):

libs/challenge-registry/ui/src/lib/navbar/navbar.component.html
tschaffter: "OC" sounds weird to me, maybe because we haven't used this yet. Also it should be evident that the Register / Join button is about joining OpenChallenges.

Another option is also to remove this button from the navbar and only include "Log in", because a login page always has a link like "Not registered yet? Create an account".

apps/challenge-registry/app/src/app/app-sections.ts
tschaffter: The official name based on the feedback received is "OpenChallenges" (no space, PascalCase).

tschaffter (changes requested):
See replies to Rong's comments

libs/challenge-registry/ui/src/lib/navbar/navbar.component.html
vpchung: Oh whoops, "OC" was not meant to be pushed.  Sorry about that!

vpchung: > Another option is also to remove this button from the navbar and only include "Log in", because a login page always has a link like "Not registered yet? Create an account".

This sounds reasonable to me.  I'd like to talk with Adam too to see what he thinks from a UX standpoint.

vpchung:
TODO (probably in a later ticket):
* remove [Join] button from the navbar and only include "Log in", because a login page always has a link like "Not registered yet? Create an account"

libs/challenge-registry/ui/src/lib/navbar/navbar.component.html
tschaffter: Perfect, thanks!

vpchung: Just spoke with Adam -- he agrees that users will get the general flow if we take out "Join"/"Register" from the navbar!

Will implement now.

vpchung:
Final product:

Screen Shot 2023-01-13 at 9 57 07 AM

tschaffter (approved):

PR #1205, opened 2023-01-17
feat: Allow API gateway to fetch challenge input data types

tschaffter:
## Changelog

- Configure the API gateway to handle requests sent to `/api/v1/challengeInputDataTypes`


rrchai (approved):
Thanks for the fix!

Issue #1207, opened 2023-01-18
Sort by Recently Started and Recently Closed


tschaffter:
## Background

We have the following sorting strategies:

- `Relevance` (default): let the backend/Elasticsearch decide the order.
- `Recently Registered`: The way to be aware of the latest challenges that "came into existence".
- `Most Starred`: Sort challenges by "popularity".
- `Starting Soon`: Find challenges to participate in before they start.
- `Closing Soon`: Last chance to participate in a challenge before it closes.

Kaggle has additional sorting strategies, including these two that may be interesting:

- `Recently Launched`: These challenges have started and you can now participate in them!
- `Recently Completed`: Time to check the results! (there may be some delay before results are published after the end of the challenge).

@rrchai @vpchung Do you agree to add these two extra sorting strategies?

tschaffter:
The above post prompt the following question: Do you remember what was the definition of `status: completed` and how is this value set? There are two options:

- Option 1: The maintainer of a challenge is responsible for setting the `status` of a challenge. This approach has the benefit that the challenge can be marked as completed after the challenge no longer accept submission, e.g. mark the challenge as completed after the publication of the results.
- Option 2: We automatically set the status depending on the start and end date. With this approach the maintainer does not have to maintain this property. They would still be responsible for maintaining the start and end dates.

@vpchung @rrchai @chepyle Thoughts?

vpchung:
I'm not a huge fan of "Recently registered" -- it wouldn't be clear to me right away what is meant by "Registered".  Instead, can we maybe use "Recently added" or "Just added"?  I like "Recently Launched" and "Recently Closed"!

vpchung:
> The above post prompt the following question: Do you remember what was the definition of status: completed and how is this value set?

Hmm... this is hard to say.  I like the idea of Option 2, however, that does require the organizers or maintainers to not only manage the timeline on their challenge-running platform of choice, but also on OpenChallenges as well.

Then again, that argument can also be applied to Option 1.... so again, hard to say.

rrchai:
> I'm not a huge fan of "Recently registered" -- it wouldn't be clear to me right away what is meant by "Registered". Instead, can we maybe use "Recently added" or "Just added"? I like "Recently Launched" and "Recently Closed"!
+ 1
[...]

PR #1234, opened 2023-01-24
feat: enable recently started/closed sort filters using `challengeService`

### LINKED ISSUE #1226: Show all the options in the sort menu without scrolling
Remove the scroll bar and show all the items together.

image

### LINKED ISSUE #1227: Add challenge sort option `Recently Launched` and `Recently Completed` to UI
The backend is ready, see #1225

rrchai:
- fixes #1227

Other issues related to the search page in the pr:
- fixes #1226 
- updates the `totalChallengesCount` to display real total number of challenges in the database
- fixes some layout in `challenge-card`


libs/openchallenges/challenge-search/src/lib/challenge-search.component.ts
```diff
@@ -169,6 +169,12 @@ export class ChallengeSearchComponent
     //       active: false,
     //     })))
     // );
+
+    // update the total number of challenges in database
+    this.challengeService
+      .listChallenges(this.query as ChallengeSearchQuery)
+      .subscribe((page) => (this.totalChallengesCount = page.totalElements));
+
```
rrchai, line 177: @tschaffter I assume here if `pageSize` is `undefined`, then all results will be returned - but let me know if it's not correct.

vpchung: I think we should set a default of 20 maybe?

libs/openchallenges/challenge-search/src/lib/challenge-search-filters-values.ts
```diff
@@ -137,4 +137,12 @@ export const challengeSortFilterValues: FilterValue[] = [
     value: 'starred',
     label: 'Most Starred',
   },
+  {
+    value: 'recently_started',
+    label: 'Recently Launched',
+  },
+  {
+    value: 'recently_ended',
+    label: 'Recently Completed',
```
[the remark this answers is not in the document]
rrchai: ah yeah, I forgot the conversation in this [ticket](https://github.com/Sage-Bionetworks/sage-monorepo/issues/1207). I think we agreed to use "recently closed". Thanks for reminding.

libs/openchallenges/challenge-search/src/lib/challenge-search.component.ts
rrchai: To clarify, this purpose of snippet is to retrieve the total number of challenges that we have in the database and will be displayed in this element:
 
Screen Shot 2023-01-24 at 3 11 33 PM

>  I think we should set a default of 20 maybe?

There is also a `defaultQuery` where we currently use `50` as default `pageSize`, but it reminds me that the pagination of this page has been added. Do we limit how many challenge cards per line for this page?

vpchung: Mine is scaled to that too :-)

tschaffter: The size of the page is defined by the property [`size`](https://github.com/tschaffter/sage-monorepo/blob/add-kafka-admin/libs/openchallenges/api-description/src/components/schemas/PageMetadata.yaml). The value of this property cannot be undefined.

tschaffter: > the purpose of this snippet is to retrieve the total number of challenges that we have in the database and will be displayed in below element.

The total number of challenges can be obtained by sending an empty query object and then extract the value of the page property `totalElements`.

tschaffter: > Can you remind me of the size of your big screen (mine is 3840 × 2160)?

Mine too but the design should not rely on user having 4K monitor. :-)

libs/openchallenges/challenge-search/src/lib/challenge-search-filters-values.ts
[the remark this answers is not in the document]
tschaffter: @rrchai I would use "Recently Completed" to stay consistent with the challenge status `completed`.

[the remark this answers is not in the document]
tschaffter: @rrchai We should go with "Recently Completed" to stay consistent with the challenge status `completed`.

tschaffter:
> I think we should set a default of 20 maybe?

20 sounds like a good start.

libs/openchallenges/challenge-search/src/lib/challenge-search.component.ts
rrchai: > The total number of challenges in the DB can be obtained by sending an empty query object

ah, that's easier.

> Can you remind me of the size of your big screen (mine is 3840 × 2160)?
Mine too but the design should not rely on user having 4K monitors.

Cool! yeah, I agree the design will reply on breakpoints, instead of 4k - that's too large, but at least we know we are now viewing at the same size~

tschaffter (approved):

rrchai:
>@vpchung: I think we should set a default of 20 maybe?
@tschaffter: 20 sounds like a good start.

Sound great! The `pageSize` will be addressed in the paginator pr.

PR #1255, opened 2023-01-31
feat: adjust styles of `organization-card` and other pages

rrchai:
- fixes the styles of `organization-card` - reduce the size similar to `person-card`
- fixes the layout in home
- fixes the incorrect redirect url in profile pages
- fixes issues on challenge-card - get real platform and title case the difficulty 
- modify some data for testing
- fixes incentive and difficulty filters (match search query names)


tschaffter (approved):
LGTM. Thanks for using svg images and starting to use the name OpenChallenges!

PR #1256, opened 2023-01-31
feat(angular): Enable to redirect to search page after typing in home search button

### LINKED ISSUE #1236: Search challenges with the search bar on the home page
Most of the code needed for this feature should be available from the prototype.

1. The user enter search terms in the search bar.
2. The user press ENTER or click on the "Search" button.
3. The user is redirected to the challenge search page.
    - The search bar has been populated with the terms.
    - The other filters are in their default state.
    - The page shows results that match the search terms.

rrchai:
- fixes #1236 

Now, using the query url is also able to search the challenges: `/challenges?searchTerms=test`. It only supports for the full text search for now, as other query would require to update the filters as well (save for later if needed).

Preview
---
(sorry I didn't realize the texts are too small)
Recording 2023-02-23 at 13 28 46


tschaffter:
The current implementation relies on a data service to share data between the home page and the challenge search page. Instead, it's preferable to use a query parameter - here `?searchTerms="genomic"` - as we should support query params anyway with the search page so that the URL can be shared and bookmarked.

libs/openchallenges/challenge-search/src/lib/challenge-search.component.ts
```diff
@@ -167,17 +171,28 @@ export class ChallengeSearchComponent
...
   ngAfterContentInit(): void {
     this.searchTerms
-      .pipe(debounceTime(400), distinctUntilChanged(), takeUntil(this.destroy))
+      .pipe(
+        skip(1),
+        debounceTime(400),
+        distinctUntilChanged(),
+        takeUntil(this.destroy)
+      )
       .subscribe((search) => {
```
tschaffter, line 193: Could we rename the parameter `searchTerms` and then later do `{ searchTerms }` instead of `{ searchTerms: search}`?

libs/openchallenges/home/src/lib/challenge-search/challenge-search.component.html
```diff
@@ -22,10 +22,16 @@ <h1>
...
+            class="input"
+            type="search"
+            name="search"
+            placeholder="I am interested in..."
+            [(ngModel)]="searchTerm"
+          />
         </div>
         <div class="control">
-          <a mat-raised-button class="btn" color="primary" href="/search">Search</a>
+          <a mat-raised-button class="btn" color="primary" (click)="onClick()">Search</a>
```
tschaffter, line 34: Does pressing ENTER when the focus is on the search field also trigger the search? If not, can you please create a ticket as an enhancement.

I just realized that Google no longer has a "Search" button, but that does not mean we should remove ours.

tschaffter (approved):
Looks good! I made a few suggestions for improvement.

PR #1260, opened 2023-02-01
feat: Make AboutComponent and FooterComponent standalone

### LINKED ISSUE #1005: Update OpenChallenges app routes to use Ng v15 router
See the example from this [article](https://blog.angular.io/angular-v15-is-now-available-df7be7f2f4c8).

tschaffter:
Fixes #1005 

## Changelog

- Explore how to create standalone components in Angular v15.
  - Convert the FooterComponent from the UI library to standalone.
  - Convert the AboutComponent from the About library to standalone.
- Create the file `about.routes.ts`.
- Remove the file `about.module.ts`
- Import the routes of the About component in the app without using a NgModule.

## Notes

- The goal of this PR is to explore the creation of standalone components.
  - Converting all the components of the OpenChallenges app is out of scope.
  - We should aim to convert the existing components to the standalone API whenever we have the chance.
- The goal of the Angular Standalone API is to ultimately get rid of the NgModules.
  - This [article](https://blog.ninja-squad.com/2022/05/12/a-guide-to-standalone-components-in-angular/) helped me to understand the new API.
  - The Ng Standalone API will lead to more performant loading of components by loading only the components needed (NgModules does s bulk loading of components).
- A standalone component can be imported from another standalone component.
- A standalone component can NOT be imported from a non-standalone component.
  - In that case, the standalone component can be imported from the NgModule of the non-standalone component.
- Not only components but the app itself can be standalone. I have not explore this yet and will do when needed.


rrchai (approved):
LGTM

Issue #1297, opened 2023-02-07
Add "notebook" to type of submissions


tschaffter:
Most/all Kaggle challenges will use this data type (Jupyter notebooks).

vpchung:
Ah, looks like I'll need to brush up on my OpenAPI commands 😅

vpchung:
TODOs:
- [ ] add to openapi spec
- [ ] add to table schema
- [ ] update OC schema to use "notebook" instead of "other" for Kaggle challenges

tschaffter:
Useful command:
- `nx run openchallenges-api-description:build`
- `nx run openchallenges-challenge-service:openapi-generate` to update the challenge service. No more actions should be required to update the Java code as you are extending an existing Enum.

vpchung:
@tschaffter exactly what I needed - thanks for the impromptu cheat sheet!

PR #1298, opened 2023-02-08
feat(data): add `notebook` to challenge model

vpchung:
Accomplishes #1297


tschaffter (approved):
You are now a Java developer!

PR #1309, opened 2023-02-13
feat(data): update db-cli to push data to mariadb instead of mongo

### LINKED ISSUE #1147: Modify the DB CLI to push data to the project MariaDB instance
Update the DB CLI to push data to the MariaDB instance (`http://challenge-registry-mariadb:3306`) instead of MongoDB.

vpchung:
Closes #1147 

### Same command: `ping`
Now, it pings the mariadb instance instead of Mongo

Screen Shot 2023-02-13 at 3 14 58 PM

### New command: `list-dbs`

Lists the available databases (filtered by *_service)

Screen Shot 2023-02-13 at 3 12 04 PM

### New command: `list-tables <database>`

Lists the available tables of a given database (will not list `flyway_schema_history`)

Screen Shot 2023-02-13 at 3 12 47 PM

### New command: `drop-tables <database>`

Drops all tables of a given database (except for `flyway_schema_history`)

Screen Shot 2023-02-13 at 3 13 25 PM

### Same-ish command: `seed-database <database> <dir>`

Seed a given database with JSON files from given directory.

Screen Shot 2023-02-15 at 6 31 44 PM


tschaffter:
This CLI looks good. How do set credentials to connect to the DB?

This CLI is actually timely for one more reason. I'm wondering how we could use it for MariaDB container healthcheck to not only ping the DB but also test that the required tables have been created. Do you have suggestions on how we could implemented this second test?

vpchung:
> Do you have suggestions on how we could implemented this second test?

@tschaffter hmm, we could use the number of files that are located in the `data/seed` folder?  Either that or write a test that explicitly looks for the tables expected.  The only issue with the latter approach is that it'll require more maintenance, since we'll have to keep updating it while time we add/remove a table.

apps/openchallenges/db-cli/src/config.ts
```diff
@@ -6,4 +6,10 @@ export const config = {
     },
     uri: 'mongodb://localhost:27017/openchallenges',
   },
+  mariadb: {
+    host: 'localhost',
```
vpchung, line 10: @tschaffter FYI, I still couldn't connect to the db any other way than `localhost`. Do you have any additional ideas?

tschaffter:
I think that the test should live in the service itself. One way could be to override/extend the existing health endpoints or implement a custom one. We could then use curl to send a request to this new healthcheck endpoint.

apps/openchallenges/db-cli/src/config.ts
tschaffter: Did you try to substitute `localhost` by `openchallenges-mariadb`? This domain name should be defined in your `/etc/hosts` as defined [here](https://github.com/Sage-Bionetworks/sage-monorepo/blob/c91e6163bded6370eaf9c437e83370a6dcad1d86/tools/configure-hostnames.sh#L14).

vpchung: Yes, I did.

vpchung: lakdjflkja apologies, I never actually tested `openchallenges-mariadb` using the CLI -- I've been testing the credentials directly with my test TS script.

Using `openchallenges-mariadb` as the host works!

Screen Shot 2023-02-14 at 10 43 29 AM

vpchung:
> @tschaffter: How do we set credentials to connect to the DB?

It is using the same config file as before.  To set custom config values, I have re-added the CLI options to do so:

Screen Shot 2023-02-14 at 5 11 59 PM

vpchung:
For future reference:
For seeding to work properly, we will need to prefix some JSON files with an underscore, so that they will be listed and read in first.  Given that some of the tables have reference constraints, the tables without constraints will need to be populated first.

vpchung:
For future reference:
For seeding to work properly, we will need to prefix some JSON files with an underscore, so that they will be listed and read in first. Some tables have reference constraints (see [_create.txt](https://github.com/Sage-Bionetworks/sage-monorepo/blob/61440ac6ba82083a13b8cb603059b55bc71bed86/apps/openchallenges/db-cli/data/seeds/dev/challenges/_create.txt) for example), ergo the tables without constraints will need to be populated first.

tschaffter:
@vpchung One test is failing for this PR.

vpchung:
@rrchai @tschaffter okay, I think the PR is officially ready~  

Seems like a lot of file changes, but 30 of those files are related to the seed files (removed or modified) and model files (removed).  If it helps, the key files to review are:

* `app.ts`
* `database.ts`

I also attached screenshots above of the updated CLI.

apps/openchallenges/db-cli/src/app.ts
```diff
@@ -25,95 +27,172 @@ export class App {
...
 
-  private async ping(): Promise<void> {
+  private async listDatabases(): Promise<void> {
     try {
-      this.mongoose = await connectToDatabase();
-      const pong = await pingDatabase();
-      logger.info(pong ? 'pong' : 'No pong received');
+      this.conn = await connect();
+      const databases = await listDatabases(this.conn);
+      console.log('database'.padEnd(this.padding, ' ') + '\ttables_count');
```
tschaffter, line 105: Use the logger, not `console.log`. If the logger is configured to save logs to a file, then this content would not be captured.

vpchung: Ah okay, so use `logger` for all STDOUT?

vpchung: Using `logger`:
Screen Shot 2023-02-16 at 10 40 52 AM

Is this sort of output still useful? (pinging @rrchai for feedback as well)

vpchung: ^ we could do something less visual/tabular, e.g.

Screen Shot 2023-02-16 at 10 45 45 AM

tschaffter: > use logger for all STDOUT?

Yes

I like the tabular format. If we already have it we may as well keep it.

rrchai: @vpchung the first format looks good to me!

vpchung: Okay, cool!  I will keep the tabular format.  The padding has been updated to be dynamic (was originally static).

tschaffter (approved):
This is awesome work! 🚀

rrchai (approved):

PR #1323, opened 2023-02-17
feat(ui): Make footer sticky at the bottom

### LINKED ISSUE #1270: Make the footer sticky at the bottom

rrchai:
- fixes #1270 

The issue was the content of each component is not taking full height pf the page. Each page now is applied the "base" class (defined in `libs/openchallenges/styles/src/lib/_general.scss`), which is used to set proper height of content (you could ignore html changes), so that the footer will be always sticky at the bottom.

### DIFF
--- a/libs/openchallenges/about/src/lib/about.component.html
+++ b/libs/openchallenges/about/src/lib/about.component.html
@@ -1,4 +1,4 @@
-<main class="mat-typography">
+<main class="base mat-typography">
   <section class="title">
     <div class="container">
       <div class="section-outer">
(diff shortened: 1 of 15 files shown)


vpchung (approved):

libs/openchallenges/about/src/lib/about.component.html
```diff
@@ -1,4 +1,4 @@
-<main class="mat-typography">
+<main class="base mat-typography">
```
tschaffter, line 1: We should prefix all the OC classes with `oc-` to easily identify project-specific classes. This is probably work for another ticket.

tschaffter (approved):
LGTM

PR #1330, opened 2023-02-21
feat(ui): Add org search page

### LINKED ISSUE #1268: [Story] Create Organization search page component

OpenChallenges

### As a user, I want

As an organizer, I want to be able to search organizations so that I can find organizations who may be interested in supporting my challenge.

### Description

Create a project for the organization search page. Several features are bundle together in this ticket because they should be almost readily available after copy-pasting the challenge search page to create this new component.

### Acceptance criteria

- The organization search page is accessible in the app (navbar button tracked in another ticket)
- The org search page can generate a query object based on the following filters.
  - Search bar
  - 1 checkbox for the type of organization
  - sorter
- Print the query object to the console.

Out-of-scope:

- Sending the query to the backend.
- Rendering org cards.

### Tasks

- [x] Copy-paste the challenge search page
- [x] Keep only the following components:
  - Full-text search bar
  - 1 checkbox filter
  - Sort menu with only the Relevance option

### Anything else?

_No response_

### LINKED ISSUE #1269: Enable user to access the Org search page from the navbar
depends on:
- #1268 

ACs: 
- [x] organization button
- [x] redirect to the org search page

rrchai:
- resolve #1268 
- resolve #1269

Screenshot
---
localhost_4200_orgs

Note 
---
- the query/filter is not actually working until the org search is implemented


libs/openchallenges/org-search/src/lib/org-search-filters-values.ts
```diff
@@ -0,0 +1,19 @@
+import { FilterValue } from '@sagebionetworks/openchallenges/ui';
+
+export const organizationTypesFilterValues: FilterValue[] = [
```
vpchung, line 3: Are `types` the same as `roles`?

libs/openchallenges/org-search/src/lib/org-search.component.html
```diff
@@ -1,5 +1,78 @@
...
+  <section id="top">
+    <div id="breadcrumbs" class="row">
+      <a aria-label="Back to home" href="/home"><mat-icon aria-hidden="true">home</mat-icon></a>
+      <mat-icon aria-hidden="true">chevron_right</mat-icon>
+      <span>Search</span>
+    </div>
+    <div id="title" class="row">
+      <div class="col col-10">
+        <h2>Search Organization Profiles</h2>
+        <!-- TODO: add "Add Challenge" button when service is available -->
```
vpchung, line 11: Not relevenat TODO?

vpchung (approved):
LGTM!

libs/openchallenges/org-search/src/lib/org-search-filters-values.ts
rrchai: Yea, the property is not in the `Organization` yet, but I think `roles` will be more appropriate term. 

cc @tschaffter

tschaffter: Correct, that would be the contribution role (previously named organization role).

libs/openchallenges/org-search/src/lib/org-search-filters-values.ts
```diff
@@ -0,0 +1,19 @@
+import { FilterValue } from '@sagebionetworks/openchallenges/ui';
+
+export const organizationRolesFilterValues: FilterValue[] = [
```
tschaffter, line 3:
```suggestion
export const contributionRolesFilterValues: FilterValue[] = [
```

The term "contribution" gives us a bit more flexibility as this opens the door to handling organizations that contributed to challenges in a non-organization way, e.g. as participant.

tschaffter (approved):
This PR looks great! I just have one minor comment.

PR #1369, opened 2023-03-13
feat(angular): Update route

### LINKED ISSUE #1267: [Story] Update the routes of the Challenge Registry pages

OpenChallenges

### As a user, I want

As a user, I want to be taken to the desired page as what I type in the search bar of browser.

### Description

Refer to #997 

### Acceptance criteria

- [x] #1370
- [x] #1340 (backend)
- [x] Update the routes for the pages following the format `/{profile-type}/{profile-slug}`:
  - ~~`/challenge/{challenge-slug}`~~ `/challenge/:challengeId/:challengeSlug`
  - `/org/{organization-slug}`
  - `/user/{user-slug}`
- [x] Update the routes of each page to ensure the redirection to `/not-found` if invalid url/slug.

### Tasks

_No response_

### Anything else?

_No response_

rrchai:
- depends on #1439 
- fixes #1267 

This pr updates the routes of app following the format below (according to [Google doc](https://docs.google.com/document/d/10ttgzMBvJitqGMhn58O2b_kZMj7to6fr-JfxdqqjLog)):
- challenge search: `/challenge`
- organization search: `/org`
- challenge profile: `/challenge/:challengeId/:challengeSlug` (will only use `challengeId` to find the challenge regardless of whether `challengeSlug` is provided)
- organization profile: `/org/:orgLogin`
- user profile: `/user/:userLogin` (not change too much since user service is not ready yet)

---
One shared function `handleHttpError` is added to handle errors and redirect to new page based on different http status codes. For example:

```ts
// redirected destination can be defined with different http status code. 
// If no status code found, it will redirect to default url, which is '/' (home page) for now
 const errorMsg = handleHttpError(err, this.router, {
    404: '/not-found',
    400: '/org',
  } as HttpStatusRedirect);
```
---
Some examples of behaviors if invalid urls provided:
- /challenge/sdqwd => redirect to home page
- /org/ss => 400 error (bad request due to too short login => redirect to home page
- /org/sadasd => 404 not found => redirect to org search page


libs/openchallenges/challenge/src/lib/challenge-routing.module.ts
```diff
@@ -1,13 +1,13 @@
...
 import { RouterModule, Routes } from '@angular/router';
 import { ChallengeComponent } from './challenge.component';
-import { ChallengeOverviewModule } from './challenge-overview/challenge-overview.module';
 
-const routes: Routes = [
+export const routes: Routes = [
+  { path: '', component: ChallengeComponent },
   {
-    path: ':challengeName',
+    path: ':challengeSlug',
```
tschaffter, line 8: @rrchai Just to clarify, we will use the challenge ID to identify a challenge, not its slug (see Google doc).

tschaffter: It's actually the second route parameter (optional)

rrchai:
TODO:
- [ ] make the challenge slug optional
- [ ] check if the challenge route works as expected

tschaffter:
@rrchai What is your target date to complete this PR?

rrchai:
@tschaffter I will complete this pr today.

rrchai:
@tschaffter  sorry for the delays. Now the pr is ready for review.

I accidentally pushed the client updates to this pr. Rather than undo the push, I would like to clear file difference via pr #1439. (it looks like the `dpulls` not installed?)

tschaffter:
@rrchai Dpulls is no longer free to use and its pricing plans are too expensive in my opinion.

image

tschaffter:
@rrchai Actually I may be wrong. The public pricing page seems to say that there is still a free plan for open source projects, see https://www.dpulls.com/#pricing.

libs/openchallenges/challenge/src/lib/challenge.component.ts
```diff
@@ -22,46 +29,47 @@ import { SeoService } from '@sagebionetworks/shared/util';
...
+    this.challenge$ = this.activatedRoute.params.pipe(
+      switchMap((params) =>
+        this.challengeService.getChallenge(params['challengeId'])
+      ),
+      switchMap((challenge) => {
+        this.router.navigate(['/challenge', challenge.id, challenge.slug]);
+        return of(challenge);
+      }),
+      catchError((err) => {
+        const errorMsg = handleHttpError(err, this.router, {
```
tschaffter, line 61: The variable name `errorMsg` is ill-defined: I would expect an error message based on the name (string or object) but this is actually a function, so the variable name should be a verb and here be more specific, e.g. `handleError`.

tschaffter (changes requested):
The behavior works mainly as expected with the following exceptions:

- `http://localhost:4200/challenge/smc-rna` (invalid challenge ID) shows a blank page, but should be the Not Found page.
- Not part of this PR technically but it would be nice to fix the challenge card link on the challenge search page so that clicking on the card redirect to the challenge page.

libs/openchallenges/challenge/src/lib/challenge.component.ts
rrchai: @tschaffter Thanks for the advice. The `handleHttpError` function is used to 
- validate if a Error object is ApiClientError (that has status and title in the object)
- redirect to the page based on http status code
- return a new error (message), which will be emitted in the console.
```ts
export function handleHttpError(
  error: any,
  router: Router,
  httpStatusRedirect: HttpStatusRedirect,
  defaultRedirectUrl = '/'
): Error {
  const err = error.error;
  const status = error.status as number;
  const message = isApiClientError(err) ? err.detail : error.message;
  // redirect to corresponding url based on http status code
  // if no url is specified for the status code, redirect to default url
  router.navigate([httpStatusRedirect[status] || defaultRedirectUrl]);
  // return the error message
  return new Error(message);
}

```

tschaffter: I see. Then maybe rename the variable `const error` instead of `const errorMsg`.

tschaffter (approved):
LGTM, thanks!

PR #1373, opened 2023-03-13
feat: Disable linter errors generated for the API client for Angular

### LINKED ISSUE #991: Disable linter errors generated for the API client for Angular
Running `nx openapi-generate api-client-angular` will include a message about how to configure the generator to automatically post process the files generated. It will even include an example to run prettier, which is possibly what we want to do. An alternative is to run prettier directly as I do to format the files generated for the microservices (using spotless to format the files).

tschaffter:
Fixes #991 

## Changelog

- Disable linter errors generated for the API client for Angular

### Before

image

### After

image


vpchung (approved):
-insert chefs kiss-

Thanks!

PR #1428, opened 2023-03-28
feat(ui): use official OC logo in app

vpchung:
Closes #1426 :

### Navbar

Screen Shot 2023-03-28 at 2 55 42 PM

### Footer

Screen Shot 2023-03-28 at 2 55 48 PM

### About page

Screen Shot 2023-03-28 at 2 55 55 PM

PR #1388, opened 2023-03-17
feat(ui): update organization card component

vpchung:
Closes #1347, updating the current organization card component to the [proposed Figma design](https://www.figma.com/file/yAeIrg6Whc8Bqj2EnG4MWP/OpenChallenges?node-id=17-115&t=B6JAMeSXmP3B12Pg-0).

PR #1431, opened 2023-03-29
fix: remove undefined typography

vpchung:
I just noticed with the latest PR (#1428) that a test in the `schematic` app was failing due to a missing typography (which was removed with #1388).

This PR will fix that failed test.


tschaffter (approved):
Thanks @vpchung

Issue #1350, opened 2023-03-06
[Story] Brainstorm UI updates to make the app more slick by end of April


tschaffter:
### What projects is this story for?

OpenChallenges

### As a user, I want

As a developer of OpenChallenges, I want the UI to be more slick for the launch of the private preview so that the app becomes more impactful.

### Description

For each page and component, identify UI improvements that can positively contribute to the UX of the beta testers.

### Acceptance criteria

- A list of UI improvements has been identified.
- The item in the list are classified with the [MoSCoW](https://en.wikipedia.org/wiki/MoSCoW_method) for the launch of the private preview.

### Have you linked this story to a GitHub Project?

- [X] I have linked this story to a GitHub Project and set its metadata.

tschaffter:
Increased the priority of this ticket to High: the sooner we have identified UI improvements the sooner we can implement them.

vpchung:
## Weekly Update

I plan to start on this tomorrow.

vpchung:
## General

- [ ] As suggested by Stockard from the Design team, we should stick with the cooler-toned colors to be "more flexible for different people looking at it - pink tones [are] more biased".
    - [ ] Remove the pink and orange color palettes perhaps?
- [ ] Maybe remove the uppercase text transformations...? Not a big fan of the look, now that we are using Lato.
- [ ] Be consistent with text sizes used, e.g. navbar text is _tiny_ compared to the rest of the app
- [ ] Replace all "Registry of Open Community Challenges" and "ROCC" with "OpenChallenges"

## Navbar component

- [ ] Increase text size
- [ ] Increase logo size
- [ ] In the dropdown, either remove all icons, or find icons for "My Challenges" and "My Organizations"

## Footer component

- [ ] As suggested above, remove uppercase transformation from the hyperlinks


## Home page

- [ ] Add some sort of icon or avatar to the topic card - they look a little lifeless
    - [ ] ^ maybe even remove this section? (since we're not showing keywords/tags yet)
- [ ] Add image to "Are you a challenge organizer" section
    - [ ] Have the button link to something, maybe the Google sheet or a Loom video?

## About page

- [ ] In the Goals section, maybe do a table instead of cards?
- [ ] Increase text size (the difference between the headers and paragraphs is kind of comical 😆 )

## Meet the Team page

 - [ ] Fix skewed heights/widths of images
 - [ ] Add Maria as a contributor

## Search pages

- [ ] Increase text size
[...]

PR #1433, opened 2023-03-29
fix(ui): UI cleanup

vpchung:
Using #1350 as a guide, this PR includes an initial cleanup to some of the more jarring issues of the app, for example: the skewed images on the Meet the Team page.

preview

> **Note** During this exercise, I noticed we also need to do a cleanup with the SCSS sheets as well (saving that for another PR).

### DIFF
--- a/apps/openchallenges/app/src/_app-theme.scss
+++ b/apps/openchallenges/app/src/_app-theme.scss
@@ -8,7 +8,7 @@
 @include mat.legacy-core();
 
 $primary: mat.define-palette(palettes.$dark-blue-palette, 600);
-$accent: mat.define-palette(palettes.$accent-pink-palette, 400);
+$accent: mat.define-palette(palettes.$accent-purple-palette, 400);
 
 $theme: mat.define-light-theme(
   (
--- a/libs/openchallenges/about/src/lib/_about-theme.scss
+++ b/libs/openchallenges/about/src/lib/_about-theme.scss
@@ -5,13 +5,6 @@
   $config: mat.get-color-config($theme);
   $primary: map.get($config, primary);
   $accent: map.get($config, accent);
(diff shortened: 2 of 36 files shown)


tschaffter (approved):
Here are some observation:

- The challenge and organization cards have a dynamic height depending on their content. I initially thought that it would be better to have the same height so that it's easier to compare a given piece of information (e.g. challenge status) across cards when they are at the same vertical location. For now the look and feel feels OK to me, though I'm interested in the thoughts of others.
- I noticed that some challenges have an invalid slug but this is probably not related to this PR.
  - I reported this issue as a bug.
- I could not access the challenge and organization pages. I think that this is expected at this point as we are still working on the routes.

rrchai:
@vpchung  Thanks for improving the app UI.

The icon and label seems not aligned in the filters - not sure if it's just on my end

Screen Shot 2023-03-31 at 8 04 52 PM

rrchai (approved):
LGTM overall! Thanks.

tschaffter:
I confirm that the label of checkboxes and radio buttons are not aligned.

image

tschaffter (changes requested):
See comments above about misaligned labels.

EDIT: I checked and there is no misalignment on `main`.

vpchung:
Thanks, @rrchai! I noticed the same for the breadcrumbs so I made a note of it in #1350.  I'll add the facet misalignments too.

vpchung:
> [tschaffter](https://github.com/tschaffter) requested changes [3 days ago](https://github.com/Sage-Bionetworks/sage-monorepo/pull/1433#pullrequestreview-1367789717)

Oh, nevermind, I'll address them in this PR then.

vpchung:
Something weird happened on my end, so I am starting over. Closing this PR for now.

PR #1462, opened 2023-04-05
feat: update challengeCount in org card

rrchai:
Without or with challenges supported: 

Screen Shot 2023-04-05 at 3 47 30 PM

### DIFF
--- a/libs/openchallenges/ui/src/lib/organization-card/organization-card.component.ts
+++ b/libs/openchallenges/ui/src/lib/organization-card/organization-card.component.ts
@@ -16,7 +16,7 @@ export class OrganizationCardComponent implements OnInit {
   mockStars!: number;
   mockMembers!: Avatar[];
   otherMembers!: number;
-  challengesSupported!: number | undefined;
+  challengesSupported: number | undefined;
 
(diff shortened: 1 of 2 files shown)


libs/openchallenges/ui/src/lib/organization-card/organization-card.component.ts
```diff
@@ -16,7 +16,7 @@ export class OrganizationCardComponent implements OnInit {
   mockStars!: number;
   mockMembers!: Avatar[];
   otherMembers!: number;
-  mockChallengesSupported!: number;
+  challengesSupported!: number | undefined;
```
tschaffter, line 19: Is the exclamation point still required if `| undefined` is specified?

tschaffter (approved):
One minor comment

libs/openchallenges/ui/src/lib/organization-card/organization-card.component.ts
rrchai: It is not! Thanks for pointing it out.

vpchung (approved):

tschaffter (approved):

PR #1502, opened 2023-05-04
fix(data): hide orgs that act as platforms from org list

vpchung:
We are currently displaying some organizations in our Organizations Search page that are platforms rather than a challenge contributor.  For the purpose of the private preview, we will hide them for now.


tschaffter:
@vpchung Do we need to manual comment out these org each time we re-export the data to SQL files?

vpchung:
@tschaffter that's a really good point.  I meant for this to be a quick fix, but it may be worth adding a `public` field to the Google sheet in case this takes longer to address than planned.

Hang tight ; more changes to come!

vpchung:
Closing this PR.

Will open a new PR that will also address this issue.

PR #1517, opened 2023-05-09
fix(data): increase description to 1000 characters

vpchung:
## Changelog

Similar to #1507, increase the character limit for `challenge.description` from 280 to 1000

## TODO

The `challenge` data model will need to be updated to reflect the [new max limit](https://github.com/Sage-Bionetworks/sage-monorepo/blob/5a87fada924f212bc75c1d7ec44e6557b152a138/apps/openchallenges/challenge-service/src/main/java/org/sagebionetworks/openchallenges/challenge/service/model/dto/ChallengeDto.java#L191). (Ask Thomas or Maria for help)


tschaffter (approved):
I'm assuming that the item listed under TODO is completed and does not refer to future work, correct?

vpchung:
@tschaffter oh no, I have not updated the data model, as I thought we have agreed this is out of my scope.  Should I use a different header instead, maybe "Future Work" or "Out of Scope TODO"?  I just haven't gotten around to creating the ticket for the ask yet 🙃 Sorry!

vpchung:
> There is actually no changes required to be made to the Java code.

Ah, okay.  I thought we needed to update the size validation but if it's not needed, I will leave it be!

> What you were referring to was probably to update the "challenge service" so that it supports the change made to the data model.

Thank you for the clarification.

PR #1524, opened 2023-05-09
fix(ui): adjust layout and font family of text inside avatar logos

### LINKED ISSUE #1518: [Bug] Fix avatar whose "initials" are too long to fit in the circle

vpchung:
## Changelog
Update layout and text size inside the avatar logo for:
* profile pages

Before | After
--|--
237200262-e91add97-61c5-4ba6-be87-deacdc4fc5a2|Screen Shot 2023-05-09 at 4 34 26 PM

* organization cards

Before | After
--|--
Screen Shot 2023-05-09 at 4 50 34 PM|Screen Shot 2023-05-09 at 4 50 42 PM

### DIFF
--- a/libs/openchallenges/challenge/src/lib/_challenge-theme.scss
+++ b/libs/openchallenges/challenge/src/lib/_challenge-theme.scss
@@ -72,6 +72,11 @@
   .profile-featured > .mat-icon {
     font-size: 60px;
   }
+  .profile-pic {
+    div.avatar-content {
+      font: 60px / 86px 'Lato', sans-serif !important;
+    }
+  }
   @media screen and (max-width: 641px) {
     #profile-details > h2 {
       font-size: 32px;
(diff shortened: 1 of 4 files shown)


tschaffter:
@vpchung I proposed an alternative solution in #1523 that limits the length of the initials.

vpchung:
Thanks, @tschaffter -- you brought up a good point during the standup that changing the avatar text size was just a bandage, not a solution.

I have since removed the text resizing, but would still like to merge this PR in, as it specifies to use the Lato font instead of Helvetica (consistent with our theme).

tschaffter (approved):

tschaffter:
Looks good, thanks!

PR #1571, opened 2023-05-18
fix(ui): challenge card update

vpchung:
## Changelog

* Increase body height of challenge card (to better accommodate for challenges with long titles, descriptions, etc)
* Fix footer so that items are center-aligned

## Preview

Before | After
-- | --
Screen Shot 2023-05-17 at 11 19 48 PM | Screen Shot 2023-05-17 at 11 17 55 PM

### DIFF
--- a/libs/openchallenges/styles/src/lib/_general.scss
+++ b/libs/openchallenges/styles/src/lib/_general.scss
@@ -262,7 +262,7 @@ table {
 .card-title {
   margin: 15px 0 !important;
   font-weight: 700 !important;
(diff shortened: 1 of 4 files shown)


tschaffter (approved):
Looks good!

vpchung:
@tschaffter there's a bit more empty space, but I think it still looks presentable!  I just updated it so that we will display the description (first 80 characters) if a tagline is not available:

Screen Shot 2023-05-18 at 10 32 18 AM

PR #1595, opened 2023-05-27
fix(ui): Minor UI updates

vpchung:
## Changelog

* Add a `max-width` to the call-to-action section on the home page (to match with the other home sections):

**Before**
Screen Shot 2023-05-26 at 6 12 56 PM

**After**
Screen Shot 2023-05-26 at 6 10 00 PM

* Add missing background color to one of the cards on the About page:

Before | After
--|--
Screen Shot 2023-05-26 at 6 12 47 PM|Screen Shot 2023-05-26 at 6 10 13 PM

### DIFF
--- a/libs/openchallenges/about/src/lib/about.component.html
+++ b/libs/openchallenges/about/src/lib/about.component.html
@@ -88,11 +88,13 @@
       </div>
       <div class="card">
         <div class="card-header">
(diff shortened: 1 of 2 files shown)


tschaffter (approved):
Looks great!

PR #1607, opened 2023-06-03
feat: Setup DNS for `dev.openchallenges.io`

### LINKED ISSUE #1597: [Story] Setup DNS for preview.openchallenges.io

OpenChallenges

### As a user, I want

As a developer, I want to check that requests sent to preview.openchallenges.io can create the preview EC2 instance.

### Description

_No response_

### Acceptance criteria

- In a browser, navigating to preview.openchallenges.io will show a response returned by a service running on the preview instance.

### Tasks

- [ ] Read [Managing Route53 DNS with Terraform and CDK in 100 LOC or less](https://kerrywilson.dev/post/2021/08-12-managing-route53-dns-with-terraform-and-cdk/)
- [ ] Implement the solution described in the above article
- [ ] Work with Xa to check that the solution works (need changes to Sage IT AWS account)

### Anything else?

_No response_

tschaffter:
Closes #1597 

## Description

The goal of this PR is to reach the HTTP server running on the preview EC2 instance via its ALB when navigating to `dev.openchallenges.io`. The communication between the client and the ALB should be encrypted (HTTPS).

## TODO

- [x] Validate the ACM Certificate for `dev.openchallenges.io`
- [x] Update the deployment of the preview stack to create the DNS record and enable the ALB to use the certificate (?)

## Changelog

- Create a new stack named `cnb-dev`
  - Deploy the zone `dev.openchallenges.io`
  - Deploy the ACM certificate for `dev.openchallenges.io` with DNS validation
- Create the A DNS record `dev.openchallenges.io` that target the ALB
- Update the deployment script of the EC2 instance as we no longer need to build the images from source

## References

- [Resource: aws_acm_certificate (Terraform)](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/acm_certificate)

## Preview

Creation of an ACM certificate with DNS validation

image


tschaffter:
@xschildw Could you please review this PR? For context, I'm manually deploying the stack with the CDK for Terraform, i.e. the deployment is not done yet in a GitHub actions (future work). The stack is currently up and running. The app can be accessed here: pi-20230622170114537400000010-1816412503.us-east-1.elb.amazonaws.com

apps/openchallenges/infra/src/openchallenges-preview-stack.ts
```diff
@@ -106,6 +107,9 @@ export class OpenChallengesPreviewStack extends SageStack {
       previewInstance.instance.privateIp
     );
 
+    // The DNS configuration
+    const dns = new Dns(this, 'dns', previewInstanceAlb.lb);
```
xschildw, line 111: Is this going to rebuild the Dns construct every time (and tear it down if you tear down the app app stack)? I think we should move the creation of the Dns record 'preview.openchallenges.io' in this stack (and construct the DNS zone in another stack and reference it here). The zone needs to be independent of the app stack because we'll need to delegate DNS to it in the root zone (openchallenges.io in sageit).

tschaffter: I created another stack named `cnb-dev` to deploy "permanent" and shared resources like zones and certs. See more info about this new stack below.

tschaffter:
## How to create a new TF workspace

The new stack `cnb-dev` must be deployed in its own workspace, otherwise it will remove all the resources currently deployed under `openchallenges-preview`, whose workspace I first tried to reuse.

Upon creating the new TF workspace, I get the following error:

```console
cnb-dev  ╷
         │ Error: Terraform Cloud initialization required: please run "terraform init"
         │ 
         │ Reason: Terraform Cloud configuration block has changed.
         │ 
         │ Changes to the Terraform Cloud configuration block require reinitialization,
         │ to discover any changes to the available workspaces.
         │ 
         │ To re-initialize, run:
         │   terraform init
         │ 
         │ Terraform has not yet made changes to your existing configuration or state.
         ╵


0 Stacks deploying     1 Stack done     0 Stacks waiting
Invoking Terraform CLI failed with exit code 1
```

Running `terraform init` as suggested above does not work.

```console
$ terraform init
Terraform initialized in an empty directory!

The directory has no Terraform configuration files. You may begin working
with Terraform immediately by creating Terraform configuration files.
```

### Solution

- In TF Cloud, create a new workspace named "cnb-dev"
  - Add AWS credentials to the variables of this workspace
- Locally, remove the folder `cdktf.out`
- Deploy the new stack with `cdktf deploy cnb-dev`

apps/openchallenges/infra/src/openchallenges-preview-stack.ts
tschaffter: The dev DNS zone and certificate are now created as part of a separate stack named `cnb-dev` (see below).

apps/openchallenges/infra/src/dns/openchallenges-dev-dns.ts
```diff
@@ -0,0 +1,23 @@
...
+export class OpenChallengesDevDns extends Construct {
+  record: Route53Record;
+
+  constructor(scope: Construct, id: string, lb: Alb) {
+    super(scope, id);
+
+    // Create an alias record that points to the ALB
+    this.record = new Route53Record(this, 'record', {
+      name: 'dev.openchallenges.io',
+      zoneId: `Z08311133Q4NQKBJPU4ZX`,
```
xschildw, line 14: The zoneID could be exported from cnb-dev-dns and used here.

apps/openchallenges/infra/src/preview-instance-alb/preview-instance-alb.ts
```diff
@@ -54,10 +54,12 @@ export class PreviewInstanceAlb extends Construct {
       tags: { Name: `${nameTagPrefix}-preview-instance-tg` },
     });
 
-    this.listener = new LbListener(this, 'preview_instance_alb_http_80', {
+    this.listener = new LbListener(this, 'preview_instance_alb_https_443', {
       loadBalancerArn: this.lb.arn,
-      port: 80,
-      protocol: 'HTTP',
+      port: 443,
```
xschildw, line 59: Could redirect http to https at ALB

apps/openchallenges/infra/src/dns/openchallenges-dev-dns.ts
tschaffter: This is a good id. The zone id is an output of the stack `cnb-dev`. I need to learn how to access an output from another stack.

apps/openchallenges/infra/src/preview-instance-alb/preview-instance-alb.ts
tschaffter: Good idea!

apps/openchallenges/infra/src/dns/openchallenges-dev-dns.ts
tschaffter: This improvement will be addressed in a future PR #1875.

apps/openchallenges/infra/src/preview-instance-alb/preview-instance-alb.ts
tschaffter: Entering `http://dev.openchallenges.io` in the browser actually already redirects to `https://dev.openchallenges.io`.

When using curl, there is no response for `http://`:

```console
$ curl http://dev.openchallenges.io
curl: (7) Failed to connect to dev.openchallenges.io port 80 after 8 ms: Connection refused
```

We get a response when using `https://`:

```console
$ curl https://dev.openchallenges.io
<!DOCTYPE html><html lang="en"><head><link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="">
    <meta charset="utf-8">
    <title>OpenChallenges</title>
    <base href="/">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <link rel="icon" type="image/x-icon" href="favicon.ico">
...
```

tschaffter: https://docs.bridgecrew.io/docs/ensure-that-alb-redirects-http-requests-into-https-ones

tschaffter: I have added a listener to redirect port 80 to 443. Now the response from `curl` looks like:

```console
$ curl http://dev.openchallenges.io
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
</body>
</html>
```

And when telling `curl` to follow the redirection:

```console
$ curl -L http://dev.openchallenges.io
<!DOCTYPE html><html lang="en"><head><link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="">
    <meta charset="utf-8">
    <title>OpenChallenges</title>
    <base href="/">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <link rel="icon" type="image/x-icon" href="favicon.ico">
...
```

tschaffter: Reference: https://docs.bridgecrew.io/docs/ensure-that-alb-redirects-http-requests-into-https-ones

Issue #1534, opened 2023-05-11
[Feature] Collect post-challenge manuscript/publication DOIs


vpchung:
### What projects is this feature for?

OpenChallenges

### Description

As a user, I would like to learn more about a challenge and what came out of it.  One way I can do this is by reading the post-challenge manuscript/publication.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1617, opened 2023-06-07
fix(ui): DOI linkout

vpchung:
## Changelog

* update anchor element to actually link out instead of going back to home (due to `#`)
* link will not be displayed if there is no DOI
* add `doi` to SQL files (default value will be an empty string for now)
* add `doi` to the challenge service

## Preview

Recording 2023-06-08 at 15 29 27

### Remaining TODOs

* ~The challenge service will need to be modified to support the `doi` feature - to be added by @tschaffter~
* ~Replace `mockDoi` with `challenge.doi` - to be completed by @vpchung after #1534~


vpchung:
`doi` is not yet part of the data model, so we are still using a `mockDOI` variable, which is currently set to `09.1937/09219137`.  This will link out to an error page (as shown in the screen capture).  For the private preview, should we continue with this behavior or should we instead have DOI be empty?

tschaffter:
@vpchung What is your preference?

vpchung:
@tschaffter I much rather leave it empty if we can't find a real DOI.  But I also know you want to showcase what we could do, so.... ¯\_(ツ)_/¯

tschaffter:
> I much rather leave it empty if we can't find a real DOI

You are right. What we should avoid are behaviors that the users may think are bugs or use incorrect data, unless they can reasonably look like placeholder such as `12345...`.

We can either:
1. leave the field empty for now (showing the key `DOI` will be enough to let users know that these value will be available at some point) and come back to this feature later.
2. update the backend to support the DOI (update API description, update SQL files, update challenge service, update API client for angular) and close this feature.

I can modify the challenge service to support this feature. Could you update the SQL file(s)?

vpchung:
> Could you update the SQL file(s)?

Yes, I can add that.

vpchung:
@tschaffter updated!

Screen Shot 2023-06-08 at 10 30 03 AM

tschaffter:
@vpchung I updated the API description and the challenge service. Do you want to give a try to updating the API client for Angular (`nx openapi-generate openchallenges-api-client-angular`) and updating the app to use the new property of the `Challenge` object?

tschaffter (approved):

PR #1629, opened 2023-06-12
feat: No longer proxy requests to the API gateway

### LINKED ISSUE #1628: [Story] No longer proxy requests to the API gateway

OpenChallenges

### As a user, I want

NA

### Description

Send requests to the API gateway directly instead of proxying them when using the app dev server and container.

### Acceptance criteria

_No response_

### Tasks

_No response_

### Anything else?

_No response_

tschaffter:
Closes #1628

## Changelog

- Enable CORS on the API gateway (not via this PR but by modifying the remote config GH repo)
- Configure the app to no longer proxy requests to the API gateway (dev server and container)
- Set the app nginx (used in app container) to listen to port 4200 instead of 80.
- Rename the following app environment variables
  - `SERVICE_HOST` => `API_GATEWAY_HOST`
  - `SERVICE_PORT` => `API_GATEWAY_PORT`
- Remove the config `"publicHost": "http://localhost:4200"` for the task `serve` of the OC app
  - > **Note**: This config was added when I was working on Keycloak, but I don't remember why and Git blame does not provide information. I'm removing this configuration and will add it back - and document it - when needed.
- Add information about the app config to the README

## Notes

- 📢 The following variables have been renamed in `apps/openchallenges/app/[.env,.env.example}`
  - `SERVICE_HOST` => `API_GATEWAY_HOST`
  - `SERVICE_PORT` => `API_GATEWAY_PORT`
  - Each developer must applies this change to their `.env`

## TODO

- [ ] Notify developers to update their `apps/openchallenges/app/.env` (see below)


tschaffter:
## Configure the web app to proxy requests or send them directly to the API gateway

### Background

Using a proxy is a great way to bypass the CORS issues. An alternative solution is to configure the REST API server to enable CORS, but this is not possible for third-party servers so a solution that enables the use of a proxy when deploying the app with its dev server and docker is preferred.

Prozying requests adds an extra layer. Improved performance can be achieved when sending requests to our API gateway by enable CORS on the API gateway.

### When deploying the web app with its Dev server

The web app is currently proxying the requests to the API gateway to solve the CORS issue that the browser would otherwise raise. When starting the app with its dev server, the app config parameter `apiUrl` is defined in `src/config/config.json`. The proxy configuration lives in `src/proxy.conf.json`.

Initially the value of `apiUrl` is `http://localhost:4200/api` and the proxy redirects the requests to the API gateway. If we set `apiUrl` to `http://localhost:8082/api/v1` to sent the requests directly to the API gateway (without using the proxy), the browser will block requests because CORS is not enabled on the API gateway.

image

PR #1635, opened 2023-06-13
feat(fix): Add default pageSize in Org Profile

### LINKED ISSUE #1530: [Bug] Paginator not working as expected on org profile page

OpenChallenges

### Current behavior

In the Challenges tab of an organization profile page, _all_ challenge cards are displayed rather than the default 12.  If we were to adjust the number of results returned, then the correct number of cards will display:

https://github.com/Sage-Bionetworks/sage-monorepo/assets/9377970/10d7eb62-774c-4e48-a7e2-f8a09528bd88

### Expected behavior

By default, the paginator selects "12" as the number of cards to display, so I would expect only 12 cards should be shown, not all of them.

### Anything else?

_No response_

### Commit ID

_No response_

### Code of Conduct

rrchai:
- fixes #1530 
- retrieve only challenges that are relevant to the org, instead of all challenges.


vpchung (approved):

tschaffter:
Thanks!

PR #1648, opened 2023-06-15
feat(ui): sticky navbar

### LINKED ISSUE #1652: [Feature] Make navbar sticky to improve UX

vpchung:
Fixes #1652 

## Changelog
* keep navbar on top of viewport.  This will allow users to easily go to a different search page after they've done some scrolling

## Preview

**No scrolling done yet**
Screen Shot 2023-06-14 at 7 08 45 PM

**After some scrolling**
Screen Shot 2023-06-14 at 7 08 51 PM

### DIFF
--- a/libs/openchallenges/styles/src/lib/_general.scss
+++ b/libs/openchallenges/styles/src/lib/_general.scss
@@ -1,12 +1,11 @@
 @use 'libs/openchallenges/styles/src/lib/constants';
 
-html
+html {
+  margin-top: constants.$navbar-height;
+}
 body {
   height: 100vh;
   margin: 0;
-}
-html {
-  margin-top: constants.$navbar-height;
 }
 em {
   color: #00b1e5;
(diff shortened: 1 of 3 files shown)


tschaffter (approved):
This sticky navbar is a nice improvement to the UX! Thanks!

tschaffter:
@vpchung Can you please add a ticket for this PR so that it appears in the project board?

libs/openchallenges/styles/src/lib/_general.scss
```diff
@@ -1,10 +1,13 @@
 @use 'libs/openchallenges/styles/src/lib/constants';
 
-html,
+html
```
rrchai, line 3: does it work without comma between `html` and `body`?

rrchai (approved):

libs/openchallenges/styles/src/lib/_general.scss
vpchung: That brings up a good point - we don't necessarily need these defs for both `html` and `body`.

vpchung:
**Quick note**: just added some additional SCSS to account for responsiveness.  Hopefully no other errors arise from that 🤞🏼  Merging now!

PR #1649, opened 2023-06-15
fix: Disable Keycloak initialization in the OC app

### LINKED ISSUE #1644: [Task] Remove KC iframe from the OC app

OpenChallenges

### As a user, I want

NA

### Description

The iframe is used as part of the integration with Keycloak (KC). However, KC is currently not used and so the iframe fails to load, which leads to the following issue in the browser console.

image

The goal of this task is to comment out the iframe as long as KC is not used.

### Acceptance criteria

_No response_

### Tasks

_No response_

### Anything else?

cc @vpchung 

tschaffter:
Closes #1644


vpchung (approved):
LGTM - thanks!

Issue #1537, opened 2023-05-12
[Feature] Replace mock supporting organizations on the challenge profile page


tschaffter:
### What projects is this feature for?

OpenChallenges

### Description

As a prospective participant, I want to know who are supporting the organization of a challenge so that I have more information to decide on whether I want to participate to a given challenge.

As a prospective challenge organizer, I want to know what organization supported Challenge X so that I can contact them for the challenge I organize.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

tschaffter:
@vpchung pointed out that we don't currently don't have a way to filter challenges by organization (work in progress, #1403 ) and to get the orgs that contributes to a challenges. I'm marking this ticket as Blocked for now.

tschaffter:
Added to Sprint 23.06

tschaffter:
@rrchai This task is no longer blocked since we can now filter challenges by org.

vpchung:
@rrchai @tschaffter jumping into the convo here :-)

We already have a table in both the challenge and organization services that has this information available:

![Screen Shot 2023-06-13 at 2 11 04 PM](https://github.com/Sage-Bionetworks/sage-monorepo/assets/9377970/91554fed-cf6b-4bbc-875d-6a91c731d26d)

I believe the APIs just have to expose this now.

tschaffter:
@rrchai @vpchung That's right, we have the data and the challenge service partially supports this feature with the existing Java class `ChallengeContributionEntity` that links supporting organization to challenges.

The next steps are to define how we want to expose these data in the API description. One option is to add a property to the `Challenge` object that includes an array of contributors (org id, role). Another option is to provide this information via a dedicated endpoint, e.g. `challenges/{challengeId}/contributors`.

Option 1 comes the cost of extra backend processing to get challenges (even when contributors data is not needed, e.g. on the search page) and more data to transfer. For the sake of simplicity and performance, I propose to go with the second option. The API client will provide a dedicated method that we can call when needed.

@rrchai @vpchung Does one of you want to give a try adding this new endpoint to the API description?

vpchung:
A dedicated endpoint makes much more sense, imo - thanks!

I most likely won't have time to work on this until near the end of the sprint.  @rrchai , what about you?

rrchai:
@tschaffter @vpchung I can give a try to add the endpoint this sprint.

tschaffter:
[...]

PR #1663, opened 2023-06-19
feat: Add path to get the contributions to a challenge

tschaffter:
Related to #1537

## Description

This PR updates the OC API description to enable users to access contributions made to a challenge. The implementation of this change to the API description will be made in other PRs that target the upstream branch `challenge-contributions`.

## Changelog

- Add the path `/challenges/{challengeId}/contributions`
- Rename the org search query param `ChallengeContributorRole` to `ChallengeContributionRole`

## Notes

- I've been thinking about whether we should name the object `contribution` or `contributor`. The reason I prefer the former is that we could add more information about the contribution that are not directly about the contributor such as a start and end date for the contribution.
  - A source of inspiration could be [CD2H contribution data model](https://contributor-attribution-model.readthedocs.io/en/latest/info_model/classes/contribution.html).

### DIFF
--- a/libs/openchallenges/api-description/build/challenge.openapi.yaml
+++ b/libs/openchallenges/api-description/build/challenge.openapi.yaml
@@ -60,6 +60,26 @@ paths:
           $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/InternalServerError'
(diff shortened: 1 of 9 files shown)


vpchung (approved):
I like it, thanks.

PR #1674, opened 2023-06-21
feat(angular): Capture the active search parameters in the url of Challenge Search page 

### LINKED ISSUE #1299: [Story] Search page URL must capture the active search parameters

OpenChallenges

### As a user, I want

As a user, I want the URL of the challenge search page to describe my active query so that I can bookmark the page and share it with others.

### Description

To enable the user to bookmark a query on the challenge search page, the URL must describes the active query. 

### Acceptance criteria

- The URL of the challenge search page includes query parameters that describes the active query.
- The URL of the search page is automatically updated when a new query is applied.
- The search page reads the query parameter when the page loads and automatically initialize the components.

### Tasks

_No response_

### Anything else?

This story is related to this [comment](https://github.com/Sage-Bionetworks/sage-monorepo/pull/1256#issuecomment-1418340745).

### LINKED ISSUE #1505: [Bug] Challenge search page does not reset

OpenChallenges

### Current behavior

The results page does not reset after clearing the search bar:

https://user-images.githubusercontent.com/9377970/236341340-bdf0f6c0-ed01-4fa6-ae9e-7f77da5045e8.mov

This behavior is not seen with the Organizations Search page.

### Expected behavior

I expect the Challenges Search page to behave like the Organization Search page, in which clearing the search bar will display all of the results again.

### Anything else?

_No response_

### Commit ID

_No response_

### Code of Conduct

rrchai:
- resolves #1299 

### Descriptions:

The expected behavior of this pr:
- [x] When users enter `/challenge` page and select any of the filters, the search params in the url should also be updated and expected challenges are shown in the page
- [x] When users copy&paste a shareable link, i.e. `/challenge?status=active&searchterm='dream'`, the UI of filters should also immediately reflect what's in the url and expected challenges are shown in the page
- [x] The search params in the url are able to capture multiple selections or different filters in the same time. `queryParamsHandling: 'merge'` can append the params.
    ```ts
    this.router.navigate([], {
          queryParamsHandling: 'merge',
          queryParams: {
            status:'active',
          },
        });
    ```
- [x] The params should be reset to `/challenge`, if users click a challenge card (leaving to Challenge Profile page) and then click "challenges" in nav bar back to the Challenge Search page.
   - fixes #1505 
- [x] If any of start/end dates of challenges is defined in the params of url, the custom calendar range will be used when users refresh/land the page initially - this will only behave once every refresh.

### Note:
- The checkbox and dropdown filters were grouped together to make the code shorter. It takes more work to update all values with different filters in a group. For the sake of simplicity, I just split the group filters into individual filters, which is more straightforward, but of course, it creates more codes ~
- ~~The `ChallengePlatformService` does not support `searchTerm` yet, so the platform filter is going to change from a dropdown filter back to a checkbox filter~~
- The organizer filter is removed since its function will not be implemented shortly.
- One potential improvement is to open/collapse the panel of filters based on whether the query param is defined. However, it seems easier to do it when the filters are grouped together - save for a future pr if required.

### TO-DOs:
- [x] Enable the year range filters, currently, I can not make the time string in url equal to the value of filter, even though they look like the same, so the year range filter will not be selected.

     Screen Shot 2023-06-23 at 7 54 11 PM
         
     - Instead of selecting the year filter based on the params in the url, the custom range will be displayed if any of date param is defined in the url (only run once per refresh)


tschaffter:
@rrchai Is there an article or code that we can use as reference for this feature?

tschaffter (changes requested):
This PR is looking good! Here are a few notes:

- The pagination component is broken: interacting with it does nothing. The component also shows "Showing NaN to NaN of N results" (see image).
- The pagination values are part of the state of the page and should be captured in the url.
- Remove the separator (horizontal bar) below the filter "Organizations" (see image)
- This PR seems to solve #1505 so I'm marking this PR as fixing it.
- One improvement to make the URL shorter and easier to share would be to collapse multiple values for the same parameter. E.g.,
  - `status=completed&status=upcoming&status=active` could become
  - `status=active,upcoming,completed`

image

image

rrchai:
> - The pagination component is broken: interacting with it does nothing. The component also shows "Showing NaN to NaN of N results" (see image).
> - The pagination values are part of the state of the page and should be captured in the url.
> - Remove the separator (horizontal bar) below the filter "Organizations" (see image)

@tschaffter Good catches! The pagination values are meant to be captured in the url and I will look into this to fix the issues. 

> One improvement to make the URL shorter and easier to share would be to collapse multiple values for the same parameter. This could be left for another PR as it is not a must-have.

That's a great point. I can try to consolidate the params with the same name in this pr.
  - [ ] collapse the query params params by comma when updating the url
  - [ ] retrieve query params in the url by splitting with comma

rrchai:
> The ChallengePlatformService does not support searchTerm yet, so the platform filter is going to change from a dropdown filter back to a checkbox filter

Opps, this is actually implemented via #1445
- [ ] change platform filter to a dropdown filter

tschaffter:
> If any of start/end dates of challenges is defined in the params of url, the custom calendar range will be used when users refresh/land the page initially - this will only behave once every refresh.

This is totally acceptable.

tschaffter (approved):
The feature behave as expected and described. This is amazing work! Also, thank you for documenting this PR so well. Any developer who lands on this PR should find all the information they need.

PR #1711, opened 2023-06-29
feat(ui): Preserve the content space to make footer sticky at the bottom

### LINKED ISSUE #1687: [Story] Make the footer sticky again

OpenChallenges

### As a user, I want

NA

### Description

The content of a few pages is briefly empty during their loading, which leads the footer to appear just below the navbar. This behavior creates a flash that original from the light-grey background color of content section and the dark blue of the footer, then back to light-grey. This flash may impair the navigation comfort of the user.

This flash appears when navigating from the challenge search page to a challenge profile page, from the org search page to the org profile page, but not from the home page to the challenge search page. See the example shown below.

On the org search page:

image

After clicking on an org card, the org profile page starts to load. For a moment, the content is empty and the footer appears just below the navbar.

image

The page content is displayed and the footer is pushed back to the bottom of the page.

image

### Acceptance criteria

_No response_

### Tasks

_No response_

### Anything else?

_No response_

rrchai:
- fixes #1687 

Move `ngIf="organization$` to an extra inner `ng-container` to preserve the content space. The footer will always be at the bottom, even if the data is empty initially (due to delays).

>**Note**: 
> If the data is undefined after the requests are complete, the page will still be redirected to "Not Found" page. Although I haven't tested since I still have troubles loading images, I expect no flashes of height difference.

```html
 <!-- original -->
<main class="base mat-typography" *ngIf="organization$ | async as org">
  ...
<main>

 <!-- change to -->
<main class="base mat-typography">
  <ng-container *ngIf="organization$ | async as org">
  ...
  </ng-container>
<main>
```

The changes above are only applied to Challenge Profile and Org Profile pages.

### DIFF
--- a/libs/openchallenges/challenge/src/lib/challenge.component.html
+++ b/libs/openchallenges/challenge/src/lib/challenge.component.html
@@ -1,88 +1,90 @@
-<main class="base mat-typography" *ngIf="challenge$ | async as challenge">
-  <section id="profile-top" class="row">
-    <openchallenges-avatar
-        class="profile-pic"
-        [avatar]="challengeAvatar"
-      ></openchallenges-avatar>
-    <div id="profile-details" class="col">
-      <h2>
-        {{ challenge.name }}
-        <mat-icon aria-hidden="true" class="verified">verified_outline</mat-icon>
-      </h2>
-      <p class="username">@{{ challenge.slug }}</p>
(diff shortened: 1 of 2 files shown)


tschaffter (approved):
LGTM!

vpchung (approved):

PR #1765, opened 2023-07-13
feat(angular): ensure to send the proper query object to challenge/org service

### LINKED ISSUE #1397: Validate the query when searching for challenges and organizations
For the current frontend approach, any query (e.g. `{foo: "bar"}`) will be also added to the query, which might cause errors if invalid query was sent to api service.

- [ ] Add validation on the query name for `challenge-search` and `org-search` pages to capture errors while developing the app.

rrchai:
- resolves #1397 

This purpose of ticket above was trying to prevent from sending invalid properties of query, e.g. `{searchTerms: 'dream', foo: 'bar'}`, where `foo` is not valid property of `ChallengeSearchQuery`. Although the invalid property will be ignored in the backend, adding extra validation would be good as precaution for unknown errors. 

I was adding a type guard to check whether each property of query is part of `ChallengeSearchQuery` (or `OrganizationSearchQuery`) and throw an error if not. Then, I realize it's actually unnecessary. We can just specify the type for query to `ChallengeSearchQuery` or `OrganizationSearchQuery`. The error will be popped up if the query is invalid in vscode, see below:

Screen Shot 2023-07-13 at 2 03 38 PM

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,7 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+To report a security issue, please email `sagesecurity@sasebase.org` with a description of the
+issue, the steps you took to create the issue, affected versions, and, if known, mitigations for the
+issue. This project follows a 90 day disclosure timeline.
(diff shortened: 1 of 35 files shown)


tschaffter (approved):
Neat!

PR #1792, opened 2023-07-17
fix(bug): Fix key name for contribution role in the OrganizationSearchQuery

### LINKED ISSUE #1791: [Bug] The contribution filter is not working in Org Search Page

OpenChallenges

### Current behavior

The bug was reported by @tschaffter. 

Selecting any of options in Contribution Role filter will all return 244 results. 

### Expected behavior

Orgs should be filtered by contribution roles properly.

### Anything else?

The backend is working as expected and it only returns 141 orgs that are challenge organizers:

```
http://localhost:8084/v1/organizations?challengeContributionRoles=challenge_organizer

{
  "number": 0,
  "size": 100,
  "totalElements": 141,
  "totalPages": 2,
  "hasNext": true,
  "hasPrevious": false,
  "organizations": [...]
}
```

### Commit ID

_No response_

### Code of Conduct

rrchai:
- fixes #1791

### DIFF
--- a/libs/openchallenges/challenge-search/src/lib/challenge-search.component.ts
+++ b/libs/openchallenges/challenge-search/src/lib/challenge-search.component.ts
@@ -25,6 +25,9 @@ import {
   ChallengeStatus,
   ChallengeSubmissionType,
   ChallengeIncentive,
+  ChallengePlatformSort,
+  ChallengeInputDataTypeSort,
+  OrganizationSort,
 } from '@sagebionetworks/openchallenges/api-client-angular';
 import { ConfigService } from '@sagebionetworks/openchallenges/config';
 import { Filter, FilterValue } from '@sagebionetworks/openchallenges/ui';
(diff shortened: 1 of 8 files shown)


tschaffter (approved):
The patch solves the issue. In the future we should find a solution so that TypeScript highlights this error. Using the `as <model name>` does not seem to be sufficient.

rrchai:
> In the future we should find a solution so that TypeScript highlights this error. Using the as <model name> does not seem to be sufficient.

@tschaffter I agree. Instead of using `as ChallengeSearchQuery`|`as OrganizationSearchQuery` which will ignore the invalid properties in VSCodes, I have replaced them across multiple pages by using specific type to define the variables. For example:

```console
// before
 const defaultQuery = {
       challengeContributionRoles: this.selectedContributionRoles,
      } as OrganizationSearchQuery
// now
 const defaultQuery OrganizationSearchQuery = {
       challengeContributionRoles: this.selectedContributionRoles,
      }
where `selectedContributionRoles` is declared as `ChallengeContributionRole[];` 
```
If there is an invalid property, the error will occur in VSCodes, like:
Screen Shot 2023-07-17 at 8 15 42 PM


I think using `as ...` will help us to escape the typescript errors while developing in VSCodes, but we should be careful using it if we _do_ need to capture these mismatch/invalid properties of data type in the future.

Issue #1809, opened 2023-07-19
[Bug] The currently activated Python version 3.10.6 is not supported by the project (3.9.2)


tschaffter:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I'm evaluating a new version of the dev container before it's being used by developers. The default version of Python installed with `apt` got updated from v3.9.2 to v3.10.6. This leads Python projects that expect v3.9.2 to fail.

Question: But why? They should be able to install v.3.9.2 by themselves...

```console
$ nx run-many --target=prepare-python
...

    ✖  nx run schematic-api-client-python:prepare-python
       The currently activated Python version 3.10.6 is not supported by the project (3.9.2).
       Trying to find and use a compatible version. 
       
       Poetry was unable to find a compatible version. If you have one, you can explicitly use it via the "env use" command.
       Warning: run-commands command "poetry install" exited with non-zero status code

    ✖  nx run openchallenges-notebook:prepare-python
       The currently activated Python version 3.10.6 is not supported by the project (3.9.2).
       Trying to find and use a compatible version. 
       
       Poetry was unable to find a compatible version. If you have one, you can explicitly use it via the "env use" command.
       Warning: run-commands command "poetry install" exited with non-zero status code

    ✖  nx run schematic-api:prepare-python
       The specified Python version (3.10.6) is not supported by the project (3.10.0).
       Please choose a compatible version or loosen the python constraint specified in the pyproject.toml file.
       
        >  NX   ERROR: Something went wrong in run-commands - Command failed: poetry env use 3.10.0
       
          Pass --verbose to see the stacktrace.
       

    ✖  nx run schematic-notebook:prepare-python
       The currently activated Python version 3.10.6 is not supported by the project (3.9.2).
       Trying to find and use a compatible version. 
       
       Poetry was unable to find a compatible version. If you have one, you can explicitly use it via the "env use" command.
       Warning: run-commands command "poetry install" exited with non-zero status code

 ———————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 >  NX   Ran target prepare-python for 4 projects (2s)
```

### Expected behavior

- Investigate why the projects couldn't install Python v3.9.2 with Pyenv

### Are you developing inside the dev container?

- [X] I am using the dev container

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

tschaffter:
[...]

Issue #1806, opened 2023-07-19
[Story] Update the Dev Container (2023 Q3)


tschaffter:
### As a user, I want

As a Sage Monorepo Maintainer, I want to update regularly the image of the dev container so that the Developers can keep using performant and secure tools.

### Have you linked this story to a GitHub Project?

- [X] I have linked this story to a GitHub Project and set its metadata.

tschaffter:
## Ubuntu Vs. Debian

I decided to switch Debian for Ubuntu for the following reasons:

- The Ubuntu image has less [CVE](https://cve.mitre.org/cve/)s than Debian. It is often mentioned in threads that Canonical patches vulnerabilities on Ubuntu than the community does for Debian.
- Ubuntu gives developers access to newer software
- The size of the two images is about the same

<img width="1254" alt="image" src="https://github.com/Sage-Bionetworks/sage-monorepo/assets/3056480/7755b6d7-eca1-4437-b98e-4338c0607211">

<img width="1254" alt="image" src="https://github.com/Sage-Bionetworks/sage-monorepo/assets/3056480/49d6dd42-1b68-411b-800d-fa6b751636cd">

tschaffter:
## Checklist

- [ ] Can install the workspace Node.js packages with Yarn
- [ ] Can login with AWS CLI
- [ ] Can deploy the OC stack with Terraform CLI
- [ ] Can build the devcontainer image with the Devcontainer CLI

tschaffter:
## Update

Another reason to switch to Ubuntu is for a better R support as R packages are built for different version of Ubuntu but not Debian.

PR #1844, opened 2023-07-25
chore: Build new devcontainer image based on Ubuntu 22.04 (Jammy)

### LINKED ISSUE #1842: [Story] Update Dev Container to Ubuntu 22.04

OpenChallenges

### As a user, I want

NA

### Description

This is a second attempt to update the dev container image from Debian to Ubuntu. The first attempt was documented in #1806.

What has changed?

I'm work on enabling Sage Monorepo to support R in #1683. I noticed that installing packages from source was slow and could be speed up by installing the binaries from another repo, e.g. a POSIT repo. When using Debian and a POSIT repo for Ubuntu Focal (no repo available for Debian), some packages fail to install and so they must be installed from source, which add about 2-3 minutes to the setup time for this R project.

Adopting Ubuntu 22.04 would increase the DX for R developers.

### Acceptance criteria

_No response_

### Tasks

_No response_

### Anything else?

_No response_

### LINKED ISSUE #1843: [Feature] Add `renv` to the devcontainer

OpenChallenges

### Description

Downloading and installing `renv` takes about 45 seconds. The Developers and the CI workflows could save this much by providing `renv` as part of the dev container.

### Anything else?

_No response_

### Code of Conduct

tschaffter:
- Closes #1842
- Closes #1843
- Contributes to #1809

## Changelog

- Move from Debian Bullseye to Ubuntu Jammy (22.04)
- Move commands listed in `prepare-python` tasks to Bash scripts named `prepare-python.sh`
- Add `renv` to the devcontainer

## Notes

- See #1806 for a comparison between Debian and Ubuntu and the motivations for selecting Ubuntu.
- Merging this PR will build a new devcontainer image BUT this new image will not be used yet by developers. Building devcontainer images and using them is two separate operations to provide us with greater flexibility.
- The commands run by the Nx executor `nx:run-commands` uses the shell `sh` while developers use `bash`. One issue is that `pyenv` is initialized when the bash shell starts (see Devcontainer Dockerfile), which is why Nx couldn't properly use `pyenv`. The solution implemented in this PR consists in running these commands in a bash script (`prepare-python.sh`).

## How was the new Devcontainer image tested?

- Using existing GitHub checks
- Build all the Docker images inside the new devcontainer image, then
  - Start OpenChallenges and check the pages of the app
  - Start Schematic API server and test one endpoint in Swagger UI


tschaffter:
## Trivy scan

Current devcontainer image with Debian Bullseye base image released about one year ago:

```console
$ trivy image ghcr.io/sage-bionetworks/sage-devcontainer:1859079

ghcr.io/sage-bionetworks/sage-devcontainer:1859079 (debian 11.7)
================================================================
Total: 1361 (UNKNOWN: 4, LOW: 960, MEDIUM: 142, HIGH: 239, CRITICAL: 16)
...
```

New devcontainer image with updated tools:

```console
$ trivy image ghcr.io/sage-bionetworks/sage-devcontainer:testing

ghcr.io/sage-bionetworks/sage-devcontainer:testing (ubuntu 22.04)
=================================================================
Total: 270 (UNKNOWN: 0, LOW: 148, MEDIUM: 116, HIGH: 6, CRITICAL: 0)
```

The new image drastically remove vulnerabilities. There are also no longer critical CVEs.

vpchung (approved):

rrchai (approved):

apps/schematic/notebook/prepare-python.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+pyenv install --skip-existing 3.9.2
+pyenv local 3.9.2
+poetry env use 3.9.2
```
linglp, line 5: I noticed that for the `prepare-python` script in schematic, we are doing `pyenv local 3.10.0`. Should python version that we use for notebook be consistent with the code we use for schematic above?

andrewelamb:
Everything looks good to me, but I have the same question as @linglp about using different versions of python in different projects.

andrewelamb (approved):

apps/schematic/notebook/prepare-python.sh
tschaffter: No, projects may use different versions of Python. This is actually a feature of the monorepo. However, as you suggest, it would be a good practice for the projects of the same *product* (e.g. OpenChallenges, Schematic, etc.) to use the same version of Python.

PR #1849, opened 2023-07-26
fix(data): Add `createdAt` and `updatedAt` values to database

### LINKED ISSUE #1691: [Story] Capturing `createdAt` and `updatedAt` values in OC Data

OpenChallenges

### As a user, I want

_No response_

### Description

This ticket was suggested by @vpchung . The idea is to capture the values of `createdAt` and `updatedAt` properties in the OC Data and to include them to the SQL data dumps.

### Acceptance criteria

_No response_

### Tasks

_No response_

### Anything else?

_No response_

vpchung:
Fixes #1691 

## Changelog 
* Add value for `createdAt` in the `yyyy-MM-DD hh:mm:ss` format - currently using date of the GH ticket as the value (`2023-06-23 00:00:00`)
* Add value for `updatedAt` (same format as above) - timestamp set to the last edit to any of the cell values in the row
* Maintenance: convert IDs from str to int (as specified by table schema)


tschaffter:
@vpchung What timezone is used when generating the time values?

vpchung:
@tschaffter good question.  I just checked and Google Sheets uses the timezone set for your computer, so in my case, PST.

I'm currently using Google's AppScript to track when edits are made, so I'll see if it's possible to set it to the UTC timezone (is that what's preferred?).

tschaffter:
I briefly looked at the SQL specification and there doesn't seem to be a field that can store date time with timestamp. The code used to deploy the stack should set this information at some point. We will cross this bridge when we are there.

tschaffter (approved):

PR #1856, opened 2023-07-27
fix(ui): add links to Google Forms

### LINKED ISSUE #1855: Link out to Google Forms for data contributions
Following #1845, link out to the new Google Forms:

* [Adding a New Challenge](https://docs.google.com/forms/d/e/1FAIpQLSdLZ6C5ZIzLZS147bmvNm2eJ3kq_88ecHS7ADjD4KaapGhO8g/viewform) should be used on the home page (in the call-to-action section)
* [Suggest a Challenge Edit](https://docs.google.com/forms/d/e/1FAIpQLSdkgAVOuof-tE4PvCnpPbbGnQvvMhLuTkXikDXTbsCHqcMVSw/viewform?usp=pp_url) should be used on the challenge profile page
* [Suggest an Organization Edit](https://docs.google.com/forms/d/e/1FAIpQLSfQbQl11MG_5TXT7fL07YenRD3MIQYfG2lXrpo3jnh_iA4nsw/viewform)  should be used on the organization profile page

vpchung:
Fixes #1855 

## Changelog

* replace link on home page to the "Add New Challenge" google form
* add pre-filled link to the profile page footer for challenges and organizations

## Preview

Profile page update:

Screenshot 2023-07-27 at 1 39 32 AM

### DIFF
--- a/libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.html
+++ b/libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.html
@@ -76,7 +76,13 @@
     </div>
   </div> -->
   <hr />
-  <span class="mat-caption created-updated-dates"
-    >Added to the OC on {{ challenge.createdAt }} and last modified on {{ challenge.updatedAt }}
+  <span class="mat-caption created-updated-dates">
+    Added to the OC on {{ challenge.createdAt }} and last modified on {{ challenge.updatedAt }} //
+    See something missing or not up-to-date?
+    <a
+      href="https://docs.google.com/forms/d/e/1FAIpQLSdkgAVOuof-tE4PvCnpPbbGnQvvMhLuTkXikDXTbsCHqcMVSw/viewform?usp=pp_url&entry.1040159972={{challenge.slug}}"
+      target="_blank"
+      >Suggest an edit here!</a
+    >
   </span>
 </main>
--- a/libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.html
+++ b/libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.html
@@ -80,7 +80,7 @@
     Added to the OC on {{ challenge.createdAt }} and last modified on {{ challenge.updatedAt }} //
     See something missing or not up-to-date?
     <a
-      href="https://docs.google.com/forms/d/e/1FAIpQLSdkgAVOuof-tE4PvCnpPbbGnQvvMhLuTkXikDXTbsCHqcMVSw/viewform?usp=pp_url&entry.1040159972={{challenge.slug}}"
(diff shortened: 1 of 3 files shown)


libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.html
```diff
@@ -76,7 +76,13 @@ <h3 class="section-title">Hosted By</h3>
     </div>
   </div> -->
   <hr />
-  <span class="mat-caption created-updated-dates"
-    >Added to the OC on {{ challenge.createdAt }} and last modified on {{ challenge.updatedAt }}
+  <span class="mat-caption created-updated-dates">
+    Added to the OC on {{ challenge.createdAt }} and last modified on {{ challenge.updatedAt }} //
+    See something missing or not up-to-date?
+    <a
+      href="https://docs.google.com/forms/d/e/1FAIpQLSdkgAVOuof-tE4PvCnpPbbGnQvvMhLuTkXikDXTbsCHqcMVSw/viewform?usp=pp_url&entry.1040159972={{challenge.slug}}"
```
tschaffter, line 83: The challenge slug is technically mutable. This could lead to a problem where two edits are submitted for the same challenge - from potentially different persons - and the first edit asks to replace the slug.

The challenge ID immutable? If yes, it would be safer to use it.

libs/openchallenges/org-profile/src/lib/org-profile-overview/org-profile-overview.component.html
```diff
@@ -22,8 +22,13 @@ <h3 class="section-title">Organization Details</h3>
...
     </table>
   </div>
-  <hr>
+  <hr />
   <span class="mat-caption created-updated-dates"
-    >Added to the OC on {{ org.createdAt }} and last modified on {{ org.updatedAt }}
+    >Added to the OC on {{ org.createdAt }} and last modified on {{ org.updatedAt }} //
+    See something missing or not up-to-date?
+    <a
+      href="https://docs.google.com/forms/d/e/1FAIpQLSfQbQl11MG_5TXT7fL07YenRD3MIQYfG2lXrpo3jnh_iA4nsw/viewform?usp=pp_url&entry.1040159972={{org.login}}"
```
tschaffter, line 30: Same comment as above: I suggest to use the org ID. Note that the org login should definitively less mutable than the challenge slug. Yet there is still a chance that an org would change it, especially if they want one that is different from the one we picked for them for the initial record.

libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.html
tschaffter: We could still include the challenge slug or title - prepopulated - to provide submitter with a human-readable way to remember what challenge is being edited.

rrchai (approved):
Other than @tschaffter's comments, LGTM overall.

libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.html
vpchung: If only Google Forms allowed for read-only fields 😢 

> Is the challenge ID immutable? If yes, it would be safer to use it.

Good idea.

PR #1878, opened 2023-08-01
feat(UI): Set 3 challenge cards per row for MBP 13'' viewpoint.

rrchai:
- closes #1777 

## Changelog:
- Reduce the width of filter facet to make enough space allowing 3 challenge cards per row at 1440px.
- Adjust the direction of opening dropdown filter's option box to fit in a narrow space (see preview). It's a temporary UI adjustment and this [ticket](#1832 ) will address to further improve UI of dropdown filter.
- Evenly align the challenge cards in the container
- Set fixed height on the title of challenge card, so that the texts of platforms and descriptions in card will be aligned among cards

### Before (1440px width)

Recording 2023-08-01 at 10 50 53

### After (1440px width)
Recording 2023-08-01 at 10 51 46

>**Note**
>This pr is only focusing to set 3 cards for MBP 13'' and the smaller laptop was _not_ taken into account in this pr.

PR #1893, opened 2023-08-01
feat: Implement challenge contributors in Challenge Profile 

### LINKED ISSUE #1537: [Feature] Replace mock supporting organizations on the challenge profile page

rrchai:
- closes #1537 

## Changelog

- Rename `challenge-sponsors` component to `challenge-contributors`
- Implement real contributors of the challenge
- Update organization card ~~to show members (the members are only shown in the org card of `challenge-contributors`)~~
- Fixes the issue that the cards are not aligned center in home page (due to `.card-group` changed from #1878)
- Add description for "Organizers" and "Contributors" tabs by @vpchung via https://github.com/rrchai/sage-monorepo/pull/2

## Preview

Screen Shot 2023-08-02 at 7 34 28 PM


tschaffter:
@rrchai The Figma mockup (see the ticket referenced by this PR) has the name "Organizations" for the menu. I also know that we have discussed about using the terms "contributors" when implementing the backend. I feel like we should use the term "Organizations" and add a sentence that says that these organizations contributed to the organization of the challenge.

In the future, we will show the contribution type and that will make things more clear.

We can discuss this PR at the stand up meeting.

image

tschaffter:
> Sorry, I am not able to load images again :)

@rrchai Can you share the logs of the image service?

rrchai:
@vpchung The layout of cards in home should be aligned center in this pr, but lmk if it's not the case on your end.

rrchai:
As discussion, we will keep the "contributors" for now, and add description in the tab to explain the word "contributor" in a different pr.

libs/openchallenges/challenge/src/lib/challenge-contributors/challenge-contributors.component.html
```diff
@@ -0,0 +1,16 @@
+<main>
+  <div id="contributors">
+    <h3>Contributors</h3>
+    <p>List of organizations whose contributions made possible the challenge.  Contributions can range
```
tschaffter, line 4:
```suggestion
    <p>List of organizations whose contributions made possible the challenge. Contributions can range
```

tschaffter (approved):
A minor modification to increase consistency.

vpchung:
> The cards in home should be aligned center in this pr, but lmk if it's not the case on your end.

Looks good on my end! Thanks :-)

libs/openchallenges/home/src/lib/challenge-host-list/challenge-host-list.component.html
```diff
@@ -4,14 +4,12 @@
...
-    <div class="row">
-      <div class="card-group">
-        <openchallenges-organization-card
-          *ngFor="let organizationCard of organizationCards$ | async"
-          [organizationCard]="organizationCard"
-        >
-        </openchallenges-organization-card>
-      </div>
+    <div class="card-group">
+      <openchallenges-organization-card
```
vpchung, line 8: Should we add a paginator to this page?

For challenges with a long list of contributors, e.g. BraTS 2023.

libs/openchallenges/home/src/lib/challenge-host-list/challenge-host-list.component.html
```diff
@@ -4,14 +4,12 @@
...
+        *ngFor="let organizationCard of organizationCards$ | async"
+        [organizationCard]="organizationCard"
+      >
+      </openchallenges-organization-card>
```
vpchung, line 12: Relatedly, should we add some default text for challenges with no annotated organizations?

vpchung: ^ Actually, this can be discussed in a separate issue/PR.

libs/openchallenges/home/src/lib/challenge-host-list/challenge-host-list.component.html
rrchai: Good point! The paginator could be added here, but it will require more work, especially if the param in the url is also needed to reflect the pagination of contributors.

As the profile page is going to be transformed to single page, using the same approach as "Stargazers" to have a "Show More" button seems easier to me. Brats has 35 total, which is not too bad. 

Anyways, hope there's no challenge has tons of contributors before we address it in a different pr XD.

libs/openchallenges/home/src/lib/challenge-host-list/challenge-host-list.component.html
rrchai: (Currently the comment left on the home page's html) Do you mean the scenario when the challenge has no contributors at all and default message is needed in `challenge-contributors`? 

Although, I think ^ is very rare. I agree to have some info to notify user it's not a bug :) for future work~

vpchung: > have some info to notify user it's not a bug

Yes! That was my thinking.  The first challenge I looked at had an empty page so I thought it was a bug xD

rrchai: Sounds good. I can add the default text after this pr is merged. Thanks for the suggestion.

PR #1908, opened 2023-08-02
feat(UI): Open and reorder challenge filters 

### LINKED ISSUE #1831: [Enhancement] Revisit facets to highlight on Challenge Search page
## Sage Design feedback

Current list of facets on Challenge Search:

Image

By default, "Challenge Year", "Status", and "Input Data Type" are open, which assumes that these facets will be used the most frequently.  However, _will_ "Challenge Year" be used as often as the other facets? i.e. is it expected that users will want to look at older challenges when they first visit OC?  Or will they be more interested in the ones currently running?

What about "Incentive Type" or "Submission Type" instead?

rrchai:
- closes #1831

## Changelog
- expand all challenge filters initially 
- reorder challenge filters: 
  1. Status (I see it as the most frequently used filters for universal users)
  2. Challenge Year (useful for exploring purpose)
  3. Input Data Type (one of the most interested filters)
  4. Platform (useful for organizers, as well as the users come from/attracted by specific platforms) 
  5. Incentive Type (I see it'll be used often since reward is always attractive initially for users)
  6. Organization (likely, only organizers or people come with organization names will use)
  7. Submission Type (I don't see it will be used frequently for users)
  8. Category (move to the last since it now only has one option)

## Preview
Screen Shot 2023-08-02 at 6 32 34 PM

### DIFF
--- a/libs/openchallenges/challenge-search/src/lib/challenge-search-filters.ts
+++ b/libs/openchallenges/challenge-search/src/lib/challenge-search-filters.ts
@@ -38,21 +38,21 @@ export const challengeSubmissionTypesFilter: Filter = {
   query: 'submissionTypes',
   label: 'Submission Type',
   values: challengeSubmissionTypesFilterValues,
-  collapsed: true,
+  collapsed: false,
 };
 
 export const challengeIncentivesFilter: Filter = {
   query: 'incentives',
   label: 'Incentive Type',
   values: challengeIncentivesFilterValues,
-  collapsed: true,
+  collapsed: false,
 };
 
 export const challengePlatformsFilter: Filter = {
   query: 'platforms',
   label: 'Platform',
   values: challengePlatformsFilterValues,
-  collapsed: true,
+  collapsed: false,
 };
 
 // dropdown filters
@@ -68,14 +68,14 @@ export const challengeCategoriesFilter: Filter = {
   query: 'categories',
   label: 'Category',
   values: challengeCategoriesFilterValues,
-  collapsed: true,
+  collapsed: false,
 };
 
 export const challengeOrganizationsFilter: Filter = {
   query: 'organizations',
   label: 'Organization',
   values: challengeOrganizationsFilterValues,
-  collapsed: true,
+  collapsed: false,
   showAvatar: true,
 };
 
@@ -83,6 +83,6 @@ export const challengeOrganizatersFilter: Filter = {
   query: 'organizers',
   label: 'Organizer',
   values: challengeOrganizersFilterValues,
-  collapsed: true,
+  collapsed: false,
   showAvatar: true,
 };
--- a/libs/openchallenges/challenge-search/src/lib/challenge-search.component.html
+++ b/libs/openchallenges/challenge-search/src/lib/challenge-search.component.html
@@ -35,6 +35,19 @@
   <section id="bottom" class="content">
     <div class="facets">
       <p-panel
+        header="{{ statusFilter.label }}"
+        [toggleable]="true"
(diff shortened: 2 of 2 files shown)


tschaffter:
@rrchai @vpchung The input data type is virtually empty. Paul also suggested that we use a control vocabulary - I mentioned EDAM - for the values. I'd be in favor of hiding this filter until we had a chance to revisit it.

vpchung:
> Category (move to the last since it now only has one option)

Agreed.

> Challenge Year (useful for exploring purpose)

I'd actually argue that Challenge Year might be one of the ones that users may not used as often as we thought.  I suggest moving it down.

rrchai:
>The input data type is virtually empty. Paul also suggested that we use a control vocabulary - I mentioned EDAM - for the values. I'd be in favor of hiding this filter until we had a chance to revisit it.

Sound reasonable to me.

rrchai:
>I'd actually argue that Challenge Year might be one of the ones that users may not used as often as we thought. I suggest moving it down.

@vpchung You have successfully convinced me. Do you suggest to move it down down down below "Submission Type" or just below maybe "Incentive Type"?

vpchung:
Maybe we could have the checkboxes together, then challenge year?  Ah but then again, I think Input Data Types is a good one to have earlier than later so I wouldn't want to place that one so far down the facet list.....

tschaffter (approved):
The order is much better. If the goal of the Category filter is to highlight some challenges, then we should move this filter higher in the list - e.g. in 2nd or 3rd position - instead of having it last.

rrchai:
## Updated order:
1. Status
2. Incentive Type 
3. Input Data Type
4. Platform
5. Organizations
6. Challenge Year
7. Submission Type
8. Category 

@tschaffter I still keep input data type, as I agree we should show the filter sooner than later. If we decide to use EDAM ontologies, I suggest we can replace current data types to the terms used in EDAM and show minimal examples , i.e these sequence types below:
> Screen Shot 2023-08-03 at 10 16 15 AM
> 
> https://bioportal.bioontology.org/ontologies/EDAM/?p=classes&conceptid=http%3A%2F%2Fedamontology.org%2Fdata_2044

@vpchung I put all dropdown filters together in the middle. Would you prefer to former layout that dropdown and checkbox are mixed or all kind of group similar filters, like below?
Screen Shot 2023-08-03 at 10 09 38 AM

tschaffter:
> I still keep input data type, as I agree we should show the filter sooner than later.

I don't agree with this: the filter is currently useful as we have:

- 0 challenge for https://dev.openchallenges.io/challenge?inputDataTypes=gene-expression
- 2 challenges for https://dev.openchallenges.io/challenge?inputDataTypes=genomic
- 1 challenge for https://dev.openchallenges.io/challenge?inputDataTypes=metabolomic
- 1 challenge for https://dev.openchallenges.io/challenge?inputDataTypes=proteomic

The values in the filter are not comprehensive and the challenge data we have are also not comprehensive. Unfortunately we won't have very limited time before next week to revisit the values of the filter and collect more data.

The results of the above searches would probably reflect poorly on OC. Hiding the filter would be the easier solution. The other is to provide a message that explains why this filter is currently not useful.

vpchung:
> I don't agree with this: the filter is currently not useful as we have:

That is true - we don't currently have the values or annotations yet that would make this facet useful.  That being said, we should continue to keep in mind that input data type was repeatedly brought during the listening tours.  So, perhaps we should just keep in mind that many users may potentially use "Input Data Type" more than the others (once we get more data on that).

tschaffter (approved):
LGTM!

PR #1910, opened 2023-08-02
feat(ui): Modify the default organization sorting to be based on the `challenge_count`

### LINKED ISSUE #1870: [Feature] Rank orgs by number of challenges by default

OpenChallenges

### Description

Paul suggests that we sort orgs by their challenge count by default. During the call, he said that the orgs listed on top of the page when sorting by their challenge count were more relevant than the orgs displayed first when sorting by Relevance.

### Anything else?

The sort By Relevance currently has a bug that makes the sorting what seems to be not deterministic. See #1651.

### Code of Conduct

rrchai:
- closes #1870

## Changelog
- Change the default sort for orgs to `challenge_count` in `org-search`
- Sort challenge contributors alphabetically
  - The sorting directly from backend is currently not achievable, because the `ChallengeContribution` service has not supported to get a list of sorted contributions (array of `orgId`), which later is used to query orgs via `getOrganization()`
- Fix a bug that UI of the dropdown value is not updated with `defaultSortedBy`

#### Org Search
Screen Shot 2023-08-02 at 8 12 01 PM

#### Challenge Contributors (Brats 2023)
Screen Shot 2023-08-03 at 1 53 26 PM


libs/openchallenges/challenge/src/lib/challenge-contributors/challenge-contributors.component.ts
```diff
@@ -61,6 +62,12 @@ export class ChallengeContributorsComponent implements OnInit {
       );
   }
 
+  private sortOrgs(orgs: Organization[]): Organization[] {
+    return orgs.sort(
+      (a, b) => (b.challengeCount || 0) - (a.challengeCount ?? 0)
```
vpchung, line 67: Nit: on the challenge profile page, could we sort the organizations alphabetically instead?

vpchung (approved):

libs/openchallenges/challenge/src/lib/challenge-contributors/challenge-contributors.component.ts
rrchai: >Nit: on the challenge profile page, could we sort the organizations alphabetically instead?

Sure thing!

libs/openchallenges/challenge/src/lib/challenge-contributors/challenge-contributors.component.ts
```diff
@@ -61,6 +62,10 @@ export class ChallengeContributorsComponent implements OnInit {
       );
   }
 
+  private sortOrgs(orgs: Organization[]): Organization[] {
+    return orgs.sort((a, b) => (a.name || '').localeCompare(b.name || ''));
```
tschaffter, line 66: @rrchai I don't understand this. Are you doing a local sorting instead of relying on the backend sorting?

rrchai: I don't think the backend has supported to get a list of sorted contributions?

The single org request is used here after we got a list of org ids from `ChallenegContribution` service.

tschaffter: I understand now. This note should be added to the first post of this PR to provide some context.

tschaffter (approved):

PR #1913, opened 2023-08-03
feat: Add /nodes/{nodeLabel}/dependencies endpoint

andrewelamb:
Depends on https://github.com/Sage-Bionetworks/sage-monorepo/pull/1903
fixes [fds-782](https://sagebionetworks.jira.com/browse/FDS-782)

This PR:

- Adds the /nodes/{nodeLabel}/dependencies endpoint (formerly  /explorer/get_node_dependencies)
- Adds the node, nodesPage, and nodeLabel schema
- Adds the component and componentsPage schema (not needed for this PR, but kept for future ones)
- Adds returnDisplayNames and returnOrderedBySchema parameters

Files to Check:

- Everything in libs/schematic/api-description/src/ 
- Everything in apps/schematic/api/schematic_api/test/
- apps/schematic/api/schematic_api/controllers/schema_controller_impl.py

PR #1903, opened 2023-08-02
feat: Add /components/{componentLabel}/attributes endpoint

andrewelamb:
Fixes [fds-781](https://sagebionetworks.jira.com/browse/FDS-781)

This PR:

Adds the /components/{componentLabel}/attributes endpoint (previously /explorer/find_class_specific_properties)
Adds parameters for schemaUrl and componentLabel
Adds schemas for AttributesPage, Attribute, SchemaUrl and ComponentLabel

PR #1918, opened 2023-08-03
feat: add /nodes/{nodeLabel}/validationRules endpoint

andrewelamb:
Depends on https://github.com/Sage-Bionetworks/sage-monorepo/pull/1903
Depends on https://github.com/Sage-Bionetworks/sage-monorepo/pull/1913
fixes [fds-821](https://sagebionetworks.jira.com/browse/FDS-821)

This PR:

- Adds the /nodes/{nodesLabel}/validationRules endpoint (formerly /explorer/get_node_validation_rules)
- Adds validationRule and ValidationRulesPage schemas
- Adds the nodeDisplay parameter

Files to Check:

- Everything in libs/schematic/api-description/src/ 
- Everything in apps/schematic/api/schematic_api/test/
- apps/schematic/api/schematic_api/controllers/schema_controller_impl.py


andrewelamb:
@GiaJordan This PR is ready for you to review.

libs/schematic/api-description/src/components/schemas/ValidationRule.yaml
```diff
@@ -0,0 +1,11 @@
+type: object
+description: A validation rule.
+properties:
+  name:
+    type: string
+    description: The name of the rule.
```
GiaJordan, line 6: Kind of small but could you update the description (or the higher level description) to include that the rule arguments are included as well? In schematic the rule and the arguments are treated as separate parts in a way and I think it would prevent confusion in the future to be clear about what's included.

libs/schematic/api-description/src/paths/nodes/@{nodeDisplay}/validationRules.yaml
```diff
@@ -0,0 +1,19 @@
+parameters:
+  - $ref: ../../../components/parameters/path/nodeDisplay.yaml
+get:
+  tags:
+    - Schema
+  summary: Gets the validation rules associated with a given node
```
GiaJordan, line 6: Here again, we should also specify that the rule arguments are returned as well if that's the case

GiaJordan (changes requested):

GiaJordan:
@andrewelamb  Yep!

GiaJordan (approved):

PR #1922, opened 2023-08-03
feat(ui): Add default text when array is empty in profile tabs

rrchai:
## Changelog

- Add default 'Not available' when there are no content in profile tabs

## Preview
Screen Shot 2023-08-03 at 5 00 08 PM


rrchai:
I just added simple word 'Not available' (same approach @vpchung added for challenge details) to indicate there's no content is not because of bugs. Feel free to suggest better phrase if any.

tschaffter (approved):

PR #1932, opened 2023-08-04
feat(ui): Adjust layout of cards

### LINKED ISSUE #1928: [Bug] Challenge search page: inconsistent space for 2 cards per row

OpenChallenges

### Current behavior

Here is an example with a correct layout:

image

When selecting the "upcoming" challenges (2 results), the second card is not at the same location as the second card in the above example.

image

This issue may also affect the organization search page.

### Expected behavior

See above

### Anything else?

_No response_

### Commit ID

a0a3c57d9e81df9ce9a68abdda667fe71c6d3f3c

### Code of Conduct

rrchai:
- fixes #1928 

## Changelog
- Replace `auto-fit` by `auto-fill` in grid layout. `Auto-fill` will actually remain the layout and not fill the empty grid (which confused me by its name)
- ~~Make cards stretchable a bit to better fit in the container~~

## Preview

>**Note**
>The cards have remained to use fixed width as @tschaffter suggests.

Recording 2023-08-04 at 12 58 28
Screen Shot 2023-08-04 at 12 57 29 PM

Recording 2023-08-04 at 12 59 53

### DIFF
--- a/libs/openchallenges/challenge/src/lib/challenge-organizers/challenge-organizers.component.scss
+++ b/libs/openchallenges/challenge/src/lib/challenge-organizers/challenge-organizers.component.scss
@@ -3,5 +3,5 @@
 }
 
 .card-group {
-  grid-template-columns: repeat(auto-fit, minmax(240px, 23%));
+  grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
 }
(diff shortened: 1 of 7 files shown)


tschaffter:
@rrchai I don't think that we should use a dynamic width for the cards. This makes it harder to ensure that the cards always look good.

rrchai:
@tschaffter the cards are not changed back to use fixed width.

vpchung (approved):

tschaffter:
> the cards are not changed back to use fixed width.

the card are "now" changed back... ?

rrchai:
@tschaffter excuse my typos 😀

tschaffter:
This PR improves the problem reported, but we could still improve the layout when the page is resized (see below with a large gap between the two columns of cards). Let's discuss this together after Sage Week.

image

tschaffter (approved):
See above comment

PR #1935, opened 2023-08-04
feat(ui): Add timeline visualization

### LINKED ISSUE #1873: [Feature] Add Graphical timeline of all challenges to the Home page

OpenChallenges

### Description

The motivation is to explain why tracking challenges are relevant => large increase in the number of challenges over time, more difficult to track, hence why we created OpenChallenges.

One plot that Paul suggested is a bar plot over time where the height of the bar show the number of challenges organized in Year Y. If we can make the plot dynamic, we could show the fraction of challenges that are active, completed and upcoming.

### Anything else?

_No response_

### Code of Conduct

rrchai:
- closes #1873 

  - depends on #2019 

### Changelog
  - Add the timeline visualization to showcase how many cumulative challenges started over time

### TO-DO
 - [x] Review with team about the texts used in the plot
 - [x] Add back the origin statistic metrics (total challenges/orgs)

### Preview
Recording 2023-08-30 at 20 08 31

So far the animation takes about 4s~ to finish on my end. If user don't read the first paragraphs on the home page and go straight to the plot, they can still see the animation :). I can explore if there's option not to render the plot until user see it, or better delay the animation until the user scroll down to the plot - which will be a future enhance in another pr.


tschaffter:
Copy-pasting content I wrote on Slack:

I gave some thoughts about the plot for the home page. For now, let's move on with your implementation that shows "number of challenges that started a given year" for the sake of announcing more broadly OC sooner than later. Instead of a line, I propose to show a vertical bar plot that will fill more space on the home page. If there is such a thing as an animation where the bars raise from y=0 one by one, that would be cool (but optional).

Ultimately, I propose to report the "number of challenges active a given year" instead of the "number of challenges that started a given year". This should give raise to an interesting bar plot that uses three colors for upcoming, active and completed challenges. The plot would give the following information:
- A tail that shows the fraction of challenges organized the years Y (this year), Y-1 (last year), ... that are still active this year.
- Fraction of upcoming challenges that are scheduled to start this year but haven't started yet.

package.json
```diff
@@ -14,6 +14,7 @@
   },
   "private": true,
   "dependencies": {
+    "@amcharts/amcharts5": "^5.4.4",
```
tschaffter, line 17: @rrchai The version should be pinned. See https://docs.renovatebot.com/dependency-pinning/ for more info.

rrchai: I will cange to pin the package and run `yarn install` to update lock file.

rrchai:
Quick update: I have changed to use [AMCharts](https://www.amcharts.com/) instead of [ECharts](https://echarts.apache.org/en/index.html) and have implemented an example plot (including both line and bars). 
I will resume working on this pr later today and tomorrow, and provide you more information about the plot.

rrchai:
>Can you give a brief description of why you want to use AMCharts?

@tschaffter I am still in exploring stage for both libraries 😃, and AMCharts seems to have better default visualization in terms of themes and layout (at least visually better for me) - so it require less codes to improve styles. 

> the advantage of ECharts is that it's open-source

I thought AMCharts is also free and open-source, and also provides product service? As I experienced both codes for the line/bar plots, I feel the `ECharts` codes are easy to use/read, so I probably will change back to use `ECharts` eventually ~  I will keep you posted today.

rrchai:
@tschaffter 
I am changing back to use `ECharts` over `AMCharts` due to more readable codes and more docs of the library. (Note: `ECharts` would require to increase our current bundle size limit)
Regarding the static data needed from service for now, the total challengeCounts per year is sufficient. Thanks.

tschaffter:
> ECharts would require to increase our current bundle size limit

@rrchai Could you please provide more info on this? What is the actual and expected bundle size?

rrchai:
As a record, it was one design to show total cumulative challenges per year (line), as well as how many challenges for each status per year.
Recording 2023-08-30 at 14 33 10

This pr will only focus on to make a total challenge per year bar plot.

rrchai:
### Cumulative challenges count per year 
Screen Shot 2023-08-30 at 2 54 49 PM

### Total challenges started per year
Screen Shot 2023-08-30 at 3 51 10 PM

tschaffter:
Here is how to use Plotly, which could have been an option too?

- https://medium.com/@davidjohnakim/using-plotly-js-with-angular-7fc07ca9a48f
- https://github.com/plotly/angular-plotly.js/blob/master/README.md

Thanks @sagely1 for sharing this!

So far EChart looks great for our use case!

tschaffter:
## Review

- I would remove the sub-title "The Rise of Challenges"
- The font used by the plot (axis, tick values) should be a bit larger
  - We have similar font size on the challenge page for example but the color is black, which makes it easier to read.

image

rrchai:
### Updated Version

Recording 2023-09-07 at 12 57 18

package.json
```diff
@@ -35,12 +35,14 @@
...
     "constructs": "10.2.13",
     "core-js": "3.25.5",
+    "echarts": "^5.4.3",
     "keycloak-angular": "12.1.0",
     "keycloak-js": "19.0.3",
     "lodash": "4.17.21",
     "mariadb": "^3.0.2",
     "ngx-avatars": "1.5.0",
     "ngx-countup": "13.1.0",
+    "ngx-echarts": "^16.0.0",
```
tschaffter, line 45: Same commend as above: `yarn add -D ngx-echarts@16.0.0`

tschaffter (approved):
The plot looks great!
I left two minor comments

PR #1946, opened 2023-08-07
feat(ui): Convert profile pages to single-page layout

### LINKED ISSUE #1833: [Enhancement] Switch to one-page layout for Profile pages
## Sage Design feedback

Current profile pages:

Image

The initial thought was that the sidebar was a _table of contents_ rather than an actual navigation system.  It could be worth to switch to a one-page layout then re-purpose the sidenav as a table of contents?  Especially considering that each of the sub-pages are not going to be long (with the exception of "Stargazers").

vpchung:
Fixes #1833 

## Changelog
* Remove tabs
* Move all contents into a single page
* Sticky the TOC once scrolled

## Preview

**Challenge Profile, e.g.**
single-page-layout


vpchung:
A nice-to-have would be to change the active state of the TOC based on scroll, but I haven't added that yet.

I found [this component](https://github.com/zurfyx/angular-contents) that we could use in-place of the TOC I implemented, which will have the scrolling active states. It just hasn't been maintained in a while.

vpchung:
A nice-to-have would be to change the active state of the TOC based on scroll, but maybe this can be saved for another PR?

I found [this component](https://github.com/zurfyx/angular-contents) that we could use in-place of the TOC I implemented, which will have the scrolling active states. It just hasn't been maintained in a while.

tschaffter (changes requested):
> A nice-to-have would be to change the active state of the TOC based on scroll, but maybe this can be saved for another PR?

The current implementation would likely confuse users and prompt users to report a bug.

> I found [this component](https://github.com/zurfyx/angular-contents) that we could use in-place of the TOC we implemented, which will have the scrolling active states. It just hasn't been maintained in a while.

This package offers what we need, but it relies an a largely outdated version of Angular (> 2 years). If no alternative to this package can be found, we could add the code of the package mentioned above to our code base and update it so that it depends on the version of Angular used in the monorepo. There isn't much code to import so that looks reasonable to me.

## How to manage paginated content like Stargazers?

We should ideally clarify this design question before moving to single-page layout. If no paginated content is expected to be added within 6 months, then it may be fine to move ahead with the single-page layout.

vpchung:
> How to manage paginated content like Stargazers?

I thought @rrchai had suggested a plan on how we would handle the stargazers content? But I can convert this to a draft in the meantime.

vpchung:
There's been a lot of changes since this was opened.

Will close this for now; we can revisit this if/when we decide to change the layout.

PR #1950, opened 2023-08-09
feat(ui): add text + link for truncated profile descriptions

### LINKED ISSUE #1835: [Enhancement] Add "See More" to truncated descriptions
## Sage Design feedback

An example long description on profile pages:

Image

Where is the rest of the description? As a potential user, it's not clear that the intention of the truncated description is to go to the challenge website to read the rest of the description.

Consider adding some text that says: "(see more on challenge website)"

vpchung:
Fixes #1835 

## Changelog
* Add "Continue reading on ..." text for truncated descriptions
* Add MatIconModule to overview pages 
* General styling

## Preview
**Profile with description <1000 characters**
Screenshot 2023-08-09 at 2 59 42 AM

**Profile with description >1000 characters)**
Screenshot 2023-08-09 at 2 59 47 AM


tschaffter (changes requested):
- The text implies that the rest of the paragraph can be read on the challenge website. Is that always the case?
- The gray color for text should be used only for disabled elements and missing data (used recently). Having a gray link may be confusing.
- The icon should be centered vertically.

rrchai:
>The gray color for text should be used only for disabled elements and missing data (used recently). Having a gray link may be confusing.
>The text + icon is a bit too long. Can we make it shorter?

Agreed. "redirect icon" might be enough for users to understand it will be redirected to external website.

tschaffter:
> I would think so, unless you have come across an edge case for this.

If we allow external users to submit challenge descriptions like pFDA is going to do, there is no guarantee that the challenge description on OC is an excerpt that the user can continue reading on the official challenge website, in which case "continue reading" maybe not be possible.

vpchung:
@tschaffter hm. In that case, we can just be vague and just say "More details on challenge site" instead. Unless you or @rrchai have other ideas on how we can more extensive to the special cases?

tschaffter:
> we can just be vague and say "More details on challenge site" instead. 

Let's take a step back. The description of the challenge can be either:

1. Text extracted from an external source.
  a. The description has less than 1000 chars (current limit)
  b. The description has more than 1000 chars
2. Text crafted specifically for OC that meets the character limit.

In general, I believe that the user understand that they can click on the "Go to Website" button to learn more about the challenge. As such, adding a link at the end of the description that reproduce the "Go to Website" button is not required for the cases 1a and 2.

When scrapping a website - manually like we do today and later programmatically - we will extract data from sections that can easily include more than 1000 chars. Our current approach which I think is a good one is to take the first 1000 chars or so and add an ellipsis. In this case (1b), it is common to [include a “Read More” And “Continue Reading” link](https://www.smashingmagazine.com/2009/07/designing-read-more-and-continue-reading-links/).

After looking at the above article, "Read more" and "Continue reading" seem pretty standard. Compared to previously proposed solutions, they are also shorter and contribute to make the page content lighter.

"Read more" and "Continue reading" are almost equivalent. The main difference is that "Continue reading" implies that the user can resume the reading on the external site, which would be true if the content come from the external site and is up-to-date.

## Proposal

If the description ends with "...", append the link "Continue reading <external redirection icon>" that redirects to the external website. The icon is to prevent the case where the user would expect extra content to show on the OC challenge/org page when clicking on the link. The icon should be vertically centered with the text.

vpchung:
## Version 2

Screenshot 2023-09-18 at 3 45 59 PM

tschaffter (changes requested):
- The text is too small: small texts are more difficult to read and click.
- The icon should be vertically centered with the text: the current style makes the height of the last line different from the heights of the other lines of the paragraph.

rrchai:
Screen Shot 2023-09-19 at 11 15 38 AM

It seems the line height is not enough - the "g" is overlapped with underline.

I may be overly picky. I now feel both capital titled "Continue Reading" or both lowercase "continue reading" is a bit more visually pleasing than "Continue reading" :)

tschaffter:
> I may be overly picky. I now personally feel both capital titled "Continue Reading" or both lowercase "continue reading" is a bit more visually pleasing than "Continue reading" :).

This [page](https://www.smashingmagazine.com/2009/07/designing-read-more-and-continue-reading-links/) includes several examples that may help making the decision.

Given that the text is very close to the paragraph, capitalizing both words may help to distinguish them form the rest of the paragraph. In the above examples, the link is often placed below the paragraph, in which case capitalizing the first letter would probably be fine.

vpchung:
> The text is too small: small texts are more difficult to read and click.

Ah, I originally changed the text size in order to differentiate it from the description.  But perhaps the color change is sufficient enough.

vpchung:
> It seems the line height is not enough - the "g" is overlapped with underline (I initially thought it was a period at the end).

Not sure if this is something we can control with the underline text-decoration?  I could perhaps do a workaround by using a border-bottom instead.

rrchai:
> Not sure if this is something we can control with the underline text-decoration?

`text-underline-offset` can do the trick, like `text-underline-offset: 3px`, but I don't think it's a must for now since it will need to adjust entire root to make underline styles consistent.

libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.scss
```diff
@@ -4,3 +4,7 @@
 .section-title {
   margin-top: 24px;
 }
+div a mat-icon {
+  font-size: 15px;
+  vertical-align: bottom;
```
rrchai, line 9: middle maybe better?

rrchai:
either of version 3 looks good to me!

vpchung:
> `text-underline-offset`

CSS has come a long way since my web dev class... 😆

vpchung:
### Final design

Screenshot 2023-09-19 at 2 47 11 PM

tschaffter (approved):
Looks great, thanks!

Issue #1975, opened 2023-08-16
[Story] Minimize the time it takes to initialize the dev container during CI


tschaffter:
## Background

The CI workflow has recently been updated to use the Dev Container CLI to run tasks in the dev container. This ensures that the environment used by the CI workflow is the same as the development environment used by developers. This container also provides all the tools needed by the CI workflow and eliminates the need to maintain different versions of the tools in the CI workflow.

One drawback of the current implementation is that the dev container is not caching dependencies for Python, Java, Node.js, etc. This means that dependencies needs to be downloaded again from remote servers each time the CI workflow runs.

Initializing the dev container in the CI workflow is composed of two steps: 1) start the dev container and 2) run the command `workspace-install`. These two takes can take together up to 6 minutes.

## Goal

The goal of this ticket is to explore means to minimize the initialization of the dev container. The tasks are:

- [x] #1979
- [x] #1982 
- [x] #1983 
- [x] #1988 
- [x] #1852

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

tschaffter:
## Time it takes to start the dev container

This workflow step takes slightly less than 2 minutes based on the runtime of past runs.

The Dev Container CLI performs the following operations:

1. Download the base image of the dev container
2. Build a new image: base image + install features + VS Code extensions + configuration defined in `devcontainer.json`

Option to reduce the start up time:

- Cache Docker images so that the base image is cached
- Remove dev container features and VS Code extension that are not used by the CI workflow
- Go back to running the base dev container without the Docker CLI
  - One issue that prevented this is that the Docker CLI was not installed in the dev container

tschaffter:
## About stopping the dev container

This step takes 4 seconds, which is not much. It's also probably safer to shutdown the container properly.

tschaffter:
## GH Runner hardware resources

- CPU cores: 2
- RAM: 7 GB
- Storage:
    ```console
    Run df -h
    Filesystem      Size  Used Avail Use% Mounted on
    /dev/root        84G   62G   22G  75% /
    tmpfs           3.4G  172K  3.4G   1% /dev/shm
    tmpfs           1.4G  1.1M  1.4G   1% /run
    tmpfs           5.0M     0  5.0M   0% /run/lock
    /dev/sda15      105M  6.1M   99M   6% /boot/efi
    /dev/sdb1        14G  4.1G  9.0G  31% /mnt
[...]

PR #1982, opened 2023-08-17
feat: Cache R packages in CI workflow

tschaffter:
Contributes to #1975

## Description

Enable the caching of R packages when running the CI workflow. This PR makes the CI workflow runs 3 minutes faster as R packages are cached and no longer need to be downloaded for each run of the workflow.

## Changelog

- Cache R packages
- Remove `--mount type="bind,source=$HOME/.docker,target=/home/vscode/.docker"` from the workflow when pushing to `main`. It was not set for the workflow triggered by push to a PR so it should not be needed.

## Notes

- This PR is similar to #1979 but the tricky part (permissions) here is that the renv cache is outside of the repo folder cloned by the CI workflow (`~/.cache/R/renv/cache`).

### Without caching Node.js packages (current)

image

### When caching Node.js packages (this PR)

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,5 +265,5 @@ jobs:
       #         --projects=openchallenges-app,openchallenges-challenge-service,openchallenges-organization-service,openchallenges-image-service,schematic-api \
       #         --parallel=1"
 
-      - name: Stop the dev container
-        run: docker rm -f sage_devcontainer
+      # - name: Stop the dev container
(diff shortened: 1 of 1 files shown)


tschaffter:
## Permission issue

As expected, the container can't currently access the folder.

```console
Internal Error: EACCES: permission denied, mkdir '/home/vscode/.cache/node'
Error: EACCES: permission denied, mkdir '/home/vscode/.cache/node'
Error: Process completed with exit code 1.
```

PR #1977, opened 2023-08-16
feat: Change endpoint names to use nodes instead of components

andrewelamb:
Depends on https://github.com/Sage-Bionetworks/sage-monorepo/pull/1918
fixes [fds-908](https://sagebionetworks.jira.com/browse/FDS-908)

PR #1985, opened 2023-08-18
feat: add /assetTypes/{assetType}/manifests/{manifestId}/json endpoint

andrewelamb:
Depends on https://github.com/Sage-Bionetworks/sage-monorepo/pull/1977
fixes [fds-825](https://sagebionetworks.jira.com/browse/FDS-825)


apps/schematic/api/schematic_api/controllers/schema_controller_impl.py
```diff
@@ -6,39 +6,86 @@
...
+from schematic_api.models.validation_rules_page import ValidationRulesPage
+from schematic_api.models.validation_rule import ValidationRule
 from schematic_api.models.nodes_page import NodesPage
 from schematic_api.models.node import Node
 from schematic_api.controllers.utils import handle_exceptions
 
 
-def get_component_attributes(
-    component_label: str,
+def get_node_label_from_schematic(
```
mialy-defelice, line 16: is this in conflict with 1977? Where its to get the property label?

apps/schematic/api/schematic_api/controllers/storage_controller.py
```diff
@@ -13,6 +13,23 @@
 from schematic_api.controllers import storage_controller_impl
 
 
+def get_manifest_json(asset_type, manifest_id):  # noqa: E501
```
mialy-defelice, line 16: sorry I don't understand how everything in openapi works. is this a whole endpoint to return the manifest as a JSON? Or just the result of selection in the endpoint that could also return the manifest in other outputs?

apps/schematic/api/schematic_api/controllers/storage_controller_impl.py
```diff
@@ -51,6 +54,54 @@ def get_asset_storage_class(asset_type: str) -> Callable:
     return asset_type_object
 
 
+def get_manifest_from_schematic(asset_type: str, manifest_id: str) -> pd.DataFrame:
+    """Downloads a manifest in and returns the path
```
mialy-defelice, line 58: it looks like its returning the manifest not the path.

apps/schematic/api/schematic_api/openapi/openapi.yaml
```diff
@@ -232,18 +232,169 @@ paths:
       tags:
       - Storage
       x-openapi-router-controller: schematic_api.controllers.storage_controller
-  /components/{componentLabel}/attributes:
+  /assetTypes/{assetType}/manifests/{manifestId}/json:
```
mialy-defelice, line 235: Okay I think this answers a previous question, why do we need a whole endpoint for this?

mialy-defelice (changes requested):

Issue #1651, opened 2023-06-15
[Bug] Explore the result of the default org search is not deterministic


tschaffter:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What projects are you seeing the problem on?

OpenChallenges

### Current behavior

Reloading the org search page with default filter values multiple times shows different orgs.

## Take 1

<img width="1436" alt="Screenshot 2023-06-15 at 9 53 52 AM" src="https://github.com/Sage-Bionetworks/sage-monorepo/assets/3056480/4be81778-e3fd-4677-973c-b7188826d6ce">

## Take 2

<img width="1439" alt="Screenshot 2023-06-15 at 9 54 10 AM" src="https://github.com/Sage-Bionetworks/sage-monorepo/assets/3056480/3e556651-8450-41a3-b029-ccff667ae3e3">

### Expected behavior

The output of the searches should be the same if the data have not changed (they are not).

### Anything else?

The default sorter is "Relevance" that lets Hibernate Search identifies the order of the results based on "scores" it generates.

Tasks:

- Check if the scores generated are deterministic
- If the scores are deterministic, check why the results are not ordered by their score

### Commit ID

33d2826

### Are you developing inside the dev container?

- [X] I am using the dev container

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

tschaffter:
The order of the results is not fully random and usually alternate between a few results "configurations".

I can also see some randomness occurring if different results have the same score and the results of the sorting is the result of rounding floats.

PR #1987, opened 2023-08-18
feat(ui): Adjust long texts to better fit on challenge card

### LINKED ISSUE #1948: [Feature] Revisit challenge card designs for those with longer titles

OpenChallenges

### Description

Challenge cards with long titles (span over 3+ lines) doesn't look too great, e.g.

Screenshot 2023-08-08 at 10 36 16 PM

We could either:
* increase the card height
* overwrite the line height set for the card-title
* set a character limit for the card title
* hide the platform information if the title is too long

Of these, I like option 2 the best, followed by 1 or 3.  

### Anything else?

_No response_

### Code of Conduct

rrchai:
- closes #1948 

## Changelog
- Truncate following challenge's properties that could have long string into two lines with an ellipsis using css approach (`webkit-box`)
  - `name`, `headline`, `incentives`
- Adjust card's width and height
- Adding extra spaces between widgets on the card's footer
- Tested on Chrome, Firefox, Edge ✅ 

>**Note**
>According to the current schemas, the `name` and `headline` of challenge should not have more than 60 and 80 chars respectively . They should be enforced when seeding data via SQL to match the maximum chars defined in schemas. The truncation in this pr is still useful to prevent from this situation and retain the styling of the challenge card.
 
### Before
Screen Shot 2023-08-21 at 7 03 47 PM

### After
Screen Shot 2023-08-21 at 7 02 12 PM


tschaffter:
I kind of remember a discussion with Verena about 60 char being too short for titles. Even if we end up accepting longer titles, displaying max 60 chars with an ellipse on the card should be a future-proof solution.

tschaffter:
This is kind of a weird case. Maybe we should avoid breaking words with the ellipse?

image

rrchai:
## Updated
If the last word exceeds the maximum limit, remove the entire word as well as special characters.

Screen Shot 2023-08-18 at 6 36 31 PM

tschaffter:
Converting this PR to draft the time to implement the changes discussed this morning.

rrchai:
@tschaffter it seems data response by sorting relevance is returned in the same order now 😮

tschaffter:
@rrchai Are you referring to #1651?

rrchai:
>@rrchai Are you referring to https://github.com/Sage-Bionetworks/sage-monorepo/issues/1651?

Yeah, I tried to refresh multiple times, I got the same order of challenges, at least the first couple lines.

libs/openchallenges/styles/src/lib/_general.scss
```diff
@@ -244,9 +244,16 @@ table {
 }
 
 // CARDS
+@mixin line-camp($n) {
```
tschaffter, line 247: Replace `n` by a more descriptive variable.

tschaffter (approved):
This solution works great! Here are a couple of suggestions:

- rename the variable `n` with a more descriptive name
- update the title and description of this PR since its scope goes beyond the challenge card

PR #2030, opened 2023-08-28
feat: add /assetTypes/{assetType}/manifests/{manifestId}/json endpoint

andrewelamb:
fixes [fds-825](https://sagebionetworks.jira.com/browse/FDS-825)


andrewelamb:
@GiaJordan This PR is ready for review

andrewelamb:
@mialy-defelice In regards to your comments from [here.](https://github.com/Sage-Bionetworks/sage-monorepo/pull/1985)

That one had too many merge conflicts to handle so I started a new PR here

"is this a whole endpoint to return the manifest as a JSON?" That is correct. My understanding is when the schema of the output of the endpoint is different (csv vs json) they should be different endpoints. 

"it looks like its returning the manifest not the path."
You are correct, this has been fixed.

mialy-defelice:
> "is this a whole endpoint to return the manifest as a JSON?" That is correct. My understanding is when the schema of the output of the endpoint is different (csv vs json) they should be different endpoints.

Thank you for clarifying!

GiaJordan (approved):

PR #2019, opened 2023-08-26
feat: Add endpoint to the OC challenge service to get challenges per year data

tschaffter:
- Closes #2018 
- Depends on #2015

## Notes

- The plot data are hardcoded in the challenge service. A future PR will update the challenge service to generate these data dynamically, likely using the Hibernate library that relies on Elasticsearch to perform efficient searches - unless JPA provides a more developer-friendly solution with similar performance.

## TODO

> **Warning**
> The description of the API endpoint should be updated to reflect that we are now reporting the total number of challenges tracked and not the number of challenges that started a given year. Done!

## References

- https://www.baeldung.com/jpa-hibernate-projections#jpaaggregate (JPA)
- https://www.baeldung.com/jpa-hibernate-projections#hibernateaggregate (Hibernate)

## Preview

The request shown below works when sent 1) directly to the challenge service and 2) via the API gateway.

```
GET {{basePath}}/challengeAnalytics/challengesPerYear

HTTP/1.1 200
{
  "years": [
    "1969",
    "2011",
    "2012",
    "2013",
    "2014",
    "2015",
    "2016",
    "2017",
    "2018",
    "2019",
    "2020",
    "2021",
    "2022",
    "2023"
  ],
  "challengeCounts": [
    11,
    12,
    18,
    22,
    28,
    35,
    44,
    69,
    80,
    99,
    117,
    159,
    181,
    199
  ]
}
```

PR #2048, opened 2023-08-29
feat(backend): Get access to OC challenges-per-year plot data

tschaffter:


tschaffter:
Already part of #2019

PR #2086, opened 2023-09-05
fix: `workspace-docker-stop` now checks if there are containers to stop 

### LINKED ISSUE #2085: [Task] Prevent `workspace-docker-stop` from showing help menu if no containers are running

Sage Monorepo

### Description

When no containers are running:

```console
% docker stop $(docker ps -aq)
"docker stop" requires at least 1 argument.
See 'docker stop --help'.

Usage:  docker stop [OPTIONS] CONTAINER [CONTAINER...]

Stop one or more running containers
```

I tried the command below but it also show the message in the dev container but not on Mac OS.

```console
$ docker ps -aq | xargs docker stop | xargs docker rm
```

### Anything else?

_No response_

### Code of Conduct

tschaffter:
Closes #2085

### DIFF
--- a/dev-env.sh
+++ b/dev-env.sh
@@ -122,7 +122,10 @@ function workspace-welcome {
 }
 
 function workspace-docker-stop {
-  docker stop $(docker ps -q)
+  num_containers_running=$(docker ps -q | wc -l)
+  if [[ $num_containers_running -gt 0 ]]; then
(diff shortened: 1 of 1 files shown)


mdsage1 (approved):
I like the removal of the  <"docker stop" requires at least 1 argument.>  feedback as it has made me think I've done something wrong in the past.

PR #2113, opened 2023-09-11
feat: Add /assetTypes/{assetType}/assetViews/{assetViewId}/datasets/{datasetId}/manifestJson endpoint

andrewelamb:
fixes [fds-900](https://sagebionetworks.jira.com/browse/FDS-900)

### DIFF
--- a/apps/schematic/api/schematic_api/controllers/storage_controller_impl.py
+++ b/apps/schematic/api/schematic_api/controllers/storage_controller_impl.py
@@ -1,10 +1,10 @@
 """Implementation of all endpoints"""
 import os
-from typing import Optional, Union, Callable
+from typing import Optional, Union, Callable, Any
 
 from flask import request  # type: ignore
 import pandas as pd
-from schematic.store.synapse import SynapseStorage, ManifestDownload  # type: ignore
+from schematic.store.synapse import SynapseStorage, ManifestDownload, load_df  # type: ignore
 from schematic import CONFIG  # type: ignore
 
 from schematic_api.models.basic_error import BasicError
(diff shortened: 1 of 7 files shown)


apps/schematic/api/schematic_api/controllers/storage_controller_impl.py
```diff
@@ -159,6 +159,56 @@ def get_dataset_files(
...
+    """
+    access_token = get_access_token()
+    asset_type_object = get_asset_storage_class(asset_type)
+    store = asset_type_object(access_token=access_token)
+    manifest_data = store.getDatasetManifest(
+        datasetId=dataset_id, downloadFile=True, newManifestName="manifest.csv"
+    )
+
+    manifest_local_file_path = manifest_data["path"]
+    manifest = pd.read_csv(manifest_local_file_path)
```
GiaJordan, line 183: Instead of using `pd.read_csv` we might want to use the schematic function for reading in CSVs to maintain a unified approach (`schematic.utils.df_utils.load_df`)

GiaJordan (changes requested):

apps/schematic/api/schematic_api/controllers/storage_controller_impl.py
andrewelamb: Should this opther endpoint do the same?

GiaJordan: @andrewelamb yes, sorry I missed that then

andrewelamb:
No problem! I'll change them both here.

GiaJordan:
> No problem! I'll change them both here.

Cool, thank you! I'm not sure what the requirements are for these endpoints, but you might not necessarily need all the processing that happens in the function.

andrewelamb:
@GiaJordan Could you re-review?

GiaJordan (approved):

Issue #1966, opened 2023-08-15
[DX] Use Ng Required inputs (Ng v16)


tschaffter:
### What projects is this feature for?

OpenChallenges

### Description

In v16 now you can mark an input as required:

```typescript
@Component(...)
export class App {
  @Input({ required: true }) title: string = '';
}
```

Source: https://blog.angular.io/angular-v16-is-here-4d7a28ec680d

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

tschaffter:
Here is the error that we get when required inputs are not provided to a component:

```console
Error: libs/openchallenges/team/src/lib/team.component.html:80:1 - error NG8008: Required inputs 'privacyPolicyUrl', 'termsOfUseUrl' from component FooterComponent must be specified.
```

PR #2115, opened 2023-09-12
fix: Specify the missing inputs of the footer

### LINKED ISSUE #2027: [Bug] The OC web app config is not properly applied in production

OpenChallenges

### Current behavior

As an example, the link of the Privacy Policy is not properly set. This bug seems to have been introduced after updating the Angular app to use exclusively standalone components. See #1992.

image

### Expected behavior

_No response_

### Anything else?

_No response_

### Commit ID

46d81e7

### Code of Conduct

tschaffter:
- Closes #2027
- Contributes to #1966

## Changelog

- Pass all required input to the footer instances
- Start using `@Input({ required: true })`

## Notes

The footer is defined as a UI component in the project `openchallenges-ui`. Following a best practice from Nx and to make the components reusable, the components defined in this project should not rely on app-specific services. This is the reason why the value of the footer inputs are no longer defined in the footer component itself. Rather, the input values are specified where the footer instances are created.

Unfortunately, the current design is that each page content has its own instance of the footer instead of defining a single instance - e.g. in the `app.component.ts`. I believe that the reason is to provide pages with more flexibility, but as far as I remember, we are not using this flexibility. Hence, the current design leads to code duplication and maintenance. Note that code duplication is one of the metrics measured by SonarCloud.

Unless we really needs the flexibility mentioned above, I would be okay with creating a single footer instance in the `app.component.ts` to reduce duplicated code (in a future PR).

### DIFF
--- a/libs/openchallenges/about/src/lib/about.component.html
+++ b/libs/openchallenges/about/src/lib/about.component.html
@@ -125,4 +125,6 @@
 <openchallenges-footer
   [appVersion]="appVersion"
   [dataUpdatedOn]="dataUpdatedOn"
+  [privacyPolicyUrl]="privacyPolicyUrl"
+  [termsOfUseUrl]="termsOfUseUrl"
 />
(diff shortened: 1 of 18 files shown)


rrchai (approved):
LGTM

rrchai:
> Unless we really needs the flexibility mentioned above, I would be okay with creating a single footer instance in the app.component.ts to reduce duplicated code (in a future PR).

I agree to use a single footer across page in the future~

PR #2126, opened 2023-09-14
feat: add /schematicVersion endpoint

andrewelamb:
fixes [fds-1025](https://sagebionetworks.jira.com/browse/FDS-1025)

### DIFF
--- a/apps/schematic/api/schematic_api/controllers/versions_controller_impl.py
+++ b/apps/schematic/api/schematic_api/controllers/versions_controller_impl.py
@@ -1,0 +1,15 @@
+"Version endpoint controllers"
+
+import importlib.metadata
+
+from schematic_api.controllers.utils import handle_exceptions
+
+
+@handle_exceptions
+def get_schematic_version() -> str:
+    """Gets the current schematic version
+
+    Returns:
+        str: The current schematic version
+    """
+    return importlib.metadata.version("schematicpy")
--- a/apps/schematic/api/.openapi-generator-ignore
+++ b/apps/schematic/api/.openapi-generator-ignore
@@ -34,4 +34,5 @@ setup.py
 test-requirements.txt
 schematic_api/test/test_storage_controller.py
 schematic_api/test/test_schema_controller.py
+schematic_api/test/test_versions_controller.py
 tox.ini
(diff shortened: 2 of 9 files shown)


apps/schematic/api/schematic_api/controllers/versions_controller_impl.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@handle_exceptions
+def get_schematic_version() -> str:
+    """Gets the current schematic version
+
+    Returns:
+        str: The current schematic version
+    """
+    return importlib.metadata.version("schematicpy")
```
GiaJordan, line 15: Does this give an accurate result? I ask because checking the version of a locally installed schematic is kind of funky at the moment. I don't think the version spec in the `.toml` is consistently updated and schematic doesn't have functionality to check its own version. I recognize that this is not ideal; there is an [issue open to remedy](https://sagebionetworks.jira.com/browse/FDS-831). 

For the api purposes right now it's specified thorough an environment variable in the docker image. If using that doesn't work and it must be checked this way then @AmyHeiser we should prioritize the above linked issue.

GiaJordan (changes requested):

apps/schematic/api/schematic_api/controllers/versions_controller_impl.py
andrewelamb: Here's the [documentation](https://docs.python.org/3/library/importlib.metadata.html.)

It does, it returns `23.7.1` which is the version of schematic it's currently using

andrewelamb: @GiaJordan I'm not sure if this would work for getting the version of "itself" that seems complicated.

GiaJordan: Can you send the link again please? I'm getting a `404 Error`

GiaJordan: > It does, it returns `23.7.1` which is the version of schematic it's currently using

cool, thanks for confirming!

andrewelamb: Here's the link:

GiaJordan: Thanks, that's what I was looking at too. I think the difference is you've installed schematic with `pip` and I have the developers install.

GiaJordan (approved):

PR #2148, opened 2023-09-22
feat(ui): add platform logos to landing page

### LINKED ISSUE #2144: [Story] Introduce challenge platforms on the OC Home page

OpenChallenges

### Description

Add a section that introduces the user to the concept of "challenge platform". This could be achieved by showing the logo of all the platforms included in the OC DB.

### Anything else?

Suggestions:
- Adopt a dynamic solution so that the section auto-updates when we register new platforms.
- The square logo we use for platforms may not be recognizable by the user...

### Code of Conduct

vpchung:
Fixes #2144 

## Changelog 
* add "Platforms" section 
* add list of platform logos (Figma mockup [here](https://www.figma.com/file/kjlOBxTQpJY79hUju4LVUT/OC-Platforms?type=design&node-id=3-2&mode=design&t=qJ0MGxWQCXXRu2ID-0))

   > [!NOTE]
   > platforms are listed in alphabetical order, as to not show favoritism

* update number of platforms in the metrics section from 16 to 12; we will not be counting: 
   * "CASP" (@gaiaandreoletti and I concluded that this is more technically an organization rather than a platform)
   * "Other"
   * "OpenML" or "Papers With Code" (not challenge-hosting platforms)
* update wording so there's a better story (see Preview)

## Preview
Screenshot 2023-09-28 at 4 59 10 PM


tschaffter:
@vpchung Please create a new component instead of "hijacking" the home-organization component.

libs/openchallenges/assets/src/assets/images/platforms.svg
tschaffter, line 1: The image is almost 1M and is too large to be part of the app bundle and GitHub repo. The image should be placed in the S3 bucket.

tschaffter (changes requested):
- I would prefer to keep the "Discover Organizations" as it is now.
- The new Platform section should have a title that match the size of the title of "Discover Organizations" (no need for sub-titles)

vpchung:
> @tschaffter : I would prefer to keep the "Discover Organizations" as it is now.

## Option A
Section titles only

Screenshot 2023-09-28 at 8 40 26 PM

## Option B
Section title with organization short summary

Screenshot 2023-09-28 at 8 43 42 PM

## Option C
Section title with both short summaries

Screenshot 2023-09-28 at 8 47 11 PM

libs/openchallenges/assets/src/assets/images/platforms.svg
vpchung: Good idea.

I moved the svg to the bucket, but the image service doesn't work for me again.  So I'm not able to check if the change is correct 🙃

rrchai:
Option B looks good to me - some explanations for user can't go wrong in my opinion.

Maybe not showcasing the total number of platforms "12" here, instead using "various" or other alternatives, because it may not seem a big number for users who don't know challenge platforms.

rrchai (approved):

vpchung:
Thank you both for the feedback!  I took both of your suggestions and compromised with the following:

Screenshot 2023-09-29 at 11 07 07 AM

Merging now~

rrchai:
@vpchung 

The platform image looks like this on my end:

Screen Shot 2023-10-02 at 10 53 07 AM

vpchung:
@tschaffter whenever you have time, can you review the PR again?  I'm not able to merge until you approve.

tschaffter:
I can't see how the entire Platform section looks like right now (section title + logo). Here is my feedback based on what I remember seeing during our 1:1.

If I remember correctly, the vertical space A between the title and the first row of logo is smaller than the vertical space B between rows of logos. Similarly to previous feedback, the layout would be better structured IMO if the section title was one "element" and the logos were a second "element". To achieve this, the vertical space A should be larger than B.

tschaffter (changes requested):
I can't see how the entire Platform section looks like right now (section title + logo). Here is my feedback based on what I remember seeing during our 1:1.

If I remember correctly, the vertical space A between the title and the first row of logo is smaller than the vertical space B between rows of logos. Similarly to previous feedback, the layout would be better structured IMO if the section title was one "element" and the logos were a second "element". To achieve this, the vertical space A should be larger than B.

vpchung:
@tschaffter @rrchai I decreased the vertical padding between the frames to 5px, as well as added more padding between the image and text:

Screenshot 2023-10-02 at 10 47 17 AM

libs/openchallenges/home/src/lib/platforms/platforms.component.html
```diff
@@ -0,0 +1,12 @@
+<section id="platforms">
+  <div class="container text-center">
+    <h3 class="section-title">Platforms</h3>
+    <p>
+      The OC features challenges from various platforms used by the biomedical and research community
+      to evaluate and compare computational algorithms.
```
tschaffter, line 6:
```suggestion
      OC lists challenges from multiple platforms used by the biomedical and research community. These platforms enable participants to evaluate and compare the performance of their computational tools.
```

vpchung:
@tschaffter Thanks for the suggestion!  Here is the updated description:

Screenshot 2023-10-02 at 11 23 36 AM

vpchung:
Hmm... the text and image are not aligned in the center for some reason... investigating further.

vpchung:
Fixed:

Screenshot 2023-10-02 at 11 32 57 AM

libs/openchallenges/home/src/lib/platforms/platforms.component.html
```diff
@@ -2,8 +2,9 @@
   <div class="container text-center">
     <h3 class="section-title">Platforms</h3>
     <p>
-      The OC features challenges from various platforms used by the biomedical and research community
-      to evaluate and compare computational algorithms.
+      OC lists challenges from multiple platforms used by the biomedical and research community.
+      These platforms enable participants to evaluate and compare the performance of their
+      computational tools and algorithms.
```
tschaffter, line 7: Too many "and" in these two sentences?

vpchung: What about:

> OC provides a catalog of various platforms used by the biomedical and research community.  These platforms enable participants to benchmark their computational tools and algorithms.

tschaffter: I like this paragraph.

tschaffter (approved):

PR #2159, opened 2023-09-27
fix(ui): metric updates on homepage

### LINKED ISSUE #2145: [Story] Show OC DB metrics in the top part of the home page

OpenChallenges

### Description

This improvement was suggested by Paul Boutros as a replacement to the "generic" search image we currently have. An alternative would have been to customize the image with "zing" as tracked in #1836.

### Anything else?

_No response_

### Code of Conduct

vpchung:
Fixes #2145 

## Changelog 
* move metric numbers out of statistics section into landing section
* add new platform metrics

   > [!NOTE]
   > The number of platforms is currently hard-coded in

## Preview

There are two proposed designs:

**Option A**
Screenshot 2023-09-26 at 9 57 03 PM

**Option B**
Screenshot 2023-09-26 at 10 13 46 PM


rrchai:
I prefer option A where the sections have consistent width. I'm also sightly in favor in the original design of metrics that uses horizontal layout and it has less empty space around the page.
  
Screen Shot 2023-09-27 at 10 16 12 AM

https://github.com/Sage-Bionetworks/sage-monorepo/pull/2136#issuecomment-1731781362

tschaffter (commented):
I vote for Option A because it is more compact and probably easier to read for the user: their eye motion would go from top to down for each metric, instead of a "^" for Option B. The width of Option B goes beyond the width of other section of the home page. The fact that Option A is more compact also reduces the risk that the layout will be modified in response to a page resize or because of smaller displays.

> **Warning**
> Because of the introduction of server-side rendering (SSR) soon, the metrics should ideally not be visible in the viewport of the home page when the page is loaded. This is because the server cannot render the components that use the counter animation library. We can go around this relatively easily, which will probably be needed when we remove the dialog at the top of the page.

Illustration of predicted eye motion for Option B:

image

vpchung:
Thank you both for the feedback!  I have updated the layout:

Screenshot 2023-09-27 at 9 03 01 AM

rrchai (approved):

tschaffter (approved):
The only comment I had about the latest preview was to bring the three metrics closer so that they fit into the width of the page adopted by all the home page components. I think that this has already been addressed.

PR #2189, opened 2023-10-02
fix(openchallenges): customize scroll position restoration with different navigation types

### LINKED ISSUE #2154: [Bug] Reset scrolling position when navigating to another page (Take 2)

OpenChallenges

### Current behavior

See:

- #1972 
- #2153

### Expected behavior

_No response_

### Anything else?

_No response_

### Commit ID

_No response_

### Code of Conduct

rrchai:
- fixes #2154 

## Changelogs

- Add a service to enable custom scroll position restoration. The built-in `scrollPositionRestoration` seems not to support  the navigations when only parameters change. For example, in the `search` pages, the scroll position is always reset to "top" when filter values changes (if `scrollPositionRestoration: 'top'`. Therefore, adding a custom service could overwrite default scroll reset behavior and restore the previous position depending on different navigations:

    ```
    CustomScrollRestorationService: 
    A service for restoring scroll positions based on the method of navigation.
    It helps preserve the user's scroll position during various types of navigation actions.
    - Backward navigation: Scrolls to the previous position.
    - Forward navigation: Scrolls to the top of the page.
    - Anchor navigation: Scrolls to a specified anchor (no example currently).
    - Param navigation: Maintains the current scroll position when base url is the same, i.e. /challenge -> /challenge?status=active
    - Same url navigation (was not set before): Scrolls to the top of the page, i.e. /home -> /home
    ```

#### Backward navigation
backward

#### Forward navigation
forward

#### Same url navigation
sameurl

#### Param navigation
param


rrchai:
Some attempts using custom directive to prevent scroll position from being reset to 'top' is unsuccessful.

When navigate back from `challenge-profile` page to `challenge-search` page, the custom directive works to keep the previous scroll position. However, the scroll position will still be reset to 'top' when filters on the `challenge-search` page change. 

Looking for other options...

rrchai:
## Update

It looks like `scrollPositionRestoration: 'top'` will overwrite the manual scrolls at least in `ngOnit`. Not sure if it's because of race issues. I am going to test `window.scroll(0, 200)` in different lifecycle hooks with `scrollPositionRestoration: 'top'`.

### Results:
- `ngOnit` : :x:
-  `ngAfterContentInit`:
- `ngAfterViewInit`:

apps/openchallenges/app/src/app/app.config.ts
```diff
@@ -49,7 +50,11 @@ export const appConfig: ApplicationConfig = {
       routes,
       withEnabledBlockingInitialNavigation(),
       withInMemoryScrolling({
```
tschaffter: Remove if no longer needed

apps/openchallenges/app/src/app/custom-scroll-restoration.service.ts
```diff
@@ -0,0 +1,71 @@
...
+  private scrollY = 0;
+
+  constructor(
+    private router: Router,
+    private viewportScroller: ViewportScroller
+  ) {}
+
+  scrollPositionRestoration(): void {
+    // Capture the current scroll position
+    window.addEventListener('scroll', () => {
```
tschaffter, line 29: Is there a way to do this without using `window`? `window` is not available when rendering pages on the server (SSR). Hence, this code will fail is the method is called during server-side rendering. I will give it a try and report my feedback in this PR.

tschaffter (commented):
This looks like an elegant solution. I need to confirm that the use of `window` will not break SSR before approving this PR.

apps/openchallenges/app/src/app/custom-scroll-restoration.service.ts
rrchai: Not sure if it will help, but I have changed to get current scroll position directly via `viewportScroller`, instead of `window`.

rrchai:
I will explore the simpler solution @tschaffter suggested and convert this pr to draft.

rrchai:
Closing this - the new approach to address the issue is tracked in #2273

PR #2217, opened 2023-10-07
fix(openchallenges): use default text for empty start/end dates

### LINKED ISSUE #2130: [Bug] 'Not available' not shown for empty date fields

OpenChallenges

### Current behavior

Screenshot 2023-09-14 at 1 50 22 PM

### Expected behavior

`Not available` should be displayed for all empty profile fields, including start/end dates.

### Anything else?

_No response_

### Commit ID

_No response_

### Code of Conduct

vpchung:
Fixes #2130 

## Changelog
* update logic for `use_default()`
   * I am using `str == null` (instead of `str === null || str === undefined`) since this expression will return false for undefined values as well
* apply `text-gray` class for Start Date and End Date if date not available

## Preview
Screenshot 2023-10-06 at 5 23 44 PM

Just in case, I also checked the organization profile to ensure the latest logic change doesn't affect existing default texts:
Screenshot 2023-10-06 at 5 23 32 PM


libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.ts
```diff
@@ -19,8 +19,8 @@ export class ChallengeOverviewComponent {
   organizationCards: OrganizationCard[] = MOCK_ORGANIZATION_CARDS;
   // mockTopics = ['breast', 'cancer'];
 
-  use_default(str: string) {
-    return str === '' ? 'Not available' : str;
+  use_default(str: string | null | undefined) {
```
tschaffter, line 22: - Use camel case for function name in TypeScript (`useDefault` instead of `use_default`)
- Use a more descriptive name

libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.ts
```diff
@@ -19,8 +19,8 @@ export class ChallengeOverviewComponent {
...
+    return str === '' || str == null ? 'Not available' : str;
```
tschaffter, line 23: Simplify with `return str || 'Not available'`?

tschaffter (changes requested):

libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.ts
tschaffter: Another option is to use the nullish coalescing operator (`return str ?? 'Not available'`) IF we consider that an empty string is a valid value. If an empty value is not valid, we should define in the data/API model that a string should have a min length > 0.

Here is a TS playground to learn more about the [nullish coalescing operator](https://www.typescriptlang.org/play?ssl=1&ssc=8&pln=1&pc=35#code/PTAEBUAsFNQOwK4BskEsDOlQGMD2BDJadbVOAc1FwAdoAnfAF1ztA1Hzg6UfridQA3WM1AAfMQCgQoAO6RU2LHWiMEdOOlCMYoOqnKRGAWnSoAJrGgAPaivRncXVADNtuoi5NnL0sO0QUKlYEOEsXMmhzADpJP1AASS48OEYGdEYAGnExUAR0YlAXQnQATxwYbABrdGyAW2hOMkpOUGg66kZS+Iz9CmD3WEQ6gCN6UAAGOVxkc1AxnCcfeiiikuhY+IBBUHJcXDmbfA6iIpZ3dhdGtRU2LUtCZrlUHVBqfDpGVEJ43BGAK2g2EYWnkiiwkHwwlA4XwyBBchgXHwoCq0HKGDgAHJGG98A5VmRNmReHRithYFtqNQAMJOCLkdQCJygADeklAoBkABFoMV4QAuUAAIgAFHBcPBjtAAJTCgDcbQ6XVAvSeCQAyqBBI9zBypQ0hWqKPK4pyeXy4TwhcYAIyKqbsHVoPWcl7tdBC4ZjOim-UW-nW7R0BDQfX4YFCaBCkb7IicU0AXziLlCkZZCGo5iY0Cp1FFKQZQoACh8voQADx5ulwBlMr5OAB8MrZ-rAAHUXlhAkhjHhCMRSP0aPQmCx9YWDNF+A1QABeRa1qcz2AAflXIvFkpXctNnMn5Gi7rqWgXB6PvBPoHXoDte8XDOiEa+0LP9Knz6j143aVDfvNYA0uoKipKquBIAgDZwBO76Hiu87aKUtC4G457wXOGEisa5DCteD7LtKoBCmKEoGrKCowUuh7HqeiHIahsEXh686YcK3r0LhG7njRRG3valGPp+r50dAKH4YeQmwBhC7CrG4GNHAnHiU+kbQkKv7QP+XJgAAqmY-QSFQtAMMwrBgkoiyzLsX4jPgczZow+ACQRs5vlR06EYZJHbtKu7aTCuCFBKuKEEguCyGcrDCrhZDUJBznUZetHcUlOR8YqMjmIFWjBdwYURS45yOnAcWMAlKkvlJymSWlGkZWAKiheUtnmNkhCyPgpRaBpkjJvEACaMw4K0jVzHULCwPgsaQfAyBoJgiwDiQTxkIMoAAMzRAA7PMYWUNQuAZAKfjxEYjDUJ6ICWIIIx7eg0R1IodCHShjDRHgdTAF0tAkPonTAJwEpps0xjfYOf0mOtxhbcAQA).

libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.ts
```diff
@@ -19,8 +19,8 @@ export class ChallengeOverviewComponent {
   organizationCards: OrganizationCard[] = MOCK_ORGANIZATION_CARDS;
   // mockTopics = ['breast', 'cancer'];
 
-  use_default(str: string | null | undefined) {
-    return str === '' || str == null ? 'Not available' : str;
+  useNaIfEmpty(str: string | null | undefined) {
+    return str ?? 'Not available';
```
tschaffter, line 23: This instruction may return an empty string, which contradicts the name of the function.

tschaffter (changes requested):

libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.html
```diff
@@ -1,7 +1,7 @@
 <main *ngIf="challenge">
   <div id="bio">
     <h3 class="top-section-title">Overview</h3>
-    <p>{{ useNaIfEmpty(challenge.description) }}</p>
+    <p>{{ useNaIfFalsey(challenge.description) }}</p>
```
tschaffter, line 4: Don't US speakers always adopt the shortest words? ;-)

> A falsy (sometimes written falsey) value is a value that is considered false when encountered in a [Boolean](https://developer.mozilla.org/en-US/docs/Glossary/Boolean) context.

Source: https://developer.mozilla.org/en-US/docs/Glossary/Falsy

tschaffter (approved):

PR #2230, opened 2023-10-13
fix(openchallenges): move OC API docs URL to app config

### LINKED ISSUE #2219: [Bug] Add OC API docs URL to the app config file

OpenChallenges

### Current behavior

#2216 introduces an hard-coded value to `https://openchallenges.io`. This link is only valid for the production server and would not work for a local or staging deployment.

### Expected behavior

- If the app is deployed locally, the domain should be set to `http://localhost`.
- If the app is deployed on a staging server, the domain should be set to `https://staging.openchallenges.io` (or equivalent).

### Anything else?

_No response_

### Commit ID

_No response_

### Code of Conduct

vpchung:
Fixes #2219 

## Changelog
* Add new variable (`API_DOCS_URL`) to openchallenges-app env file
* Add new variable (`apiDocsUrl`) to openchallenges-app `config.json` and `config.template.json`
* Update HTML and TS files that display the footer so that the API docs URL used is dependent on where the app is deployed

## Preview
Local deployment:
Screenshot 2023-10-13 at 11 16 04 AM


tschaffter:
I changed the title of the ticket and PR to reflect that the scope is limited to externalizing the API docs URL and not the entire OC web app domain as initially described in the ticket.

tschaffter (approved):

PR #2250, opened 2023-10-19
feat(openchallenges): fetch total number of platforms from backend on Home page

### LINKED ISSUE #2241: Retrieve the number of platforms on Home page
The number of platforms displayed on the Home page's landscape section is currently based on a static value. With the ChallengePlatformService available on the backend, we should now update this to dynamically retrieve the value from the backend.

rrchai:
- closes #2241 

## Changelog

- use `challengePlatformService` to get the total number of platforms in database

## Preview

Screen Shot 2023-10-19 at 3 47 54 PM

### DIFF
--- a/libs/openchallenges/home/src/lib/challenge-search/challenge-search.component.ts
+++ b/libs/openchallenges/home/src/lib/challenge-search/challenge-search.component.ts
@@ -1,5 +1,6 @@
 import { Component, Inject, OnInit, PLATFORM_ID } from '@angular/core';
 import {
+  ChallengePlatformService,
   ChallengeService,
   Image,
   ImageHeight,
@@ -36,6 +37,7 @@ export class ChallengeSearchComponent implements OnInit {
   searchTerms!: string | undefined;
   challengeCount$: Observable<number> | undefined;
   orgCount$: Observable<number> | undefined;
+  platformCount$: Observable<number> | undefined;
   reanimateOnClick = false;
   challengeImg$: Observable<Image> | undefined;
   orgImg$: Observable<Image> | undefined;
@@ -46,6 +48,7 @@ export class ChallengeSearchComponent implements OnInit {
     private router: Router,
     private imageService: ImageService,
     private challengeService: ChallengeService,
+    private challengePlatformService: ChallengePlatformService,
     private organizationService: OrganizationService,
     @Inject(PLATFORM_ID) private platformId: string
   ) {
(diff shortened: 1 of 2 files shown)


libs/openchallenges/home/src/lib/challenge-search/challenge-search.component.ts
```diff
@@ -81,6 +84,12 @@ export class ChallengeSearchComponent implements OnInit {
         pageSize: 1,
       })
       .pipe(map((page) => page.totalElements));
+    this.platformCount$ = this.challengePlatformService
+      .listChallengePlatforms({
+        pageNumber: 1,
+        pageSize: 1,
```
tschaffter, line 90: I just checked and we do require a page to have at least 1 item. Maybe we could set the min to 0 for the sake of making this query lighter (future work).

tschaffter (approved):

libs/openchallenges/home/src/lib/challenge-search/challenge-search.component.ts
```diff
@@ -81,6 +84,12 @@ export class ChallengeSearchComponent implements OnInit {
         pageSize: 1,
       })
       .pipe(map((page) => page.totalElements));
+    this.platformCount$ = this.challengePlatformService
+      .listChallengePlatforms({
```
vpchung, line 88: Is there a way we can filter out platforms with a challenge count of 0?  We currently don't list CASP, OpenML, or. PapersWithCode on the [homepage](https://www.figma.com/file/kjlOBxTQpJY79hUju4LVUT/OC-Platforms?type=design&node-id=0-1&mode=design&t=OnoN0k5KS2RqICXu-0).

vpchung (changes requested):

libs/openchallenges/home/src/lib/challenge-search/challenge-search.component.ts
vpchung: OR - if this makes more sense - I can add a flag to the platforms page on whether we display the platform on the app or not.  This would be similar to what we did for organizations.

vpchung: @tschaffter @rrchai @gaiaandreoletti

tschaffter: I'm in favor of the second option, i.e. export to CSV only the platforms that should be exported. We could add `visibility` property in the future so that all the data are stored in the DB but that would require work to be planned to update the backend.

vpchung (approved):

libs/openchallenges/home/src/lib/challenge-search/challenge-search.component.ts
vpchung: > I'm in favor of the second option,

Okay, same.  This change will be done with the db update tomorrow.

PR #2252, opened 2023-10-19
feat(schematic): move visualization operations endpoints

andrewelamb:
This PR moves the following endpoints (previously tagged "Visualization Operations") into this repo:

- /schemaAttributes (/visualize/attributes)
- /components/{componentLabel} (/visualize/component)
- /tangledTreeLayers (/visualize/tangled_tree/layers)
- /tangledTreeText (/visualize/tangled_tree/text)

Fixes [fds-1050](https://sagebionetworks.jira.com/browse/FDS-1050)
Fixes [fds-1051](https://sagebionetworks.jira.com/browse/FDS-1050)
Fixes [fds-1052](https://sagebionetworks.jira.com/browse/FDS-1050)
Fixes [fds-1053](https://sagebionetworks.jira.com/browse/FDS-1050)


linglp (approved):
@andrewelamb I think the PR looks good and I pre-approve for now. I also tried using the endpoint, and it seems like that it could return expected result. 

But I saw that it seems like some query parameters are called "ComponentLabel" while some are called "Component"... is there any reasons why we distinguish these two? The description of "component" is: "A component in a schema, either the dsplay label or schema label" (note: there's a typo here that should be corrected in openapi.yaml), and the description of "componentLabel" is: "The label of a component in a schema". But I am still confused why we need to separate these two.

apps/schematic/api/schematic_api/openapi/openapi.yaml
```diff
@@ -1791,6 +2025,10 @@ components:
       type: object
       x-java-class-annotations:
       - '@lombok.Builder'
+    ComponentLabel:
```
linglp, line 2028: For `component/{componentLabel}` endpoint, when I clicked on executing the endpoint, I noticed that it returned: 
```
{
  "detail": "'MolecularEntity' is not a valid component in the schema.",
  "status": 500,
  "title": "Internal error"
}
```
What is the old endpoint called for this one?

andrewelamb:
> @andrewelamb I think the PR looks good and I pre-approve for now. I also tried using the endpoint, and it seems like that it could return expected result.
> 
> But I saw that it seems like some query parameters are called "ComponentLabel" while some are called "Component"... is there any reasons why we distinguish these two? The description of "component" is: "A component in a schema, either the dsplay label or schema label" (note: there's a typo here that should be corrected in openapi.yaml), and the description of "componentLabel" is: "The label of a component in a schema". But I am still confused why we need to separate these two.

I tried to differentiate between when a display name or label name is needed. Apparently it won't matter after the schema refactor as all the methods will allow the user to supply either the label or display name, so it will all be cleaned up then anyway.

apps/schematic/api/schematic_api/openapi/openapi.yaml
andrewelamb: This was previously `GET /visualize/component`

PR #2253, opened 2023-10-19
fix: Prevent invalid challenge incentive names from being added to OC DB

### LINKED ISSUE #2243: [Bug] The OC challenge service returns 500 error if the optional `direction` property is not specified

OpenChallenges

### Current behavior

This query works:

```console
http://localhost:8085/v1/challenges?direction=asc
```

The query fails if the property `direction` is not specified:

image

### Expected behavior

The query `http://localhost:8085/v1/challenges` should return a page of challenges.

UPDATE: The goal of this ticket is to prevent future invalid incentive values to be added to the DB.

### Anything else?

_No response_

### Commit ID

b98f284beafeba27548fa417e612c5d8397c0216

### Code of Conduct

tschaffter:
Closes #2243

### DIFF
--- a/apps/openchallenges/challenge-service/src/main/java/org/sagebionetworks/openchallenges/challenge/service/model/entity/ChallengeIncentiveEntity.java
+++ b/apps/openchallenges/challenge-service/src/main/java/org/sagebionetworks/openchallenges/challenge/service/model/entity/ChallengeIncentiveEntity.java
@@ -10,7 +10,6 @@ import javax.persistence.Id;
 import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
 import javax.persistence.Table;
-import javax.validation.constraints.Pattern;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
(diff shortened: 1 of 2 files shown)


tschaffter:
## Trying adding a constraint to the SQL table

```sql
CREATE TABLE `challenge_incentive`
(
    `id`                    int NOT NULL AUTO_INCREMENT,
    `name`                  ENUM('monetary', 'publication', 'speaking_engagement', 'other'),
    `challenge_id`          bigint(20) NOT NULL,
    `created_at`            DATETIME DEFAULT CURRENT_TIMESTAMP,
    PRIMARY KEY (`id`),
    FOREIGN KEY (`challenge_id`) REFERENCES challenge(`id`),
    CONSTRAINT unique_item UNIQUE (`name`, `challenge_id`),
    CONSTRAINT check_name CHECK (`name` in ('monetary', 'publication', 'speaking_engagement', 'other'))
);
```

But I can't get the service to complain when data are loaded from CSV files. Could it be that constraints are bypassed when using this import mechanism?

Also, shouldn't the `ENUM('monetary', 'publication', 'speaking_engagement', 'other')` be enough to invalidate value that do not match (e.g. `others`)?

cc: @vpchung

tschaffter:
I'll close this PR and the linked issue as it has been technically fixed with a data update. We are working in parallel on increasing the quality of the data to prevent future occurrences of this issue.

vpchung:
From what I can find online, the general consensus is for data validation to be done before LOAD LOCAL INFILE.  This will mean the Python script will need to be updated to perform data validation.

PR #2264, opened 2023-10-22
feat(openchallenges): add new `operation` property to Challenge schema

vpchung:
## Changelog

* Add `operation_id` to Challenge data model
* Regenerate API spec + Java files
* Add `operation_id` to `challenges` table in DB
* Include operation in CSV dump file for challenges
* Update db-update script to include `operation_id` into the CSV dump
* Add **EDAM Operation Class** to challenge profile, but I think there is a bug ([see this comment](https://github.com/Sage-Bionetworks/sage-monorepo/pull/2264#issuecomment-1953610523))

## Preview
**`challenges` table**
Screenshot 2024-02-16 at 4 49 18 PM

**API docs**
Screenshot 2024-02-16 at 4 49 40 PM

Screenshot 2024-02-16 at 4 49 50 PM

**Challenge profile**
Screenshot 2024-02-19 at 11 04 34 PM


apps/openchallenges/challenge-service/src/main/resources/db/migration/V1.0.0__create_tables.sql
```diff
@@ -39,6 +39,7 @@ CREATE TABLE `challenge`
     `difficulty`            ENUM('good_for_beginners', 'intermediate', 'advanced'),
     `platform_id`           int,
     `doi`                   varchar(80),
+    `operation`             varchar(80),
```
tschaffter, line 42: Do we need 80 chars are can we use less?

libs/openchallenges/api-description/src/components/schemas/ChallengeOperation.yaml
```diff
@@ -0,0 +1,5 @@
+description: The EDAM operation class of the challenge.
+type: string
+maxLength: 80
```
tschaffter, line 3: Same comment as above

libs/openchallenges/api-description/src/components/schemas/ChallengeOperation.yaml
tschaffter, line 1: Add `nullable: true`?

tschaffter (commented):
This is related to the other PR, but 0 should not be a valid operation (ID).

vpchung:
## 2024 update:

* `operation` is now `operation_id`

Screenshot 2024-02-16 at 4 49 18 PM

Screenshot 2024-02-16 at 4 49 40 PM

Screenshot 2024-02-16 at 4 49 50 PM

I have started adding the operation class to the challenge profile, but I think some updates to the REST API is needed before I can use `challenge.operationId`.

gaiaandreoletti (approved):

tschaffter (commented):
The first comment in the thread should be updated as it includes deprecated information.

Have we discussed how the operation ID is going to be used? I can then implement the necessary changes to the challenge service.

vpchung:
### Remaining TODO (maybe save for another PR?)

After updating the Angular CLI client, I was able to use `challenge.operation_id` on the frontend; "EDAM Operation Class" is now part of a challenge profile:

Screenshot 2024-02-19 at 11 04 34 PM

However, there should be 4 profiles that have a defined operation class, yet they are not shown on the profile:

Screenshot 2024-02-19 at 11 07 56 PM

The "Patient Mortality Prediction" challenge (top screenshot) is one of them.

tschaffter:
Refactored in #2534. Thanks @vpchung for this new feature!

PR #2265, opened 2023-10-22
feat(openchallenges): add EDAM terms SQL table and update `Challenge.inputDataTypes`

vpchung:
## Changelog
* add 2 new tables to challenge-service db
    * `edam_ontology_term` : a table consisting of EDAM ontology terms currently used to annotate challenges in the OC; includes parent classes
    * `challenge_input_data_annotation` : a table mapping challenges to their EDAM annotation(s)
* update API spec
* update Angular API client
* update Java files
* automated script that dumps the CSV files has been updated to include the EDAM terms and annotations into the dump

## References

- https://github.com/edamontology/edamontology
- https://bioportal.bioontology.org/ontologies/EDAM (used as reference for the EDAM schema)
 
## Preview
**API docs**
Screenshot 2024-02-19 at 10 51 16 PM

**`edam_ontology_term` table**
Screenshot 2024-02-19 at 8 19 01 PM

**`challenge_input_data_annotation` table**
Screenshot 2023-10-22 at 1 31 42 AM


vpchung:
## 2024 update:

* automated script that dumps the CSV files has been updated to include the EDAM terms and annotations into the dump

apps/openchallenges/challenge-service/src/main/resources/db/challenge_data_edam.csv
```diff
@@ -0,0 +1,304 @@
+"id","challenge_id","edam_id"
+"1","1","topic_3120"
```
tschaffter, line 2: This table should only reference foreign keys, which is not the case of `edam_id`. A related question is: should we make `edam_id` the primary key in the table `edam_terms.csv`?

tschaffter:
@vpchung Please update the information included in the first comment of this PR.

apps/openchallenges/challenge-service/src/main/resources/db/challenge_data_edam.csv
vpchung: Yes, using `edam_id` as PK makes sense to me.

vpchung: DB-related question: now that `edam_id` is the PK, should I go ahead and remove `id` from the table?

tschaffter: Yes

apps/openchallenges/challenge-service/src/main/resources/db/migration/V1.0.0__create_tables.sql
```diff
@@ -12,15 +12,16 @@ CREATE TABLE `challenge_platform`
 );
 
 
--- challenge_input_data_type definition
-CREATE TABLE `challenge_input_data_type`
+-- challenge_data_edam definition
+
+CREATE TABLE `edam_ontology_term`
```
tschaffter, line 17: Shall we rename this table `edam_term` since you renamed the file `edam_terms.csv`? Actually, do we maintain a convention between the name of the CSV files and their corresponding table names?

libs/openchallenges/api-description/src/components/schemas/SimpleChallengeInputDataType.yaml
tschaffter, line 1: Does the change to this file break the existing stack? Does the web app rely on this object?

The name and description of the object don't seem related. Shall the schema be renamed as EdamAnnotation?

Another question is whether we want to replace the existing data input object by an edam annotation object, or do we augment the existing data input data type object with an EDAM annotation?

apps/openchallenges/challenge-service/src/main/resources/db/migration/V1.0.0__create_tables.sql
vpchung: > do we maintain a convention between the name of the CSV files and their corresponding table names?

No, we do not.  I am not picky though - we can update the table name if you like. My reasoning was that `edam_ontology_term` is more descriptive than `edam_term`, which I know you prefer.

tschaffter: I will leave the CSV filenames up to you.

libs/openchallenges/api-description/src/components/schemas/SimpleChallengeInputDataType.yaml
vpchung: > Does the change to this file break the existing stack? Does the web app rely on this object?

I tested the app, and no breakage was encountered.  But it was a superficial test, so perhaps I didn't hit all of the edge cases.

> Shall the schema be renamed as EdamAnnotation?

This makes sense to me.

> Another question is whether we want to replace the existing data input object by an edam annotation object, or do we augment the existing data input data type object with an EDAM annotation?

I think this depends.  Do we see the OC using other ontologies in the future?  If so, I'd be inclined to keep it as `SimpleChallengeInputData` (or `ChallengeInputData`)

tschaffter:
Replaced by #2557

PR #2268, opened 2023-10-24
fix(openchallenges): expore `Location.replaceState()` to fix scrolling issue

### LINKED ISSUE #2154: [Bug] Reset scrolling position when navigating to another page (Take 2)

tschaffter:
Closes #2154

## Description

Explore an alternative solution to #2189 that may be simpler.

> Angular's Location service should be used when interacting with the browser's URL and not for routing. Thats why we want to use [Location](https://angular.io/api/common/Location#replaceState) service.

[Source](https://stackoverflow.com/a/68724008)

## References

- https://angular.io/api/common/Location#replaceState

### DIFF
--- a/apps/openchallenges/app/src/app/app.config.ts
+++ b/apps/openchallenges/app/src/app/app.config.ts
@@ -49,7 +49,7 @@ export const appConfig: ApplicationConfig = {
       routes,
       withEnabledBlockingInitialNavigation(),
       withInMemoryScrolling({
-        // scrollPositionRestoration: 'top',
+        scrollPositionRestoration: 'enabled',
       })
     ),
   ],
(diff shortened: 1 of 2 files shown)


tschaffter:
@rrchai I was wondering whether there was a simpler solution to the scrolling issue. I may have found one (see above). It seems that the existing code relies on on page navigation to trigger new searches. A more performant approach aligned with this PR would be to trigger queries when the object `this.query` changes.

rrchai:
@tschaffter  Thanks for the initial exploration of `replaceState()`. The changes based on your codes will be tracked in #2273 .

tschaffter:
See comment from Rong above. Closing this exploratory PR.

PR #2272, opened 2023-10-24
fix(sage-monorepo): build and publish the images of all the affected projects in CI workflow

### LINKED ISSUE #2271: [Feature] Build all the affected Docker images

Sage Monorepo

### Description

We had to limit the number of projects that could build and publish their image to GHCR because of limited storage space on the GH-hosted runner. We are now using a larger runner with enough space to build the images.

Tasks:

- Build and publish all the affected images (`main` and PR)
- Remove the Nx tasks that remove the image so that we benefit from local caching
  - This will use more storage space but we should be fine in the CI environment

### Anything else?

_No response_

### Code of Conduct

tschaffter:
Closes #2271

## Description

The CI workflow now builds - and publishes when pushing to `main` - all the Docker images of the affected projects. This set of images had to be reduced to a small, static number because of limited storage space available on the default GH-hosted runner. We recently upgraded Sage Monorepo to use a larger runner, which allows us to build the images of all the affected projects.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,9 +23,6 @@ jobs:
     # env:
     #   NX_BRANCH: main
     steps:
-      - name: Show storage space usage
-        run: df -h
-
       - uses: actions/checkout@v3
         name: Checkout [${{ github.ref_name }}]
         with:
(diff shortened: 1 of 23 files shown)


tschaffter:
@andrewelamb @sagely1 Could you please approve this PR as owner of the Schematic and Agora product?

sagely1 (approved):

PR #2273, opened 2023-10-25
fix(openchallenges): improve query parameter management using location service

### LINKED ISSUE #2154: [Bug] Reset scrolling position when navigating to another page (Take 2)

rrchai:
- fixes #2154 

## Changelogs

- Replace `router.navigate` with `location.replaceState()` to update url params without initiating navigation or scroll restoration
- Trigger API call by updating query once params are updated
- Refactor search pages' `*_component.ts` files
  - challenge-search: 608 -> 397 lines
  - org-search: 332 -> 316 lines (could reduce more by moving codes of generating orgCard to a global service, but save for a future pr)


libs/openchallenges/challenge-search/src/lib/challenge-search-filters.ts
```diff
@@ -1,88 +1,163 @@
...
+  },
+];
+
+export const challengeOrganizationsFilter: Filter[] = [];
+
+export const challengeOrganizersFilter: Filter[] = [];
+
+export const challengeSortFilter: Filter[] = [
+  {
+    value: 'relevance',
```
tschaffter, line 152: Would that work if we replace the string value by its enum value? This approach is more robust.

```suggestion
    value: ChallengeSort.Relevance,
```

tschaffter (commented):
The code refactoring looks great!

I'm having a CORS issue on my end that prevent the challenges from loading in the search page. I will check if this issue is also in `main`.

tschaffter:
@rrchai Filtering orgs by "Contribution Role" does not filter the results.

tschaffter (changes requested):
See comment above.

tschaffter (approved):

PR #2277, opened 2023-10-26
feat(openchallenges): add temporal information to challenge card

### LINKED ISSUE #2240: [Task] Add temporal information to the challenge card ⭐
Update the challenge card, so that a user can more easily infer what the start/end dates are at a glance. For example:

Image

See #2185 for more details.

vpchung:
Fixes #2240 

## Changelog
* add logic for calculating time difference in units of years, months, weeks, days, and hours
* display time until start date OR time until end date (depending on status), using the largest non-zero time unit
* due to limited space in the footer: re-arrange the card layout, switching placements of status and platform
* update "Incentives" icon to something more obvious (trophy)
* add fixed heights for status and headline divs; that way, cards stay uniformed in height no matter the contents

## Preview

Footer icons using the same color:

Screenshot 2023-10-26 at 12 31 22 AM

OR one footer icon uses the blue color from our logo and one icon uses the other:

Screenshot 2023-10-26 at 4 49 56 PM

Alternatively, we can forgo displaying the platform altogether (something Addison mentioned as potentially not helpful information):

Screenshot 2023-10-26 at 12 44 12 AM

### DIFF
--- a/libs/openchallenges/ui/src/lib/challenge-card/challenge-card.component.ts
+++ b/libs/openchallenges/ui/src/lib/challenge-card/challenge-card.component.ts
@@ -27,6 +27,7 @@ export class ChallengeCardComponent implements OnInit {
   desc!: string;
   incentives!: string;
   statusClass!: string;
+  time_info!: string | number;
   // difficulty!: string | undefined;
 
   constructor(
@@ -63,6 +64,45 @@ export class ChallengeCardComponent implements OnInit {
         : this.imageService.getImage({
             objectKey: 'banner-default.svg',
           });
+      if (this.challenge.endDate && this.status === 'completed') {
+        this.time_info = `Ended ${this.calcTimeDiff(
+          this.challenge.endDate
+        )} ago`;
+      } else if (this.challenge.endDate && this.status === 'active') {
+        this.time_info = `Ends in ${this.calcTimeDiff(this.challenge.endDate)}`;
+      } else if (this.challenge.startDate && this.status === 'upcoming') {
+        this.time_info = `Starts in ${this.calcTimeDiff(
+          this.challenge.startDate
+        )}`;
+      }
     }
+  }
+
+  calcTimeDiff(date: string | null | undefined) {
+    if (!date) {
+      return '';
+    }
+    const refDate: any = new Date(date + ' 00:00:00');
+    const now: any = new Date();
+    const diffMs = Math.abs(refDate - now);
+
+    // Calculate the time difference in years, months, weeks, days, and hours.
+    const timeDiff = {
+      year: Math.floor(diffMs / 31_556_952_000),
+      month: Math.floor(diffMs / 2_629_746_000),
+      week: Math.floor(diffMs / 604_800_000),
+      day: Math.floor(diffMs / 86_400_000),
+      hour: Math.floor(diffMs / 3_600_000),
+    };
+
+    // Find the largest unit of time and return in human-readable format.
+    let timeDiffString = '';
+    for (const [unit, value] of Object.entries(timeDiff)) {
+      if (value > 0) {
+        timeDiffString = `${value} ${unit}` + (value > 1 ? 's' : '');
+        break;
+      }
+    }
+    return timeDiffString;
   }
 }
--- a/libs/openchallenges/ui/src/lib/challenge-card/challenge-card.component.ts
+++ b/libs/openchallenges/ui/src/lib/challenge-card/challenge-card.component.ts
@@ -78,8 +78,9 @@ export class ChallengeCardComponent implements OnInit {
     }
   }
 
-  calcTimeDiff(date: string | null | undefined) {
-    if (!date) {
+  calcTimeDiff(date: string) {
+    const pattern = /\d{4}-\d{2}-\d{2}/;
+    if (!pattern.test(date)) {
       return '';
     }
     const refDate: any = new Date(date + ' 00:00:00');
--- a/libs/openchallenges/ui/src/lib/challenge-card/challenge-card.component.ts
+++ b/libs/openchallenges/ui/src/lib/challenge-card/challenge-card.component.ts
@@ -4,8 +4,6 @@ import { MatIconModule } from '@angular/material/icon';
 import { RouterModule } from '@angular/router';
 import {
   Challenge,
-  ChallengePlatformService,
-  SimpleChallengePlatform,
   Image,
   ImageService,
 } from '@sagebionetworks/openchallenges/api-client-angular';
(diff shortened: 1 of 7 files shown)


vpchung:
Latest iteration:

Screenshot 2023-10-27 at 2 07 44 PM

tschaffter:
@vpchung "Completed" could be reminiscent of errors to users. What about using the default (black) color for completed challenges?

tschaffter:
Another suggestion would be to hide time information for challenges that have no longer been "recently" completed (3 mo ago?).

vpchung:
### Black color for `completed` status; show all timeSince info
Screenshot 2023-10-27 at 2 42 25 PM

### Black color for `completed` status; hide timeSince info older than 3 months
Screenshot 2023-10-27 at 2 50 37 PM

### Colorized `completed` status; 
Screenshot 2023-10-27 at 2 50 58 PM
hide timeSince info older than 3 months

vpchung:
I'm not a big fan of the default color for the `Completed` status, but I can also see your point about red = error.

tschaffter:
@vpchung Why are you not a fan of the default color?

vpchung:
For me, it lacks consistency - either all statuses should be a different color from the default text color, or they should all be the same.

tschaffter:
@vpchung Thanks for the insight. I understand this viewpoint. I gave some thought about this and I think that the new status style capture more the attention that previously. The two reasons are that this information is now at the center of the card and that it's more colorful compared to the solution with the colored dot. With the proposed design, I think that I'm looking first at the status, then at the title. I think that this may disrupt the flow of reading that should be from top to bottom - or at least the title should be read first? Maybe my eyes are attracted to the status first because it's the topic of this PR...

Could we ask Adam to review the current proposal for the card?

@vpchung What do you think about hiding the platform information from the card? Did you mention on Friday that Addison suggested this too? The two reasons that prompt me to suggest this change are: 1) we know that the challenge ecosystem and information are fragmented by benchmarking communities and platforms. Removing this information from the card could be a way for us to "mend" this fragmentation to some extend if that makes sense. 2) enable the user to read more challenge cards by removing some information from the card.

tschaffter:
I like the "Slightly bigger bullet size" design.

vpchung:
PR is now ready for (re-)review.

libs/openchallenges/ui/src/lib/challenge-card/challenge-card.component.ts
```diff
@@ -63,6 +57,48 @@ export class ChallengeCardComponent implements OnInit {
...
+        this.time_info = `Ends in ${this.calcTimeDiff(this.challenge.endDate)}`;
+      } else if (this.challenge.startDate && this.status === 'upcoming') {
+        this.time_info = `Starts in ${this.calcTimeDiff(
+          this.challenge.startDate
+        )}`;
+      }
     }
   }
+
+  calcTimeDiff(date: string) {
```
tschaffter, line 75: Declare the return type.

```suggestion
  calcTimeDiff(date: string): string {
```

libs/openchallenges/ui/src/lib/challenge-card/challenge-card.component.ts
```diff
@@ -63,6 +57,48 @@ export class ChallengeCardComponent implements OnInit {
...
+    const pattern = /\d{4}-\d{2}-\d{2}/;
+    if (!pattern.test(date)) {
+      return '';
+    }
```
tschaffter, line 79: This function `calcTimeDiff` should validate the input data as you do but validation error should be handled instead of silently returning an empty string.

One approach is to `throw Error(`The input date does not match the schema ${pattern}: ${date}`)`. Then try/catch the function where it is called. The catch block should at least log the error with `console.error()` (note to myself: we should start using a logger within the app).

In the catch block you could then set `timeSince` to `undefined`, which is preferable to an empty string, since an empty since is probably not an expected value for `timeSince`.

Note that a function that may throw an error also has the return type `never`. The return type of the function would then be `string | never`.

Another approach is to change the function `calcTimeDiff` to require a `Date` object as input. The validation would happen where the Date object is constructed. The validation code in `calcTimeDiff` could be removed and the return type would simply be `string`.

The advantage of the second solution is that the function could be reused directly with "native" `Date` objects without the need to do `Date` => `string` => `Date`.

tschaffter (changes requested):

libs/openchallenges/ui/src/lib/challenge-card/challenge-card.component.ts
vpchung: To be addressed in a different PR.

libs/openchallenges/ui/src/lib/challenge-card/challenge-card.component.ts
vpchung: To be addressed in a different PR.

vpchung:
Thank you, @rrchai , for pointing out the inconsistency of the status colors during the standup.  I have addressed the issue with the following design options:

**Complete as red**
Profile | Card
-- | -- 
Screenshot 2023-11-01 at 12 22 11 PM | Screenshot 2023-11-01 at 12 22 20 PM

**Complete as blue**
Profile | Card
-- | -- 
Screenshot 2023-11-01 at 12 24 17 PM | Screenshot 2023-11-01 at 12 24 28 PM

Note that above, I changed the status color on the profile pages from black to red/blue.  The original design is black text, i.e.

Screenshot 2023-11-01 at 12 35 54 PM

Screenshot 2023-11-01 at 12 35 24 PM

libs/openchallenges/ui/src/lib/challenge-card/challenge-card.component.ts
tschaffter: @vpchung Please open a ticket with my comment as Description

vpchung:
Moving forward with blue for "Completed" status. As a safe measure, I also simulated the updated status colors with [Coblis](https://www.color-blindness.com/coblis-color-blindness-simulator/) to verify that this change is still accessible to users with color-blindness (they passed, according to the simulator).

tschaffter (approved):

rrchai:
LGTM with one small suggestion. Since the status has a circle in front of the text, align-center might not be perfectly center the status box with "ends/starts ..." box. Adding some margin-left on status box to move a bit toward left can make alignment better in my oppinion. 

### Before
Screen Shot 2023-11-01 at 6 49 59 PM

### After moving status box to left by 8px 
Screen Shot 2023-11-01 at 6 49 43 PM

rrchai (approved):

PR #2299, opened 2023-11-01
feat(openchallenges): add strict eslint rule on null checks 

### LINKED ISSUE #1544: [Story] Remove bang operator wherever possible

OpenChallenges

### As a user, I want

NA

### Description

The bang operator is useful when prototyping to tell the TS transpiler to not complain about a value potentially being null or undefined. That does not mean that the value will not take a null or undefined value.

We should review the occurrences of the bang operator and remove them whenever possible. Configuring eslint/prettier to throw a warning or an error when the operator is used is a good way to spot existing occurrences and prevent new ones from being created.

See [Typescript: Bang operator considered harmful](https://dev.to/tmaximini/typescript-bang-operator-considered-harmful-3hhi) 

### Acceptance criteria

_No response_

### Tasks

_No response_

### Anything else?

_No response_

rrchai:
- closes #1544

Currently, we can use "bang" operator (!) in TypeScript to bypass the null checks, but it has the potential risk, like actual null or undefined values at runtime leading to unforeseen errors.

## Changelog
- Instead of relying on the "bang" operator, use TypeScript's strict mode to deal with null and undefined values, so that the developers can be informed initially

## Preview

Now using "bang" operator (!) will not bypass the null checks, instead we can use (?) and (??) to handle nulls.
Screen Shot 2023-11-01 at 2 10 41 AM

➡️ see more details about [no-non-null-assertion](https://typescript-eslint.io/rules/no-non-null-assertion)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -29,6 +29,7 @@
         "dot-notation": "off",
         "no-useless-constructor": "off",
         "@typescript-eslint/no-explicit-any": "off",
+        "@typescript-eslint/no-non-null-assertion": "error",
         "prettier/prettier": [
           "error",
           {


rrchai:
> What happens to the bang operators we use in class member variables, e.g.:

It does not capture the variable declarations yet by using this rule alone I think. I'll explore if there are more strict rules on bang operators.

rrchai:
## Update
Not able to find rules to report errors on [definite-assignment-assertions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#definite-assignment-assertions) yet.

tschaffter:
Following up on using `!` when declaring variable members, consider the example below:

```typescript
private test!: string;
```

We are basically say "Hey TypeScript, this variable can not be `null` or `undefined`. The value is defined somewhere (other than inline or in the constructor) even if you don't see it.".

Instead, we could use:

```typescript
private test: string | undefined;
```

This makes it clear that the variable may be undefined and this will force us to address this when using the variable, instead of creating the possibility that we say "the variable can not be undefined" but it may actually be.

Let's move ahead and merge this PR and keep in mind the comments above when declaring new member variables.

rrchai:
I'll dedicate some time to explore the creation of a custom rule for this case. If this approach doesn't work, we can proceed with merging the current changes, which include the `no-non-null-assertion` rule alone.

tschaffter:
@rrchai I did a quick search and couldn't find an article that says that using the bang operator when defining member variables should be avoided. The fact that the eslint rules that you have enabled in this PR does not cover this case makes me think that it's OK to use it in this case, even though I think that we should prefer specifying the type `x | undefined` instead of telling TS "I'm an engineer, trust me this variable will never be undefined!".

I'll test this PR locally and then merge it.

tschaffter (approved):

PR #2306, opened 2023-11-01
feat(openchallenges): add Team button to OC navbar

### LINKED ISSUE #2303: [Task] Add Team button to OC navbar ⭐

_No response_

### As a user, I want

_No response_

### Description

- Add the Team button to the navbar
- Remove the Team link from the footer

### Acceptance criteria

_No response_

### Tasks

_No response_

### Anything else?

_No response_

tschaffter:
Closes #2303

## Changelog

- Add the button "Team" to the right panel of the navbar
- Replace `href` by `routerLink` in the navbar and footer for links that target app pages

> **Note**
> Navigating to a page of the app with `href` destroys and reloads the app. This can be observed by looking at the navbar that flickers when navigating with `href`. Instead, navigating to pages of the app should be done with `routerLink` (requires to import the `RouterModule`). In that case, the navbar will remain and only the components that are no longer used will be destroyed, which leads to faster navigation.

### Option 1

image

### Option 2

image

### DIFF
--- a/libs/openchallenges/ui/src/lib/navbar/navbar.component.html
+++ b/libs/openchallenges/ui/src/lib/navbar/navbar.component.html
@@ -14,6 +14,7 @@
 
   <div class="flex-spacer"></div>
 
+  <a mat-button class="navbar-item" routerLink="/team">Team</a>
   <openchallenges-discord-button />
 
   <!-- <a mat-button class="navbar-item" *ngIf="!isLoggedIn" routerLink="/login" aria-label="Log In">
--- a/libs/openchallenges/ui/src/lib/navbar/navbar.component.html
+++ b/libs/openchallenges/ui/src/lib/navbar/navbar.component.html
@@ -14,7 +14,7 @@
 
   <div class="flex-spacer"></div>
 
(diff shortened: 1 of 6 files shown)


libs/openchallenges/ui/src/lib/navbar/navbar.component.html
```diff
@@ -14,6 +14,7 @@
 
   <div class="flex-spacer"></div>
 
+  <a mat-button class="navbar-item" routerLink="/team">Team</a>
```
vpchung, line 17: Keep it consistent with the footer? "Meet Our Team"

tschaffter: I added a previous for "Meet Our Team". Do you prefer this option?

vpchung: Looks good! Thanks

vpchung (approved):
Between the two options, I like 2 for consistency.

LGTM

rrchai (approved):
LGTM

PR #2313, opened 2023-11-02
feat(schematic): change authentication so that only endpoints that need it have it

andrewelamb:
- Fixes [fds-1286](https://sagebionetworks.jira.com/browse/FDS-1286)
- Only endpoints that require synapse tokens have the synapse token auth requirement(storage, manifest submissions)
- Updates Schematic to 23.11.1
- Removed python client as a locked dependency (The version will be what schematic requires)
- Change jsonschema to `>=4.0.0 <4.10.0` see [here](https://sagebionetworks.jira.com/browse/FDS-1324). This avoids ~100 DeprecationWarnings during testing.
- Manifest validation endpoints now use synapse token if provided (needed for cross-manifest validation Schematic v23.11.1)
- Mypy was updated to 1.6.1, this change causes mypy to notice some issues with schematic and some workarounds had to be added
- Other packages updated
- Various typos fixed
- get_asset_storage_class(asset_type) function no longer used as it caused typing issues
- asset_type parameter not used until synapse storage refactor


linglp (approved):
@andrewelamb thanks. I think this PR looks good to me. I tried running all the endpoints that require authentication, and they all seemed to return desired output.

andrewelamb:
Thanks @linglp !

PR #2316, opened 2023-11-03
fix(openchallenges): refactor tab switching to preserve scroll position on profile pages

### LINKED ISSUE #2310: [Bug] Challenge & org pages jump to the top when selecting a section ⭐

OpenChallenges

### Current behavior

Challenge & org pages jump to the top when selecting a section

### Expected behavior

The scrolling position should not be reset when selecting a section.

### Anything else?

We should be able to fix this bug using `Location` as we do on the search pages.

### Commit ID

7ae8482caa921c479f38b5cc0f871baf5334d2e4

### Code of Conduct

rrchai:
- fixes #2310 

## Changelog

- Update tab switching on profile pages to avoid navigations, which will fix
  1. the scrolling position jumps back to top after switching tabs
  2. the challenge slug disappeared from the URL after switching tabs

  reset-scroll

- Ensure URL parameters sync with active tab 

  i.e. Typing `/org/dream?tab=members` will go to 'member' section

  org-member


libs/openchallenges/org-profile/src/lib/org-profile.component.ts
```diff
@@ -74,11 +71,9 @@ export class OrgProfileComponent implements OnInit {
   organization$!: Observable<Organization>;
   organizationAvatar$!: Observable<Avatar>;
   loggedIn = true;
-  // organizationAvatar!: Avatar;
   tabs = ORG_PROFILE_TABS;
-  tabKeys: string[] = Object.keys(this.tabs);
   activeTab!: Tab;
-  private subscriptions: Subscription[] = [];
+  private subscriptions = new Subscription();
```
tschaffter, line 76: Should be singular

```suggestion
  private subscriptions = new Subscription();
```

tschaffter (approved):
LGTM! Only one small comment

PR #2357, opened 2023-11-16
chore(openchallenges): update OC DB 2023-11-16

vpchung:
## Changelog

* Added:
   * 1 platform
   * 4 orgs
   * 6 challenges
* Updated:
   * headlines (going from 120 -> 80 max characters)
   * general updates, e.g. challenge counts
   * challenge counts for the plot

## Preview

**Updated plot:**
Screenshot 2023-11-16 at 12 38 36 PM

**New platform + their challenges:**
Screenshot 2023-11-16 at 10 43 28 AM

**Upcoming DREAM challenges:**
Screenshot 2023-11-16 at 10 42 57 AM

PR #2351, opened 2023-11-10
chore(openchallenges): 2023-11 DB update 

vpchung:
Fix #2266

## Changelog
* Added:
   * 175 Grand Challenge (GC) challenges
   * 53 organizations
* Updated:
   * some challenge headlines, e.g. remove challenge title from summary, fix casing, etc.
   * other general updates, e.g. challenge counts, submission types, etc.

> [!NOTE]
> I've only added the contributors of 21 challenges so far; I will continue to add these as the weeks go by.

## Preview
**Latest landscape numbers**
Screenshot 2023-11-10 at 1 51 51 PM

**All 181 public GC challenges are now in!**
Screenshot 2023-11-10 at 2 17 05 PM

PR #2358, opened 2023-11-16
chore(openchallenges): add all public Grand Challenges

tschaffter:
## Changelog

- [x] #2351 
- [x] #2357
- [x] #2360


tschaffter:
@vpchung I opened this PR so I can't technically review it. I did test the PR locally and it looks good to me. Couple of notes:

- Most GC challenges don't have start and/or end date as discussed on Slack
- The challenge logo comes from GC website. This comes with some limitations:
  - The image may much larger than needed and take time to load. Meanwhile, the logo will appear white with the alternative text (see screenshot)
  - The image may not be square. The OC style displays the image with an 1:1 aspect ratio, in which case the image would appear distorted.

image

vpchung:
Thanks, @tschaffter . I know this isn't what you intended to catch, but your screenshot for the CAMEO challenge (not GC as stated) was a bug in the curation.  This will be fixed with #2360 .

Re: the other image links, should we ignore the image links from GC until a solution for outside avatar links is handled?

tschaffter:
> Re: the other image links, should we ignore the image links from GC until a solution for outside avatar links is handled?

Let's keep the images to bring more colors to OC.

The "loading glitch" is a minor issue that we can fix in one of the next releases. Based on a random sampling of the GC challenges, most of the logos have a 1:1 aspect ratio and look fine on the challenge profile page.

vpchung:
> Let's keep the images to bring more colors to OC.

You know, I never actually checked whether the logos were displayed.  This is awesome!!

Screenshot 2023-11-16 at 3 11 22 PM

tschaffter:
@vpchung Checkout this one: http://localhost:8000/challenge/376

vpchung (approved):

PR #2359, opened 2023-11-16
fix(openchallenges): implement responsive width for `mullti-select` filter on search pages

### LINKED ISSUE #2349: [Bug] Set a max width to the challenge and org search menu in OC

OpenChallenges

### Current behavior

The width of the menu depends on the content of the filters. It's an issue when selecting DREAM:

image

image

### Expected behavior

_No response_

### Anything else?

_No response_

### Commit ID

a6a0bebc9dfba88c23146287bccbdcae38111cc4

### Code of Conduct

rrchai:
- fixes #2349 

## Changelog
- Add a responsive max-width for the multi-select dropdown filter to avoid unexpected expansion of the filter panel

## Preview

Laptop screen:
Screen Shot 2023-11-16 at 3 06 31 PM

Mobile screen:
Screen Shot 2023-11-16 at 2 55 04 PM

### DIFF
--- a/libs/openchallenges/challenge-search/src/lib/challenge-search.component.scss
+++ b/libs/openchallenges/challenge-search/src/lib/challenge-search.component.scss
@@ -45,9 +45,12 @@
   padding-top: 0;
   padding-bottom: 0;
 }
+::ng-deep .p-multiselect.p-component {
+  max-width: 200px;
+}
 
-@media screen and (min-width: constants.$md-breakpoint) {
+@media screen and (max-width: constants.$md-breakpoint) {
   ::ng-deep .p-multiselect.p-component {
-    max-width: 200px;
+    max-width: none;
   }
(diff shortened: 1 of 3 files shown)


libs/openchallenges/challenge-search/src/lib/challenge-search.component.scss
```diff
@@ -43,3 +45,12 @@
   padding-top: 0;
   padding-bottom: 0;
 }
+::ng-deep .p-multiselect.p-component {
```
vpchung, line 48: Something for us to keep in mind is that `ng-deep` will be deprecated one day.  But I think we can keep using it until W3C has an official replacement.

vpchung (approved):
⭐

tschaffter (approved):
Tested and approved!

PR #2368, opened 2023-11-17
feat(openchallenges): cap challenges to 2023 on timeline plot

### LINKED ISSUE #2365: [Feature] clarify future challenge counts on timeline plot

_OpenChallenges_

### Description

The plot currently extends to the years 2024 and 2025 (future years), which includes upcoming challenges beyond the current year. This seems to be causing confusion among users interpreting the chart.

The easiest fix is to cap to the current year.

### Anything else?

_No response_

### Code of Conduct

rrchai:
- closes #2365 

## Changelog

- change html to reflect caption
- update caption text by not specifying detailed number of missing challenges
- remove challengerPerYear data after current year

## Preview

Screen Shot 2023-11-17 at 6 23 40 PM

### DIFF
--- a/libs/openchallenges/home/src/lib/statistics-viewer/statistics-viewer.component.html
+++ b/libs/openchallenges/home/src/lib/statistics-viewer/statistics-viewer.component.html
@@ -3,7 +3,8 @@
   <div class="col container">
     <div echarts [options]="chartOptions"></div>
     <span class="caption">
-      *The plot excludes future challenges and/or challenges with no known start date.
+      *The plot does not include challenges that are planned beyond this year, or those with
+      unspecified start dates.
     </span>
   </div>
 </section>
--- a/libs/openchallenges/home/src/lib/statistics-viewer/statistics-viewer.component.html
+++ b/libs/openchallenges/home/src/lib/statistics-viewer/statistics-viewer.component.html
@@ -1,7 +1,7 @@
 <section id="statistics" class="mat-typography">
   <h3 class="section-title">Total Challenges Tracked</h3>
   <div class="col container">
-    <div echarts [options]="chartOptions"></div>
+    <div id="timeline-plot" echarts [options]="chartOptions"></div>
     <span class="caption">
       *The plot does not include challenges that are planned beyond this year, or those with
       unspecified start dates.
--- a/libs/openchallenges/home/src/lib/statistics-viewer/statistics-viewer.component.html
+++ b/libs/openchallenges/home/src/lib/statistics-viewer/statistics-viewer.component.html
@@ -2,5 +2,8 @@
   <h3 class="section-title">Total Challenges Tracked</h3>
   <div class="col container">
     <div echarts [options]="chartOptions"></div>
+    <span class="caption">
+      *The plot excludes future challenges and/or challenges with no known start date.
+    </span>
   </div>
 </section>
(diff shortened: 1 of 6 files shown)


tschaffter:
@rrchai Why not simply update the data on the backend?

rrchai:
### Option 1:

Simply cut off the challenges after current year, but due to missing upcoming challenge counts, the total number won't be matched with the number in database. 
 - we could expand the caption to say more future challenges not shown on the plot, but need to take into consideration that the overlaps between "unDatedChallengeCounts" and "futureChallengeCounts"

    Screen Shot 2023-11-17 at 3 59 34 PM

### Option 2:

Adjust the UI on future challenges

Screen Shot 2023-11-17 at 3 47 06 PM

vpchung:
Option 2 is beautiful!!!

vpchung:
> we could expand the caption to say more future challenges

To be honest, I still have an issue with the current caption we have.  It's not as concise as it should be, imo.  Could we maybe change it to:

"Plot does not include upcoming challenges and/or challenges with no known start date"

libs/openchallenges/home/src/lib/statistics-viewer/statistics-viewer.component.ts
```diff
@@ -31,75 +31,71 @@ export class StatisticsViewerComponent implements OnInit, OnDestroy {
...
+            },
+          ],
+          graphic: res.undatedChallengeCount
+            ? {
+                elements: [
+                  {
+                    type: 'text',
+                    style: {
+                      text:
+                        `*The OC database includes an additional ${res.undatedChallengeCount} challenges ` +
```
vpchung, line 86: Sorry,  I still have an issue with the current caption we have...  it's not as concise as it could be, imo.  Could we maybe change it to:

"Plot does not include upcoming challenges and/or challenges with no known start date."

vpchung (approved):
Option 2 looks really lovely and I'm a big fan, but I'm worried it may still be a bit confusing to users?

I also made a suggestion that could address your concerns about the caption.

libs/openchallenges/home/src/lib/statistics-viewer/statistics-viewer.component.html
```diff
@@ -2,5 +2,8 @@
   <h3 class="section-title">Total Challenges Tracked</h3>
   <div class="col container">
     <div echarts [options]="chartOptions"></div>
+    <span class="caption">
+      *The plot excludes future challenges and/or challenges with no known start date.
```
rrchai, line 6: any final suggestions on this sentence? @tschaffter @vpchung

rrchai: as @tschaffter suggests, the future/upcoming is not super accurate and might confuse with users, as there could be some upcoming challenges launching this year. 

How about: 
>*The plot does not include challenges that are planned beyond this year, or those with unspecified start dates.

vpchung: I like!  Thanks for updating the caption~

tschaffter: I like the latest caption! Nicely done!

tschaffter (approved):

rrchai:
```console
[ECharts] Can't get DOM width or height. Please check dom.clientWidth and dom.clientHeight. They should not be 0.For example, you may need to call this in the callback of window.onload.
```
I have accidentally encountered this error in browser, which caused the plot is not rendered. Applying a width of 100% resolves the problem, and this adjustment should not alter how the plot originally looks.

PR #2370, opened 2023-11-17
fix(openchallenges): reset pagination settings when filters change on search pages

### LINKED ISSUE #2364: [Bug] Triggering a new challenge or org search should reset the page number

OpenChallenges

### Current behavior

1. Go to the challenge search page
2. Go to the page 3 or more
3. Filter to show only the active challenge (currently 9 results)
4. The page shows "9 results" but no cards are displayed

### Expected behavior

The page number should reset to 1 when a new search is triggered (full text search, filters, sorters).

The same comment applies to the org page.

### Anything else?

Cc: @rrchai 

_No response_

### Commit ID

8442d2d6a505e2c6e3262d9665ae0be5516382df

### Code of Conduct

rrchai:
- fixes #2364 

## Changelog

- update logics to reset page to the 1st page whenever the filter changes (except pagination changes)
- avoid adding pageNumber and pageSize to the params in the url if they are default (`pageNumber=0`, `pageSize=24`)


tschaffter (changes requested):
The content of the page is now updated as expected. There is one remaining issue: the page number selected in the UI may may not be reset. E.g.

1. Go to the challenge search page
2. Go to page 4
3. Filter to keep only the active challenges

=> The content of the page shows the first page (correct) but the page selector still shows Page 4.

rrchai:
>the page number selected in the UI may may not be reset.
Good catch! Updating soon...

rrchai:
Updating the `[first]` attribute of the `<p-paginator>` doesn't seem to reflect correctly in the UI, which appears to be a bug. As a workaround, I'm manually switching the page to align with the page changes.

tschaffter (changes requested):
The latest update brings a new issue.

1. Go to the challenge search page
2. Click on page 2

=> Nothing happens

rrchai:
The codes have been updated. Here are expected behaviors for testings:

1. Page can be switched and expected challenges are displayed without problems
2. Go to any page number > 1 -> change any filters (including sorters, search bar) -> the page is reset to the first page
3. When page is reset to the first page, the param url should not contain `pageSize=0&pageSize=24`
     - note: when copy&paste a url containing `pageSize=0&pageSize=24`, it will not disappear, which is expected for now

libs/openchallenges/ui/src/lib/paginator/paginator.component.ts
```diff
@@ -9,22 +18,28 @@ import { PaginatorModule as PrimengPaginatorModule } from 'primeng/paginator';
...
+  // ngOnInit(): void {
+  //   this.itemIndex = this.pageNumber * this.pageSize;
+  // }
 
   onPageChange(event: any): void {
     this.pageChange.emit(event);
   }
+
+  // Reset paginator to first page
+  resetPage(): void {
```
tschaffter, line 42: Use a more descriptive name? "resetting the page" does not sound correct. Maybe `resetPageNumber` or `reset` as in reset the paginator.

libs/openchallenges/ui/src/lib/paginator/paginator.component.ts
```diff
@@ -9,22 +18,28 @@ import { PaginatorModule as PrimengPaginatorModule } from 'primeng/paginator';
   templateUrl: './paginator.component.html',
   styleUrls: ['./paginator.component.scss'],
 })
-export class PaginatorComponent implements OnInit {
+export class PaginatorComponent {
   @Input({ required: true }) pageNumber = 0; // index of the new page
   @Input({ required: false }) pageLinkSize = 5;
   @Input({ required: true }) pageSize = 0; // number of rows to display in new page
```
tschaffter, line 24: Is the comment correct? The page size is usually a number of items in a page, not a number of rows.

libs/openchallenges/ui/src/lib/paginator/paginator.component.ts
```diff
@@ -9,22 +18,28 @@ import { PaginatorModule as PrimengPaginatorModule } from 'primeng/paginator';
...
   @Input({ required: true }) totalRecords = 0;
```
tschaffter, line 25: That that correspond to the `totalElements` included with a page metadata returned by the backend? If yes, consider using the same name.

tschaffter (commented):
The behavior is as expected. Nice job! And we have shorter URLs now, which is definitively a plus.

libs/openchallenges/ui/src/lib/paginator/paginator.component.ts
rrchai: >Is the comment correct? The page size is usually a number of items in a page, not a number of rows.

In `primeng` terms, they use "rows" to represent "page". I will update the comment to match our OC language : )

tschaffter (approved):
Please merge once the comments are resolved.

PR #2393, opened 2023-12-06
fix(openchallenges): remove usage of "Other" for platform

### LINKED ISSUE #2383: [Task] Update CODA TB Challenge now that it's a continous benchmark

OpenChallenges

### As a user, I want

_No response_

### Description

Solly asked at the DREAM conference that we update the [CODA TB challenge](https://openchallenges.io/challenge/55/cough-diagnostic-algorithm-for-tuberculosis?tab=overview) now that the challenge is a continuous benchmark.

### Acceptance criteria

_No response_

### Tasks

_No response_

### Anything else?

_No response_

### LINKED ISSUE #2385: [Bug] Set the platform to `NULL` for challenge that don't use a registered platform

OpenChallenges

### Current behavior

The OC Data sheet registers a platform named "other". A first issue is that counting the platforms from the list of challenge platform returns currently 14 while we technically capture information only about 13 platforms. Other issues may arise as we plan to create a challenge platform search page.

Instead, we should no longer create a platform for "other" and set the property `challenge.platform` to null / undefined when no known challenge platform is used by the challenge. This approach implies changes at the following levels:

1. [x] OC Data sheet (@vpchung )
2. [x] OC data model (SQL) (@vpchung )
3. [x] OC API model (`openchallenges-api-description`) (@rrchai )
4. [ ] Challenge service (@tschaffter )
5. [ ] Web app (@rrchai )

The above tasks should be performed in this order, and this will be reflected on the person assigned to this ticket. Let's try to do everything in a single feature PR that Verena will create as the first contributor to keep thing simple.

Cc: @vpchung @rrchai 

### Expected behavior

_No response_

### Anything else?

_No response_

### Commit ID

_No response_

### Code of Conduct

vpchung:
Fixes #2385 

## Changelog
- remove "Other" from list of platforms
- replace `challenge.platform_id` for "Other" (14) with NULL (\N)

## Preview
**Landscape now shows 13 platforms**
Screenshot 2023-12-05 at 10 38 32 PM

**"Other" no longer shows in platform facet**
Screenshot 2023-12-05 at 10 38 48 PM

**Platform info shows "Not available" in profiles for challenges hosted on non-OC platforms**
Screenshot 2023-12-05 at 10 40 09 PM


tschaffter:
@vpchung What about representing missing values by an empty string in the OC Data sheet and settings the default value of the SQL property `challenge.platform` to `NULL`? If that works on the SQL side, this could help make the OC Data sheet less cluttered and reduce the size of the CSV files tracked on GitHub.

vpchung:
I don't think this will work, at least, not for the dump files to dump properly.  If some rows contain less columns than other rows, then that row's data will be offset.  But willing to try to see.

rrchai:
The updates on API descriptions regarding of removing "Other" option of platform are pushed in this pr.

Initially, I considered removing the `slug` of the platform and allowing the challenge service to generate the slug automatically in the backend. However, this might also require removing the 'slug' column from the data. 

To keep things simple, I've restricted the options to both the `slug` and `name` fields of the platform. @tschaffter  Please let me know if there are any issues.

libs/openchallenges/api-description/src/components/schemas/ChallengePlatformName.yaml
```diff
@@ -1,4 +1,17 @@
...
+  - Kaggle
+  - CodaLab
+  - CodaBench
+  - Eterna
+  - Nightingale OS
+  - EvalAI
+  - CACHE
+  - CAMEO
+  - DrivenData
+example: Synapse
```
tschaffter, line 17: We don't want an enum as platform data live in the DB, not in the API description.

rrchai:
> What is the problem with the platform slug? Isn't it enough to set the challenge.platform property to nullable in the API spec?

It is enough with nullable challenge.platform. I thought we would like to limit options using these known platforms as a validation. If not restricting platform options using `enum` in API description, I think we should have a validation in the backend/frontend (in addition to data) to exclude unknown platforms?

libs/openchallenges/api-description/src/components/schemas/ChallengePlatformName.yaml
rrchai: I have remove the enum values in descriptions

tschaffter:
Because the platform data live in the DB, we shouldn't do any validation in the spec, which would require us to update the API description each time we add/update platform data in the DB.

vpchung:
@tschaffter done ✅

tschaffter:
I updated the challenge service with the generator and weird things are happening like the challenge platform Java and TS object loosing some of its properties. Could this be a side effect of @rrchai using `AllOf` to make `challenge.platform` nullable? Currently investigating... 

image

tschaffter:
@vpchung I fixed the styling issue visible on your screenshot where "Not Available" is not greyed out for platform. Now that platform is nullable AND optional, the property in TypeScript is now:

```typescript
platform?: SimpleChallengePlatform | null;
```

The styling check should check for these two values:

```html
<td [ngClass]="{ 'text-grey': challenge.platform?.name === (undefined || null) }">
```

which I propose to simplify to the following:

```html
<td [ngClass]="{ 'text-grey': !challenge.platform }">
```

vpchung (commented):
Works on my end! Thanks, @rrchai @tschaffter

PR #2398, opened 2023-12-07
Fds 1284 testing

andrewelamb:
- fixes [fds-1288](https://sagebionetworks.jira.com/browse/FDS-1284)
- fixes [fds-1288](https://sagebionetworks.jira.com/browse/FDS-1288)

PR #2400, opened 2023-12-08
feat(schematic): integration testing

andrewelamb:
This is identical to [#2398](https://github.com/Sage-Bionetworks/sage-monorepo/pull/2398) plus some fixes to the testing workflow.

This PR chas a new wrokflow that is trggred when PRs are made from the new `Schematic-API-Staging` branch into `main`.

This PR:

- fixes [fds-1288](https://sagebionetworks.jira.com/browse/FDS-1284)
- fixes [fds-1288](https://sagebionetworks.jira.com/browse/FDS-1288)
- PRs are now done from forks to Schematic-API-Staging instead of main
- PRs to Schematic-API-Staging only trigger tests that don't require secrets
- PR's are then don form Schematic-API-Staging to Main
- A new workflow that triggers when PRs are done from Schematic-API-Staging to Main that runs all tests has been added
- All non synapse tests are not mocked at all
- A new set of tests test_synapse_endpoints.py has tests for all synapse endpoints where nothign is mocked
- handle_exceptions is now tested directly
- When the schema url isn't correct an InavlidSchemaURL exception is now raised
- handle exceptions returns a 404 status when InavlidSchemaURL is caught

### DIFF
--- a/.github/workflows/schematic-api-ci.yml
+++ b/.github/workflows/schematic-api-ci.yml
@@ -249,7 +249,7 @@ jobs:
           ]
           for secret in secrets:
               assert secret is not None
-              assert isinstance(sercret, str)
+              assert isinstance(secret, str)
               assert len(secret) > 0
           dict = {
             "synapse_token": "${{ secrets.SCHEMATIC_SYNAPSE_TOKEN }}",
--- a/.github/workflows/schematic-api-ci.yml
+++ b/.github/workflows/schematic-api-ci.yml
@@ -15,129 +15,12 @@ env:
   DOCKER_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
-  push:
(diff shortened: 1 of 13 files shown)


andrewelamb:
@tschaffter @GiaJordan Can I get your approval on this?

GiaJordan (approved):
approving per discussion on slack with @andrewelamb

tschaffter (changes requested):
- Can you please add a high-level description of what this PR is doing? When is this workflow being triggered?
- Why does the title of the PR includes "(#2398)"?
- The new GH workflow duplicates parts of the CI workflow common to all projects. This would lead to a waste of resources. Based on previous discussions, you may want this PR to focus on integration testing for Schematic. Also, now or in another PR, you could consider running it AFTER the common CI workflow successfully passes, so that e2e testing is not run if the linting check fails, for instance.

andrewelamb:
@tschaffter

tschaffter (approved):

PR #2403, opened 2023-12-13
feat(schematic): added nginx to schematic docker container

linglp:
## Context
To make sure that schematic APIs could be deployed to production, we have to add nginx. This is similar but different from what I did here in this PR: https://github.com/Sage-Bionetworks/schematic/pull/1202

## Changes
Some configuration files are the same as before. For example, `self-signed.conf`, `ssl-params.conf`, `uwsgi.ini`.. but there are also some changes: 
1. I remembered that I had to modify "entrypoint.sh" in the original image. But that wasn't a clean solution, and it is not a good practice to modify the original nginx.conf. To avoid that, I ended up adding `redirect.conf`
2. since the base image that we are using is still: `tiangolo/uwsgi-nginx-flask:python3.10`, the python version that we use in schematic API has to match the python version allowed by the base image. I ended up having to change python version from `3.10.12` to `3.10.13` in mono repo. If you do `nx prepare schematic-api` again, you should be able to get `3.10.13`
3. When updating the file permission, for simplicity, I just updated the file permission of `root` folder and `app` folder. But this is NOT recommended. Please see "future items" for more details. 

## To test this PR: 
1. `cd apps/schematic/api`
2. In `apps/schematic/api` folder, make sure that you have renamed `.env.example` to `.env` or you already have a `.env` file
3. run `nx prepare schematic-api` to get the correct version of Python
4. Run `poetry shell` to make sure that you are in an virtual environment. And to prepare for certificate and private key locally, run `prepare_key_certificate.py` This script will generate `private_localhost_certificate.crt` and `private_localhost.key` files and save the content of private key and certificate as environment variables in .env
5. Run the build command to build a schematic image: `docker build -t schematic-api .`
And run the schematic command by doing: `docker run -d --name mycontainer --env-file .env  -p 7443:7443 schematic-api`

OR
`nx build-image schematic-api`
when running `docker images`, you should be able to see something like: 
```
REPOSITORY                               TAG           IMAGE ID       CREATED          SIZE
ghcr.io/sage-bionetworks/schematic-api   local         3d4d4e9fd980   57 minutes ago   2.3GB
```
And then you could create a docker container by doing something like: `docker run -p 7443:7443 -t <docker image id>`

And if you are using VS code, you should be able to open the docker container like this: 
Screen Shot 2023-12-12 at 8 32 05 PM

When seeing the page below: 

Screen Shot 2024-02-14 at 4 08 45 PM

make sure you click on: Advanced -> "Proceed to localhost(unsafe)" 

## Test by `curl`
Using `curl` to test and ensure that everything is working as expected. The actual port number depends on your setting. Below are examples: 

1. `curl -I https://localhost:7443/api/v1/ui/`
I saw: 
```
curl: (60) SSL certificate problem: self signed certificate
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```
This makes sense because we are using a self-signed certificate

5. `curl -I -k https://localhost:7443/api/v1/ui/`
I saw: 
```
HTTP/2 200 
server: nginx/1.23.2
date: Thu, 11 Jan 2024 20:42:57 GMT
content-type: text/html; charset=utf-8
content-length: 1488
access-control-allow-origin: *
x-frame-options: DENY
x-content-type-options: nosniff
x-xss-protection: 1; mode=block
```
This is also expected because `-k` allow connections to SSL sites without certificates or to ignore certificate validation errors.

6. `curl -I -k http://localhost:7443/api/v1/ui/`: 
```
HTTP/1.1 302 Moved Temporarily
Server: nginx/1.23.2
Date: Thu, 11 Jan 2024 20:44:19 GMT
Content-Type: text/html
Content-Length: 145
Connection: close
Location: https://localhost:60110/api/v1/ui/
X-Frame-Options: DENY
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
```
This makes sense because we redirect http to https 

## Future items 
This Dockerfile is not a final product. There are some important issues that need to be address before actually using it for production deployment: 
1) .synapseCache is stored in `/root`
This is not a good practice. But currently schematic doesn't allow cache of .synapseCache folder to be configurable. This work is captured in ticket: FDS 1446
After we move `.synapseCache` out of root folder, we need to revisit this dockerfile and update the file permission according to the new location of cache. 

2) Run the dockerfile as a non-root user
I tried adding `USER www-data` in dockerfile but it didn't work because the base image: `tiangolo/uwsgi-nginx-flask:python3.10` that we are using is using `supervisord` to bind pots. And binding pots in docker requires super user permission. I can't have `www-data` to bind pots or work with `supervisord` for now. This will require additional tickets and attention if we want to meet the best practice. 

3) purging cache function needs to be moved out of schematic library
The purging cache function in schematic library currently calculate the cache size of .synapseCache and if it meets a certain criteria, the cache will get purge. We should move that logic to the mono repo (see issue FDS 1445)

4) Modify check_synapse_cache_size function to deal with empty .synapseCache folder
For the container built in schematic library, even when `.synapseCache` is empty, `du -sh .synapseCache` returns 4KB. But this is not the case when I tried running the same command in this docker container for some reasons. When .synapseCache folder is empty, `du -sh .synapseCache` returns 0 without a unit. We have to modify `check_synapse_cache_size` to deal with 0. (See FDS 1445) 

## Note: 
This should be merged after: #2417


apps/schematic/api/Dockerfile
```diff
@@ -1,41 +1,70 @@
...
+    ROOT=/ \
+    UWSGI_INI=/app/uwsgi.ini \ 
+    NGINX_WORKER_PROCESSES=1 \
+    VERSION=$TAG
 
-WORKDIR ${APP_DIR}
-COPY schematic_api schematic_api/
-COPY pyproject.toml poetry.lock uwsgi.ini ./
+# run open ssl and generate certificate
+RUN apt update && \
```
tschaffter, line 28: Clean the cache of the package manager to reduce the size of the image:

```
    && apt-get -y autoclean \
    && apt-get -y autoremove \
    && rm -rf /var/lib/apt/lists/*
```

apps/schematic/api/Dockerfile
```diff
@@ -1,41 +1,70 @@
-FROM python:3.10.12-slim-buster
+FROM tiangolo/uwsgi-nginx-flask:python3.10
 
-ENV APP_DIR=/opt/app
+# add version tag as a build argument
+ARG TAG
```
tschaffter, line 4: How do tag behave when no default value is specified?

apps/schematic/api/Dockerfile
```diff
@@ -1,41 +1,70 @@
...
+  && poetry install --with prod --no-root --no-interaction --no-ansi
 
-EXPOSE 7080
+# Update file permission 
+RUN mkdir /root/.synapseCache
+RUN mkdir /app/app/manifests
+# temporary here to ensure .synapseCache is not empty
+RUN echo "This is a test file." > /root/.synapseCache/test.txt
+# temporary here until we move .synapseCache to a different path
+RUN chmod -R 777 /root /app
```
tschaffter, line 65: Permission too broad? Also, be careful when changing the permission of the home folder of the root user.

apps/schematic/api/Dockerfile
```diff
@@ -1,41 +1,70 @@
...
+    NGINX_WORKER_PROCESSES=1 \
+    VERSION=$TAG
 
-WORKDIR ${APP_DIR}
-COPY schematic_api schematic_api/
-COPY pyproject.toml poetry.lock uwsgi.ini ./
+# run open ssl and generate certificate
+RUN apt update && \
+    apt-get install openssl && \
+    openssl req -x509 -nodes -days 365 \
```
tschaffter, line 30: - Is it OK to make the self-signed certificate publicly available (assuming a public image)?
- Is it OK for the certificate to change each time the image is rebuilt?

apps/schematic/api/docker-compose.yml
```diff
@@ -1,8 +1,11 @@
-version: "3.8"
+version: "3.10"
 
 services:
   schematic-api:
-    image: ghcr.io/sage-bionetworks/schematic-api:local
+    # image: ghcr.io/sage-bionetworks/schematic-api:local
+    build: 
+      dockerfile: Dockerfile
+    image: sagebionetworks/schematic-aws-api:latest
```
tschaffter, line 8: This breaks a few standards that the monorepo attempts to establish:
- images should ideally be pushed to GHCR
- the name of the image should ideally be the name of the project in the monorepo
- the project should define "tasks" in its project.json to build, scan and publish the image, hence the image build should not be defined in the docker-compose file.

apps/schematic/api/Dockerfile
```diff
@@ -1,41 +1,70 @@
...
-ENTRYPOINT ["/docker-entrypoint.sh"]
+WORKDIR ${APP_DIR}
 
-# Run server in development mode
-# CMD ["python", "-m", "openapi_server"]
-# Run server in production mode
-CMD ["uwsgi", "--ini", "uwsgi.ini", "--lazy", "--http", ":7080"]
\ No newline at end of file
+# Expose ports
+EXPOSE 443
```
tschaffter, line 70: - It is recommended to include a default command at the end of the Dockerfile
- It is not a good practice to "leave" the Dockerfile as root, unless you use the entrypoint strategy
  - The main process should not be run with root privileges

apps/schematic/api/docker-compose.yml
```diff
@@ -12,8 +15,9 @@ services:
     #   - ./docker-entrypoint-initdb.d:/docker-entrypoint-initdb.d
     networks:
       - schematic
-    ports:
-      - "${SERVER_PORT}:7080"
+    ports: 
+      - "${SERVER_PORT}:80"
+      - "443:443"
```
tschaffter, line 20: The monorepo aims to standardize how ports are defined In order to prevent ports from clashing.

- Schematic ports should be 7xxx
- The port exposed should be defined in `devcontainer.json`

tschaffter (changes requested):
This PR introduces changes that break best practices (security) and standards that the monorepo aims to establish. I suggest to schedule a call to clarify the content of this review.

apps/schematic/api/Dockerfile
tschaffter: Here's a breakdown of the considerations for deciding whether to add a self-signed certificate to a Docker image or mount it to the container:

Adding to the Image:

Pros:

    Self-contained: The certificate is embedded within the image, making it portable and usable across environments.
    Simplified deployments: No need to manage external volumes or mounts during container deployment.

Cons:

    Image updates: Requires rebuilding the image if the certificate needs to be updated or replaced.
    Security: Less secure if the image is shared publicly, as the certificate is exposed.

Mounting to the Container:

Pros:

    Flexibility: Allows for easy updates or replacement of the certificate without rebuilding the image.
    Security: Keeps the certificate external to the image, reducing exposure if the image is shared.
    Centralized management: Can be managed independently of the image, simplifying certificate updates across multiple containers.

Cons:

    Mounting complexity: Requires managing external volumes or mounts, adding complexity to deployment.
    Potential permissions issues: Ensure proper permissions are set for the mounted certificate within the container.

Best Practices:

    For sensitive environments: Prioritize security and flexibility by mounting the certificate.
    For frequent certificate updates: Mounting is more efficient.
    For simple deployments and self-contained images: Adding to the image might be suitable.
    Consider using Docker Secrets: Provides a secure mechanism for managing sensitive information like certificates.

Additional Considerations:

    Certificate expiration: Implement a process for monitoring and renewing certificates regardless of the chosen approach.
    Image distribution: If sharing images publicly, avoid embedding sensitive certificates.
    Volume drivers: Choose a suitable volume driver for mounting certificates, considering performance and security implications.

Ultimately, the best approach depends on your specific requirements and security considerations.

apps/schematic/api/Dockerfile
linglp: Thanks for the feedbacks Thomas. I decided to remove this line since it is not really useful

apps/schematic/api/docker-compose.yml
linglp: Thanks for the feedbacks. But I wonder what are the reasons that we don't build docker files locally using docker compose..

apps/schematic/api/Dockerfile
linglp: Thanks for the feedbacks. I did more googling and noticed that a lot people mounted the certificate instead of including the certificate in the docker image. But I am not sure how they managed deployment. Let me reach out to IT and see if adding a self-signed certificate to a Docker image creates security concerns

apps/schematic/api/Dockerfile
linglp: Yes I agree this is not a good practice. This line is temporarily there because schematic stores .synapseCache in root folder. After changes on schematic side get made, this line won't be needed.

apps/schematic/api/Dockerfile
linglp: Yes. I tried adding USER www-data in dockerfile but it didn't work because the base image: tiangolo/uwsgi-nginx-flask:python3.10 that we are using is using supervisord to bind pots. And binding pots in docker requires super user permission.

apps/schematic/api/Dockerfile
linglp: Thanks for the suggestion Thomas. I have added those lines

apps/schematic/api/docker-compose.yml
linglp: Thanks for responding to this question on slack. I will keep this part as pulling docker image from GHCR then.

apps/schematic/api/Dockerfile
linglp: Based on conversation with @tschaffter, better not to use ports below 1024 (does not include 1024). Next step is to check if I could configure nginx to use a different port than 443.

apps/schematic/api/docker-compose.yml
linglp: Based on conversation with @tschaffter , port 80 is also protected. Maybe tried to configure http portocol on 8080 (or anything above 1024)

apps/schematic/api/Dockerfile
linglp: @tschaffter thanks for the feedbacks again. I have updated the nginx config to use port 8443 instead of port 443. Is that okay? Or should I use port 7443?

apps/schematic/api/docker-compose.yml
linglp: @tschaffter marked this as resolved for now. I have updated nginx to use port 8080 for http request instead of port 80.

apps/schematic/api/Dockerfile
linglp: per conversation on slack, we need to use port 7443

apps/schematic/api/docker-compose.yml
linglp: I have reverted changes in docker-compose.yml other than updating the port number. Mark this as "resolved" for now.

apps/schematic/api/Dockerfile
linglp: @tschaffter thanks for your feedback. I have updated the Dockerfile and made sure that we pass certificate and private key as environment variables.

apps/schematic/api/Dockerfile
```diff
@@ -1,41 +1,82 @@
-FROM python:3.10.12-slim-buster
+FROM tiangolo/uwsgi-nginx-flask:python3.10
 
-ENV APP_DIR=/opt/app
+# add label 
+LABEL maintainer="Milen Nikolov, Andrew Lamb, Mialy DeFelice, Gianna Jordan, Lingling Peng"
```
tschaffter, line 4: Use the opencontainer standard for labels: https://github.com/opencontainers/image-spec/blob/main/annotations.md

.github/workflows/build-devcontainer-image.yml
tschaffter, line 1: Clean up

apps/schematic/api/docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
     networks:
       - schematic
     ports:
-      - "${SERVER_PORT}:7080"
+      - "7443:7443"
```
tschaffter, line 16: This change should be mirrored in `.devcontainer/devcontainer.json`

tschaffter (changes requested):
I skimmed thought the PR and made a few comment. I can make a formal review once the PR is ready (after cleanup and conflict resolution).

Does the command `nx serve schematic-api` listen to the new port? This is a convention of the monorepo so that developers can run the REST API either manually or with Docker while keeping the same behavior for the REST API.

apps/schematic/api/docker-compose.yml
linglp: Thanks for the reminder Thomas. I have updated devcontainer.json

apps/schematic/api/Dockerfile
linglp: Thanks Thomas. I tried reading through the documentation but I got a bit confused about what the "standard" format needs to be (in the documentation, they don't have any examples other than "contact details of the people or organization responsible for the image (freeform string)"). Is the standard something like: 
```
LABEL maintainer="A <email address of A>, B <email address of B>,"
```

tschaffter: Example (check value format from spec):

```
LABEL org.opencontainers.image.authors='airship-discuss@lists.airshipit.org, irc://#airshipit@freenode' \
      org.opencontainers.image.url='https://airshipit.org' \
      org.opencontainers.image.documentation='https://airship-armada.readthedocs.org' \
      org.opencontainers.image.source='https://opendev.org/airship/armada' \
      org.opencontainers.image.vendor='The Airship Authors' \
      org.opencontainers.image.licenses='Apache-2.0'
```

linglp: Thanks Thomas. I changed the label to: 
```
LABEL org.opencontainers.image.authors= 'A <email address of A>, B <email address of B>'
```

tschaffter: Thank you for adopting this standard labels. Yes, looks good to me. One potential improvement would be to split the label value over multiple lines. I'm not sure if our linter (hadolint) cares about this. I wasn't sure that it was actually possible but it looks like it is using this format based on this [article](https://dockerlabs.collabnix.com/beginners/dockerfile/Label_instruction.html):

```
LABEL description="This text illustrates \
that label-values can span multiple lines."
```

I'm resolving this comment now.

apps/schematic/api/debug.py
```diff
@@ -0,0 +1,76 @@
+from environs import Env
```
linglp, line 1: give the script a more descriptive name

apps/schematic/api/Dockerfile
```diff
@@ -1,41 +1,82 @@
...
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-get update \ 
+    && apt-get install -y openssl jq 
 
-WORKDIR ${APP_DIR}
-COPY schematic_api schematic_api/
-COPY pyproject.toml poetry.lock uwsgi.ini ./
+# add dhparam.pem
+# dhparam.pem was used in ssl-params.conf
+COPY ./dhparam.pem /etc/ssl/dhparam.pem
```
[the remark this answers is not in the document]
tschaffter: You should be able to replace `./dhparam.pem` by `dhparam.pem`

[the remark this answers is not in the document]
tschaffter: Prefer the most simple form when possible

apps/schematic/api/Dockerfile
```diff
@@ -1,41 +1,82 @@
...
   && poetry config --local virtualenvs.create false \
   && poetry run pip install "cython<3.0.0" \
   && poetry run pip install --no-build-isolation pyyaml==5.4.1 \
-  && poetry install --with prod --no-root --no-interaction --no-ansi \
-  && pip cache purge
+  && poetry install --with prod --no-root --no-interaction --no-ansi
+
+# Update file permission 
+RUN mkdir /root/.synapseCache
+RUN mkdir /app/app/manifests
```
tschaffter, line 58: Consecutive RUN instructions should be consolidated into a single one to adhere to the best practice of minimizing the number of image layers.

tschaffter (approved):

tschaffter:
@linglp You check is failing. The error message should point to the solution :-)

apps/schematic/api/README.md
```diff
@@ -11,39 +11,41 @@ This example uses the [Connexion](https://github.com/zalando/connexion) library
...
+and open your browser to here:
 
-To launch the integration tests, use tox:
 ```
-sudo pip install tox
-tox
+http://127.0.0.1:7080/api/v1/ui/
 ```
 
 ## Running with Docker
```
linglp, line 34: @tschaffter I updated the documentation in the README.md like we discussed today. Does this look good? The nx command should be sufficient to build the docker image and run the container

apps/schematic/api/Dockerfile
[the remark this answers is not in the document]
linglp: done. thank you for the suggestion

apps/schematic/api/README.md
tschaffter: The instructions should be even more simple using the Nx API. Developers should run Nx tasks with `nx <task> <project>` or `nx run <project>:<task>` with non-nx commands being exceptions.

1. Prepare the development environment of the project with `nx prepare schematic-api`. This will create a venv and install all the Python dependencies. We could even extend this task to run your setup script automatically.

2. Start the REST API with `nx serve schematic-api`

3. Build the image with `nx build-image schematic-api`

4. Start the containerized REST API with `nx serve-detach schematic-api` (the image must be built first)

apps/schematic/api/Dockerfile
linglp: Yes. I will do that. Thanks for the feedback.

apps/schematic/api/debug.py
tschaffter: LGTM

apps/schematic/api/README.md
linglp: Thanks Thomas. I updated the documentation and created two sessions: "running schematic APIs with Docker" and "running schematic APIs without Docker" . Can I also remove the overview related to "OpenAPI generated server" ? I image developers will just read the documentation to figure out running schematic APIs? And the documentation related to "OpenAPI generated server" won't be as useful?

tschaffter: Yes

linglp:
@tschaffter If the PR looks good, maybe you could merge it on your end? I am not authorized to merge the PR. Thank you again for reviewing it

andrewelamb (approved):

PR #2399, opened 2023-12-08
fix(openchallenges): update OC data model and API specs

vpchung:
Fixes #2064 

## Changelog
* SQL
   * allow `avatar_key`s to be NULL (not all platforms/organizations/challenges will have an image)
   * update max characters for `website_url` and `avatar_url` to 500 characters ([ref](https://stackoverflow.com/questions/219569/best-database-field-type-for-a-url))
   * update max characters for `challenge.doi` to 120 characters - to be reasonably more generous with URLs
   * update max characters for`organization.acronym` to 10 characters ([ref](https://community.developers.refinitiv.com/questions/25495/what-is-the-max-length-of-a-field-name-acronym-in.html))
   * add constraint that `challenge.slug` must be unique, in kebab-case, and at least 3 characters 
   * add constraint that `organization.login` must be unique, in kebab-case, and at least 2 characters 
   * add constraint that `challenge.name` cannot be NULL
   * add constraint that combination of `challenge_id`, `organization_id`, and `role` must be unique in `challenge_contribution` tables
   * set default value for `organization.challengeCount` to 0
   * remove redundant constraint of `id` having to be unique - they are set as primary keys so they should already be unique
   * remove deprecated columns (`challenge.difficulty`, `organization.email`)
   * general cleanup (fix formatting, remove commented out lines, etc)
   * update dump files (using latest data from #2394 
* API descriptions
   * add schemas for ChallengeDoi, CreationDatetime, ModifiedDatetime, Url 
   * remove unneeded schemas: ChallengeDifficulty, ChallengeReadme*
   * add missing descriptions to Organization properties, e.g. Organization.name
   * update API models to reflect changes noted above in SQL ^  e.g. specify certain properties as nullable, update maxLength, set default values, etc.
 * Web app
    * update mock objects, search queries and HTML files to not include deprecated properties, e.g. difficulty and email

> [!NOTE]
> There were some updates made to the Java scripts as well, as not updating them prevented the challenges to display on the web app.  This was done with the generator: `nx run openchallenges-*-service:generate`

## Preview

**No changes to search pages e.g.**
Screenshot 2023-12-07 at 5 03 13 PM

**Organization profile no longer displays contact email:**
Screenshot 2023-12-07 at 5 03 24 PM

**API spec for challenges:**
Screenshot 2023-12-07 at 5 06 12 PM

**API spec for organizations:**
Screenshot 2023-12-07 at 5 06 22 PM

PR #2410, opened 2023-12-13
chore(openchallenges): 2023-12-13 OC DB update

vpchung:
> [!NOTE]
> This branch was checked out using #2399 as the starting point.  I suggest merging in #2399 before merging in this one.

- Depends on #2399

## Changelog
* Added:
   * 4 challenges from EvalAI, 1 from CodaBench
   * 8 organizations
   * missing description of a challenge
* Updated:
   * general challenge and organization updates, e.g. challenge counts, last-modified timestamps, etc.
* Fixed:
   * platform ids 

## Preview
**Latest landscape**
Screenshot 2023-12-13 at 2 02 08 PM

**Search results for CodaBench**
Screenshot 2023-12-13 at 2 02 40 PM

**Search results for EvalAI**
Screenshot 2023-12-13 at 2 02 51 PM


gaiaandreoletti (approved):

tschaffter:
@vpchung I made this PR dependent on #2399 with [Dpulls](https://www.dpulls.com/) (see checks)

apps/openchallenges/organization-service/src/test/java/org/sagebionetworks/openchallenges/organization/service/OpenApiGeneratorApplicationTests.java
tschaffter, line 1: @tschaffter: Remove this file and prevent future generation

libs/openchallenges/api-description/src/components/schemas/CreationDatetime.yaml
tschaffter, line 1: @tschaffter Remove now that we have `CreatedDatetime`

tschaffter (changes requested):

rrchai (approved):
LGTM, other than @tschaffter's comments.

libs/openchallenges/api-description/src/components/schemas/CreationDatetime.yaml
```diff
@@ -0,0 +1,4 @@
+description: Datetime when metadata was added to the OC database.
```
tschaffter, line 1: A more reusable description would be (to apply to `CreatedDatetime`):

> Datetime when this object was added to the database.

Similar comment probably applies to `ModifiedDatetime`

rrchai:
@tschaffter two extra files are removed.

libs/openchallenges/api-description/src/components/schemas/CreationDatetime.yaml
rrchai: I agree. I ideally we should have both `CreatedDatetime` (when it's initially  added to the data) and `ModifiedDatetime`. By the way, do `CreatedDateTime` looks better than `CreatedDatetime` to you? 

I would prefer @vpchung to alter the metadata spreadsheet next year~ I'm afraid to break something in data :).

tschaffter: > By the way, do CreatedDateTime look better than CreatedDatetime to you?

Yes. Also we could rename `ModifiedDatetime` to `UpdatedDateTime` since we use it for the property `updatedAt`.

tschaffter: Done

tschaffter (approved):

PR #2417, opened 2023-12-15
feat(schematic): modified docker image when running schematic APIs with uWSGI

linglp:
Related to: https://sagebionetworks.jira.com/browse/FDS-1346

## Context
The goal of this PR is to make sure schematic can work well with uWSGI. 

Even though a lot of content in this docker image would need to get changed when adding nginx, I still want to make a PR to make sure that we have `schematic+uwsgi` working if there's anything wrong with nginx and we want to trouble shoot by isolating uswgi and nginx. 

## Changelog
* Thomas already have uswgi working with schematic APIs. To make sure that all the endpoints work properly, I modified permission of some folders. I previously did it like this PR: https://github.com/Sage-Bionetworks/schematic/pull/1142/files#diff-d04806eafd231d6405de22d8c29951c269796bcea02a3fc8116c48f33e96ac6d but I think the way that i have now is clearer. 
* I updated the uswgi.ini so that it looks like our current set up in schematic. 

## Test
Get into the right dir
```
cd apps/schematic/api
```
Build a docker image: 
```
docker build -t schematic_api .
```
Run docker container: 
```
docker run -p 7080:7080 schematic_api
```
Should then be able to visit `http://localhost:7080/api/v1/ui/#/`
I tested all the endpoints and make sure that they could return expected output.

### DIFF
--- a/apps/schematic/api/Dockerfile
+++ b/apps/schematic/api/Dockerfile
@@ -27,6 +27,23 @@ RUN pip install poetry \
   && poetry install --with prod --no-root --no-interaction --no-ansi \
   && pip cache purge
 
+
+WORKDIR /var/www/
+# create user group www-data
+# and create www-data as a user
+RUN usermod -aG www-data www-data
+# change the group ownership of /var/www, /opt/app recursively to www-data
+# /opt/apt folder is relevant for storing manifests when executing endpoints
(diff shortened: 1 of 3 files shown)


linglp:
Since Nginx has gotten added. This PR is no longer needed.

PR #2421, opened 2024-01-03
feat(schematic): fds 1447: implement pagination

andrewelamb:
- fixes [fds-1447](https://sagebionetworks.jira.com/browse/FDS-1447)
- For all endpints that return an array of items, there are now 2 endpoints:
  - A non-paginated version "*Array"
  - A paginated verison "*Page"
- Pagainated versions implement a limited version of pagination:
  - Users can set their amount per page to any amount
  - Pagiantion does not use tokens, changes to the underlying data in between queries will have incorrect results
  - Assumes schematic methods are deterministic(don't result in a random oreder)
  - Gets entire result of schematic method and breaks this up over several queries
- Non-paginated queries don't have a cap on results
- Endpoints that return data about a synapse entity (as opposed to the etnity itself such a a manifest) have been renamed Metadata
- For code review check:
  -  controllers/paging.py
  -  controllers/schema_controller_impl.py
  - controllers/storage_controller-impl.py
- endpoints paginated [old;**new**]
   - GET /storage/project/datasets; **/assetTypes/{assetType}/projects/{projectId}/datasets**
   - GET /storage/project/manifests; **/assetTypes/{assetType}/projects/{projectId}/manifests**
   - GET /storage/projects; **/assetTypes/{assetType}/assetViews/{assetViewId}/projects**
   - GET /storage/dataset/files; **/assetTypes/{assetType}/assetViews/{assetViewId}/datasets/{datasetId}/files**
   - GET /explorer/get_node_dependencies; **nodes/{nodeLabel}/dependencies**
   - GET /schemas/get_graph_by_edge_type; **/connectedNodePair**


apps/schematic/api/schematic_api/controllers/schema_controller_impl.py
```diff
@@ -217,118 +250,171 @@ def get_schema_attributes(schema_url: str) -> tuple[Union[str, BasicError], int]
...
+) -> list[NodeProperty]:
     """Gets the properties associated with the node
 
     Args:
         schema_url (str): The URL of the schema in jsonld form
         node_label (str): The label of the node
 
     Returns:
-        list[str]: A list of properties of the node
+        list[NodeProperty]: A list of properties of the node
```
mialy-defelice, line 261: I am having a little trouble following what is happening here... is NodeProperty an object or class? And you're adding properties to it, then that is in a list?

mialy-defelice (commented):
Overall it looks good, but I am a bit confused by the new arrays. I am also not so sure we need pagination for the Validation Rules or properties.

apps/schematic/api/schematic_api/controllers/schema_controller_impl.py
andrewelamb: Yeah, NodeProperty is a class, that currently just just contains the name of the node. I noticed its exactly the same as the node class. I think after the schema refactor has been released, but before the API release I'm goign to go through and clean up some of th names of the classes.

andrewelamb: You're correct this is confusing, I'm going to fix this.

andrewelamb:
@mialy-defelice Do you think we should just remove the paginated versions of validation rules and properties?

tschaffter:
@andrewelamb @milen-sage Would you be interested in trying to standardize the pagination specification between OpenChallenges REST API and FAIR Data REST API?

milen-sage:
Interested - yes, but we need to determine scope and priority.

andrewelamb:
@mialy-defelice I removed pagination form node properties and validation rules endpoints, and changed node property forma  class to a string

apps/schematic/api/schematic_api/controllers/schema_controller_impl.py
```diff
@@ -340,25 +346,26 @@ def get_node_dependencies(
...
 
     """
     schema_generator = SchemaGenerator(path_to_json_ld=schema_url)
-    return schema_generator.get_node_dependencies(
+    nodes = schema_generator.get_node_dependencies(
         source_node=node_label,
         display_names=return_display_names,
         schema_ordered=return_ordered_by_schema,
     )
+    return [Node(node) for node in nodes]
```
mialy-defelice, line 358: For all the endpoints, I have the same comment as I did for Node Properties...its confusing that its recorded as a list of nodes, bc nodes does not have an implicit type.. here it is `str` but in another context it could be a graph object... I feel like this nomenclature is obscuring what users should expect as an output. Im not familiar with this mode of typing so im not suggesting you change everything, but could you explain the choice further so I can understand it?

andrewelamb: I'm open to changing the name of the node class here. Its described here.. So just to clairify it is not a string, it is a class that has an attribute "name" which is a string. It's designed this way so that it can be extended in the future if we want to supply more informaiton about the node, and to be used by other endpoints.

PR #2428, opened 2024-01-17
feat(schematic): fds 1447: implement pagination (#2421)

andrewelamb:
fixes [fds-1447](https://sagebionetworks.jira.com/browse/FDS-1447)

For all endpints that return an array of items, there are now 2 endpoints:

- A non-paginated version "*Array"
- A paginated verison "*Page"

Pagainated versions implement a limited version of pagination:

- Users can set their amount per page to any amount
- Pagiantion does not use tokens, changes to the underlying data in between queries will have incorrect results
- Assumes schematic methods are deterministic(don't result in a random oreder)
- Gets entire result of schematic method and breaks this up over several queries
- Non-paginated queries don't have a cap on results
- Endpoints that return data about a synapse entity (as opposed to the etnity itself such a a manifest) have been renamed Metadata


mialy-defelice (commented):
Previously reviewed, so just approving.

tschaffter (approved):
Approved changes to the CI workflow (change to file format)

mialy-defelice (approved):

PR #2440, opened 2024-01-23
feat(agora): update `agora-mongo` image to include Agora dataset

tschaffter:


tschaffter:
@sagely1 Here is the code that we wrote together + cleanup.

tschaffter:
@sagely1 Something we may have overlooked: when building the docker image, MongoDB is not running, so the dataset files can't be load with a MongoDB CLI. Is there a way to export the Agora dataset files into a file format so that MongoDB can understand these data when it's instance is starting? Other the data processing may need to be done outside of the MongoDB Dockerfile.

Issue #2439, opened 2024-01-23
[Story] Review integration of Sonar with PRs


tschaffter:
### What product(s) is this story for?

Sage Monorepo

### Description

Review the current integration of Sage Monorepo with Sonar. Especially in the context of reviewing PRs.

### Have you linked this story to a GitHub Project?

- [X] I have linked this story to a GitHub Project and set its metadata.

tschaffter:
I configured the `main` branch so that the following check is required:

![image](https://github.com/Sage-Bionetworks/sage-monorepo/assets/3056480/dd0fcbb2-d8c3-4b06-8cae-fb77c0c321bb)

All PRs would now include a check that would only pass if the project `schematic-api` was to push a report to Sonar and if the report satisfies the code quality gateway configured for this project.

<img width="912" alt="image" src="https://github.com/Sage-Bionetworks/sage-monorepo/assets/3056480/de781383-fefd-4b33-acce-16d14368a44a">

In the example above, the project `schematic-api` is not modified, so no report will be pushed to Sonar and the check will never pass. So we can't make this specific check required.

tschaffter:
## How to enable feedback from Sonar in PR thread?

See this [example](https://github.com/Sage-Bionetworks/schematic/pull/1351) from Schematic:

![image](https://github.com/Sage-Bionetworks/sage-monorepo/assets/3056480/8788122a-0329-420a-9a9b-5ce3fb0899d7)

Is it when the check shown in the above post is executed and completed?

tschaffter:
## Who can trigger Sonar scans and push reports to SonarCloud.io

Currently only pushing to the `main` branch triggers the scan with sonar. The last commit pushed to `main` that modified `schematic-api` created the check "[schematic-api] SonarCloud Code Analysis" (see screenshot).

<img width="652" alt="image" src="https://github.com/Sage-Bionetworks/sage-monorepo/assets/3056480/712fd747-b7c8-49d0-9c6d-b52283c3ad66">

tschaffter:
## Permission of the app SonarCloud

<img width="516" alt="image" src="https://github.com/Sage-Bionetworks/sage-monorepo/assets/3056480/a3975d77-2d89-490f-8483-80ed05612db3">

tschaffter:
## Experiment: remove sonar scan from CI workflow

[...]

PR #2449, opened 2024-01-24
ci(sage-monorepo): will SonarCloud GitHub app annotate this PR?

tschaffter:
Contributes to #2439


tschaffter:
The SonarCloud app was not able to post the result of the scan - or even trigger it - because this feature is not supported for monorepo.

> [Multiple projects bound to a single repository](https://docs.sonarsource.com/sonarcloud/advanced-setup/monorepo-support/) (the monorepo strategy) are not supported.

PR #2480, opened 2024-01-31
feat(schematic): fds 828 generate manifest endpoint

andrewelamb:
- Fixes [fds-828](https://sagebionetworks.jira.com/browse/FDS-828)
- Adds the generate google sheet manifest endpoint
- Adds step in schematic specific ci workflow that daves google creds


.github/workflows/schematic-api-ci.yml
```diff
@@ -119,6 +119,18 @@ jobs:
               yaml.dump(secrets, file)
         shell: python
 
+      - name: create google credentials
```
tschaffter, line 122: - "Set up" instead of "Create" would be more accurate
- Capitalize first letter of the step name

tschaffter: Does the scope of the credentials adhere to the principle of least privilege? I.e. does the credentials only give access to the Google resources used by this workflow?

andrewelamb: @tschaffter good question!
@linglp These are google crednetials I created the instructions in the schematic readme.  Are these appropriate for this purpose?

libs/schematic/api-description/src/openapi.yaml
```diff
@@ -106,3 +106,6 @@ paths:
 
   /tangledTreeText:
     $ref: paths/tangledTreeText.yaml
+
```
linglp, line 109: For generating an existing manifest, user would need to put in access token like other endpoints (because schematic would need to pull manifests from a dataset ID provided by the user). Could you add a way for user to add their synapse token like other APIs?

apps/schematic/api/schematic_api/test/test_synapse_endpoints.py
```diff
@@ -38,6 +38,29 @@
...
 
+@pytest.mark.synapse
+@pytest.mark.secrets
+class TestGenerateGoogleSheetManifests(BaseTestCase):
+    """Tests google sheet manifest endpoint"""
+
+    def test_success(self) -> None:
+        """Test for successful result"""
+        url = (
+            f"/api/v1/generateGoogleSheetManifests?schemaUrl={TEST_SCHEMA_URL}"
```
linglp, line 49: I don't think here the API is interacting with synapse because dataset_id is not used as a parameter

linglp (changes requested):
when `schematic_service_account_creds` does not exist locally, currently the APIs return a 500 error. Can we let the APIs return 4XX error?

apps/schematic/api/schematic_api/controllers/manifest_generation_controller_impl.py
```diff
@@ -0,0 +1,103 @@
...
+            raise InvalidValueError(
+                "When generate_all_manifests is True dataset_id_array must be None",
+                {"dataset_id_array": dataset_id_array},
+            )
+        if node_label_array:
+            raise InvalidValueError(
+                "When generate_all_manifests is True node_label_array must be None",
+                {"node_label_array": node_label_array},
+            )
+        node_label_array = ["all_manifests"]
```
linglp, line 65: here, if should be "all manifests" instead of "all_manifests". Otherwise schematic would error out.

linglp:
@andrewelamb I tested by generating new manifests with single data type or multiple datatypes. They both worked. But when I tried "generateAllManifests" option, it didn't work and generating an existing manifests also didn't work. But they should be easy to fix. After you address all the comments, I will test again.

.github/workflows/schematic-api-ci.yml
```diff
@@ -119,10 +119,22 @@ jobs:
               yaml.dump(secrets, file)
         shell: python
 
+      - name: Set up google credentials
```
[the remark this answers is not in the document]
andrewelamb: Two reasons:
 I couldn't get it to work as an environmental variable. :)
This is also more consistent so that local development, and this github workflow work the same by taking in google credentials as a file.

libs/schematic/api-description/src/components/schemas/NodeLabelArray.yaml
```diff
@@ -0,0 +1,4 @@
+description: An array of node labels
```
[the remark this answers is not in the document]
andrewelamb: Is'nt node labels more consistent with other endpoints? The problem with using data is that it doesn't tell the user if they shoulkd be using the node name or display label

libs/schematic/api-description/src/openapi.yaml
andrewelamb: Yep, I must have forgotten to add that!

apps/schematic/api/schematic_api/controllers/manifest_generation_controller_impl.py
andrewelamb: Yes you are right, I thought it was the other way around

.github/workflows/schematic-api-ci.yml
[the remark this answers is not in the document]
linglp: That's fair. I can't see anything wrong by using schematic_service_account_creds.json file. I was just curious about your reasoning of using it.

PR #2521, opened 2024-02-22
feat(openchallenges): add EDAM concept table to the chalenge service

### LINKED ISSUE #2520: [Task] Add EDAM ontology to the OC Challenge Service
Identify a workflow for integrating EDAM into OC. The workflow should be documented and minimize manual operations.

> [!WARNING]  
> The [EDAM browser](https://bioportal.bioontology.org/ontologies/EDAM/?p=classes&conceptid=http://edamontology.org/data_2366) uses an unstable version of EDAM (see Summary section).

tschaffter:
Closes #2520

## Description

Add a minimal version of EDAM ontology as a new SQL table in the challenge service DB.

This contribution is a refactored version of #2265 that aims to integrate EDAM into OC one step at a time.

In the near future, I can see us using the following workflow, which the present PR brings us close to.

1. We fork the repo https://github.com/edamontology/edamontology to prevent data from being deleted
2. We have an ETL workflow that
    - E: download the CSV file from the EDAM fork
    - T: a python script that converts the CSV file into a format that OC can work with; optionally validate the input and output data.
    - L: load the processed data into OC DB

Why? The entire EDAM ontology is more than 2MB and too large to be tracked in Sage Monorepo. Using an ETL workflow would allow us to store the raw data externally (e.g. S3 bucket, Synapse).

I actually have implemented this workflow for iAtlas a few weeks ago (see `apps/iatlas/data`). This workflow could be reused almost as is with OC. Note that the project `iatlas-data` uses Schematic to validate the iAtlas data, which is a feature that @vpchung suggested recently for OC.

## Changelog

- add minimal EDAM ontology to OC Challenge Service
- document in the ticket how the processed EDAM CSV file is generated with minimal manual operations
- add the Google sheet `OC_EDAM_1.25` to OC shared folder in Google Drive

## Testing

- The new SQL table is created and populated when starting the challenge service
  - [x] using Docker
  - [x] using the development server

## Future Work

Add the property `Challenge.operation`

## Preview

image


apps/openchallenges/challenge-service/src/main/resources/db/edam_concept.csv
```diff
@@ -0,0 +1,3474 @@
+"id","class_id","preferred_label"
```
vpchung, line 1: Can we additionally grab the `Parents` column?  IIRC, we had a discussion about building an ontology-like search in the OC, so getting the parent EDAM classes would be required.

apps/openchallenges/challenge-service/src/main/resources/application.yml
```diff
@@ -18,3 +18,4 @@ spring:
       db_incentives_csv_path: ${DB_INCENTIVES_CSV_PATH:src/main/resources/db/incentives.csv}
       db_submission_types_csv_path: ${DB_SUBMISSION_TYPES_CSV_PATH:src/main/resources/db/submission_types.csv}
       db_categories_csv_path: ${DB_CATEGORIES_CSV_PATH:src/main/resources/db/categories.csv}
+      db_edam_concept_csv_path: ${DB_EDAM_CONCEPT_CSV_PATH:src/main/resources/db/edam_concept.csv}
```
vpchung, line 21: For my own clarify, the automated CSV dumping script should no longer dump [this sheet](https://docs.google.com/spreadsheets/d/1XRSUVqbZk_k9_rV1vpgPVMQ0fVV0mavwY8Aqyy0C2HQ/edit#gid=1849016276), correct?

vpchung (changes requested):
Minor request, unless it's no longer relevant.

apps/openchallenges/challenge-service/src/main/resources/application.yml
tschaffter: Correct

apps/openchallenges/challenge-service/src/main/resources/db/edam_concept.csv
vpchung: Sounds good - another item for our backlog! 😅

vpchung (approved):
Works on my end - thanks!

PR #2523, opened 2024-02-22
docs(sage-monorepo): add docs about co-authors

tschaffter:
## Changelog

- document how to add co-authors to commits


vpchung (approved):

vpchung:
Did we also want to expose this on our docs site?  If so, we would need to [add it to `mkdocs.yml`](https://github.com/Sage-Bionetworks/sage-monorepo/blob/2814e882506f535db1d2f1744619aa46a473138a/mkdocs.yml#L11).

PR #2531, opened 2024-02-23
feat(openchallenges): dump only EDAM annotations

vpchung:
Complimentary PR to #2521 (more context [here](https://github.com/Sage-Bionetworks/sage-monorepo/pull/2521#discussion_r1499886099))

Now that we are grabbing the EDAM ontology directly from source, there is no longer a need for us to dump a CSV file.

## Changelog
* update CSV dumping script to only dump the EDAM annotations
* remove the unneeded function


vpchung:
Closing this PR to rename the branch and include other changes to the dumping script

Issue #2256, opened 2023-10-20
[Story] Add EDAM annotations (data and full-text search)


vpchung:
Following up on #2209, add EDAM annotations curated by @gaiaandreoletti to the OC db.

Tasks:

- [x] #2520
- [x] Add a simplified version of EDAM to OC DB
- [x] Add the property `Challenge.operation` to OC API model (EDAM term)
- [x] #2551

tschaffter:
Added to Sprint 10.11

tschaffter:
Actually I'm sending that one back to Backlog. We need to better validate the existing data (#2064) before we can add more properties to the data mode.

tschaffter:
## Update 2024-03-04

The API property `Challenge.operation` is now described with EDAM Operation. We have one more task about representing challenge input data type with EDAM before be can close this story.

tschaffter:
This task is done. There is a pending task tracked by @gaiaandreoletti that consist in making sure that only "Data" concepts are used for the challenge input data types. The existing implementation supports this decision.

@gaiaandreoletti We are welcome to revisit this decision any time, e.g. if you want to start using ["Format" or "Topic" concepts](https://edamontology.org/page) to annotate the input data types of challenges.

gaiaandreoletti:
only data concepts of edam were added to the challenge format

PR #2534, opened 2024-02-26
feat(openchallenges): add API schema property `Challenge.operation`

tschaffter:
Contributes to #2256

## Changelog

- add the API schema `EdamOperation`
- add the API schema property `Challenge.operation` 
- update the challenge service to support the new Challenge property
- updated the API client for Angular

## Notes

- A previous version of this PR was created by @vpchung in #2264.
- The operation value added to challenge with ID 1 is fake; it's expected to be replaced in future data dumps.
- Three invisible characters observed in `challenges.csv` have been replaced by a white space (" "). Future data dumps should ideally no longer include these invisible characters.
- The challenge operation is not indexed in ES (future PR).

## Future work

For future PRs:

- Display the challenge operation on the challenge profile page
- Represent challenge input data types with EDAM Data concepts
- Implement fuzzy search for challenge operation and input data types (full-text search)
- Implement precise search for challenge operation and input data types (e.g., checkboxes)

### When `challenge.operation_id` is specified in the DB

The value of `Challenge.operation` returned by the REST API is an `EdamOperation` object.

```
GET {{basePath}}/challenges/1

{
  "id": 1,
  "slug": "network-topology-and-parameter-inference",
  "name": "Network Topology and Parameter Inference",
  "headline": "Optimize methods to estimate biology model parameters",
  "description": "Participants are asked to develop and/or apply optimization methods, including the selection of the most informative experiments, to accurately estimate parameters and predict outcomes of perturbations in Systems Biology models.",
  "doi": "",
  "status": "completed",
  "platform": {
    "id": 1,
    "slug": "synapse",
    "name": "Synapse"
  },
  "websiteUrl": "https://www.synapse.org/#!Synapse:syn2821735",
  "avatarUrl": "",
  "incentives": [
    "publication"
  ],
  "submissionTypes": [
    "prediction_file"
  ],
  "inputDataTypes": [
    {
      "id": 1,
      "slug": "genomic",
      "name": "genomic"
    },
    {
      "id": 2,
      "slug": "proteomic",
      "name": "proteomic"
    }
  ],
  "categories": [],
  "startDate": "2012-06-01",
  "endDate": "2012-10-01",
  "starredCount": 2,
  "operation": {
    "classId": "http://edamontology.org/operation_0232",
    "preferredLabel": "Sequence merging"
  },
  "createdAt": "2023-11-15T22:40:15Z",
  "updatedAt": "2024-02-19T18:17:46Z"
}
```

### When `challenge.operation_id` is NULL in the DB

The value of `Challenge.operation` returned by the REST API is `null`.

```
GET {{basePath}}/challenges/2

{
  "id": 2,
  "slug": "breast-cancer-prognosis",
  "name": "Breast Cancer Prognosis",
  "headline": "Predict breast cancer survival from clinical and genomic data",
  "description": "The goal of the breast cancer prognosis Challenge is to assess the accuracy of computational models designed to predict breast cancer survival, based on clinical information about the patient's tumor as well as genome-wide molecular profiling data including gene expression and copy number profiles.",
  "doi": "",
  "status": "completed",
  "platform": {
    "id": 1,
    "slug": "synapse",
    "name": "Synapse"
  },
  "websiteUrl": "https://www.synapse.org/#!Synapse:syn2813426",
  "avatarUrl": "",
  "incentives": [
    "publication",
    "speaking_engagement"
  ],
  "submissionTypes": [
    "prediction_file"
  ],
  "inputDataTypes": [
    {
      "id": 1,
      "slug": "genomic",
      "name": "genomic"
    }
  ],
  "categories": [],
  "startDate": "2012-07-12",
  "endDate": "2012-10-15",
  "starredCount": 1,
  "operation": null,
  "createdAt": "2023-11-14T20:36:32Z",
  "updatedAt": "2024-02-19T18:17:47Z"
}
```


tschaffter:
## Error while populating the DB

```
2024-02-26 22:00:07 WARN  [restartedMain] o.f.c.i.s.DefaultSqlScriptExecutor - DB: Cannot add or update a child row: a foreign key constraint fails (`challenge_service`.`challenge`, CONSTRAINT `challenge_ibfk_2` FOREIGN KEY (`operation_id`) REFERENCES `edam_concept` (`id`)) (SQL State: 23000 - Error Code: 1452)
2024-02-26 22:00:07 WARN  [restartedMain] o.f.c.i.s.DefaultSqlScriptExecutor - DB: Incorrect integer value: '' for column `challenge_service`.`challenge`.`operation_id` at row 28 (SQL State: HY000 - Error Code: 1366)
```

tschaffter:
## Invisible character

There is an invisible character that throws Rainbow CSV off. This PR fixed it.

image

gaiaandreoletti (approved):

libs/openchallenges/api-description/build/challenge.openapi.yaml
```diff
@@ -463,6 +463,22 @@ components:
       format: date
       nullable: true
       example: '2017-07-21'
+    EdamOperation:
```
vpchung, line 466: Nit: include `nullable: true`?

apps/openchallenges/challenge-service/src/main/resources/openapi.yaml
```diff
@@ -682,6 +682,25 @@ components:
       format: date
       nullable: true
       type: string
+    EdamOperation:
```
vpchung, line 685: Nit: include `nullable: true`?

libs/openchallenges/api-description/build/challenge.openapi.yaml
tschaffter: I agree with your suggestion but this is a generated file that we can't edit manually.

Implementing the suggestion at the source level is a bit tricky. The reason is that `$ref:` doesn't combine well with `nullable:`, which is at least true for our version of OpenAPI (3.0.x).

There is a workaround with `AllOf` that I tried in the past, but this leads to issues with some generators (Python?). The alternative would be to add `nullable:` to the schema itself, i.e. `EdamOperation.yaml`. But a property may not always accept that its value may be `null`, so `nullable:` should ideally be defined independently for each variable, and not force all the variables to accept a `null` value, which would happen if we add `nullable: true` to `EdamOperation.yaml`. Does that make sense?

Recently I kind of settled on not specifying `nullable:`. Instead I used `required:` to specify whether a property should be expected. Here, `Challenge.operation` is not listed as required so the user should not expect to always receive this value, whether the property is missing or its value is `null`.

There are three actions that we could take:

1. Do nothing; keep the current implementation
2. Add `nullable: true` to `EdamOperation.yaml`: we only have one property that references this schema so why bother?
3. Create two versions of the schema: `EdamOperation.yaml` with `nullable: false` and `NullableEdamOperation.yaml` (or `EdamOperationOrNull.yaml`) with `nullable: true`.

apps/openchallenges/challenge-service/src/main/resources/openapi.yaml
tschaffter: See my reply to the other comment.

This file is auto-generated too.

tschaffter:
See comment from @vpchung ^

When specifying `nullable: true`, the docs becomes more accurate.

image

libs/openchallenges/api-description/build/challenge.openapi.yaml
vpchung: > Add nullable: true to EdamOperation.yaml: we only have one property that references this schema so why bother?

Adding `nullable: true` to the properties that can be null is the approach used with the other null-acceptable properties, e.g. SimpleChallengePlatform:

Screenshot 2024-02-29 at 9 56 17 AM

My inclination would to stay consistent and go with option 2. (I would very much like to avoid option 3 😅 )

andrewelamb (changes requested):

apps/schematic/api/project.json
```diff
@@ -88,17 +88,17 @@
...
+    //   "executor": "nx:run-commands",
+    //   "options": {
+    //     "commands": [
+    //       "poetry run black ./schematic_api --exclude 'models'",
+    //       "poetry run mypy --non-interactive --install-types --disallow-untyped-defs schematic_api/controllers/*impl.py schematic_api/test/test*.py",
+    //       "poetry run pylint schematic_api/controllers/*impl.py schematic_api/test/test*.py"
+    //     ],
+    //     "cwd": "apps/schematic/api"
+    //   }
+    // },
```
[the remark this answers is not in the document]
tschaffter: The lint issue is gone. The task `schematic-api:lint` is no longer commented out in this PR.

libs/openchallenges/api-description/build/challenge.openapi.yaml
tschaffter: @vpchung I added `nullable: true` to the OpenAPI schema `EdamOperation.yaml`

PR #2541, opened 2024-03-01
chore(openchallenges): update CSV dumping script

vpchung:
Complimentary PR to #2521 (more context [here](https://github.com/Sage-Bionetworks/sage-monorepo/pull/2521#discussion_r1499886099)) and #2534

Now that we are grabbing the EDAM ontology directly from source, there is no longer a need for us to dump a CSV file for EDAM terms.

## Changelog
* update CSV dumping script to not dump the CSV file for EDAM terms
* remove the unneeded function
* update CSV dumping script to handle select invisible unicode characters (more context [here](https://github.com/Sage-Bionetworks/sage-monorepo/pull/2534#issue-2154814288))
* include `operation_id` in the CSV dump, using NULL as the default value

### DIFF
--- a/apps/openchallenges/db-update/update_db_csv.py
+++ b/apps/openchallenges/db-update/update_db_csv.py
@@ -53,6 +53,7 @@ def get_challenge_data(wks, sheet_name="challenges"):
             "doi",
             "start_date",
             "end_date",
+            "operation_id",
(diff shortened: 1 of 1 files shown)


tschaffter (approved):
LGTM

PR #2553, opened 2024-03-06
fix(openchallenges): display zero search results

### LINKED ISSUE #2468: [Bug] Let the user know that there are no results (regression)

OpenChallenges

### Current behavior

We have recently decided to remove the (0) from "Results (0)" to prevent a flickering due to "Results (0)" to "Results (N)". However, it is no longer clear that a search resulted in 0 elements being found.

image

### Expected behavior

The user should be clearly notified that the search resulted in 0 elements found. This could be achieve by adding "(0)" back BUT ONLY after a search, i.e. "(0)" should not be displayed right away upon arriving on the search page.

An alternative solution would be to display a text message below Results. E.g. here is what Amazon shows:

> No results for asd asd ada da dasd.
> Try checking your spelling or use more general terms

### Anything else?

_No response_

### Commit ID

a82867c5fb593a9fc338239ffeb901fbccc126a0

### Code of Conduct

rrchai:
- fixes #2468 

### Changelog
- not display "Results(N)" when page initially load (when its value is undefined)
- display "Results(0)" if the no search results found.

### DIFF
--- a/libs/openchallenges/challenge-search/src/lib/challenge-search.component.html
+++ b/libs/openchallenges/challenge-search/src/lib/challenge-search.component.html
@@ -150,8 +150,8 @@
       </p-panel>
     </div>
     <div class="main col">
-      <h3>
-        Results <span *ngIf="challenges.length > 0">({{ searchResultsCount }})</span>
+      <h3 *ngIf="searchResultsCount !== undefined">
+        Results <span>({{ searchResultsCount }})</span>
       </h3>
       <div class="card-group">
         <openchallenges-challenge-card
--- a/libs/openchallenges/challenge-search/src/lib/challenge-search.component.ts
+++ b/libs/openchallenges/challenge-search/src/lib/challenge-search.component.ts
@@ -110,7 +110,7 @@ export class ChallengeSearchComponent
 
   challenges: Challenge[] = [];
   totalChallengesCount = 0;
-  searchResultsCount = 0;
+  searchResultsCount!: number;
 
   @ViewChild('calendar') calendar?: Calendar;
   customMonthRange!: Date[] | undefined;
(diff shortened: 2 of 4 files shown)


libs/openchallenges/challenge-search/src/lib/challenge-search.component.html
```diff
@@ -150,8 +150,8 @@ <h2>Challenges</h2>
       </p-panel>
     </div>
     <div class="main col">
-      <h3>
-        Results <span *ngIf="challenges.length > 0">({{ searchResultsCount }})</span>
+      <h3 *ngIf="searchResultsCount !== undefined">
+        Results <span>({{ searchResultsCount }})</span>
```
tschaffter, line 154: The section title "Results" must always been displayed. This is so that the user can understand how the page is structured. The suffix " (N)" can be added after successfully fetching data.

Could you also confirm that the solution you come up with does not reintroduce the flickering that led to the regression issue being fixed here to reappear?

tschaffter (changes requested):
Th

libs/openchallenges/challenge-search/src/lib/challenge-search.component.html
rrchai: > Could you also confirm that the solution you come up with does not reintroduce the flickering that led to the creation of the present regression issue?

Yes, I have also added a frame-by-frame recording to demonstrate that the flickering issue has not been reintroduced.

tschaffter:
@rrchai The flickering is back for me. This happens when reaching the page directly from another website or reloading the page but not when navigating between pages. Can you reproduce this flickering.

Recording 2024-03-08 at 10 13 53

tschaffter (changes requested):
See comment above about the flickering

rrchai:
Interesting, the flickering is not reproduced on my end. 

Recording 2024-03-08 at 13 37 30

I wonder if this flickering (from reloading the page) occurred before this PR was implemented on your end?

tschaffter:
You are right, the flickering is already present on the production site (https://openchallenges.io/challenge). So we never got rid of it...

tschaffter (approved):
I'm approving this PR. There is still a flickering on that page that we may never have gotten rid off. A separate ticket has been created.

Issue #2546, opened 2024-03-04
[Story] Develop the ETL pipeline for the EDAM ontology


tschaffter:
### What product(s) is this story for?

OpenChallenges

### Description

The goal of this story is to update the existing `openchallenges-edam-etl` to effectively load the EDAM ontology into the OC Challenge Service DB.

### Acceptance criteria

- The command `nx serve-detach openchallenges-edam-etl` loads the data into the challenge service DB.
- The file `edam_concept.csv` is removed from the challenge service codebase.

### Tasks

- [x] #2547
- [x] #2548

### Have you linked this story to a GitHub Project?

- [X] I have linked this story to a GitHub Project and set its metadata.

tschaffter:
Added to Sprint 24.4

tschaffter:
Added to Backlog

Issue #2548, opened 2024-03-04
Load the transformat EDAM data into the OC Challenge Service DB


tschaffter:

tschaffter:
Added to Sprint 24.4

tschaffter:
Moved to Backlog

mdsage1:
*** Update: 05/13/2024
Challenges: N/A 
Remaining Tasks: Implement the load aspect of the ETL process so that the generated dataset is available/accessible in the MariaDB

mdsage1:
**Update: 05/15/2024**
Challenges:

@tschaffter I have written code to connect to MariaDB using python. The OC_DB_URL within the .env file, jdbc:mysql://openchallenges-mariadb:3306/edam_etl, isn't used according to the documentation I have located. [Resource1](https://mariadb.com/resources/blog/how-to-connect-python-programs-to-mariadb/) and [Resource2](https://mariadb.com/docs/server/connect/programming-languages/python/connect/#Code_Example:_Connect) for connecting to MariaDB.

I've received this error when using jdbc:mysql://openchallenges-mariadb:3306/edam_etl as the Host:
`Error connecting to MariaDB Platform: Plugin jdbc:mysql could not be loaded: /usr/lib/x86_64-linux-gnu/libmariadb3/plugin/jdbc:mysql.so: cannot open shared object file: No such file or directory`
Warning: command "poetry run python src/main.py" exited with non-zero status code

I get this error when I change the host to openchallenges-mariadb and don't use the OC_DB_URL in the .env file:
```
Error connecting to MariaDB Platform: Access denied for user 'edam_etl'@'%' to database 'challenge_service'
Warning: command "poetry run python src/main.py" exited with non-zero status code
```

I'm wondering if the variables are being assigned incorrectly.

tschaffter:
I believe that we have solve the issue since your last message. Feel free to get rid of the config variable `OC_DB_URL`. We use it the OC microservice because the DB client we use in Java accept this URL as a parameter, which not be the case of the DB client for Python you use.

mdsage1:
[...]

Issue #2524, opened 2024-02-22
[Story] Scaffold EDAM ETL project


tschaffter:
### What product(s) is this story for?

OpenChallenges

### As a user, I want

As a Developer, I want a programmatic way to ETL the EDAM ontology so that this process is reproducible and EDAM data no longer need to be tracked in this repository.

### Description

Copy-paste and adapt the project `iatlas-data`

### Acceptance criteria

- The new project should generate a Docker image that includes an ETL Python script.
- The ETL Python script does be empty for now; the development of the ETL script will be tracked as a separated task.
- The ETL container should be included to the OC stack.

### Have you linked this story to a GitHub Project?

- [X] I have linked this story to a GitHub Project and set its metadata.

PR #2564, opened 2024-03-13
feat(openchallenges): add EDAM Extract and Transform Processes

### LINKED ISSUE #2547: [Task] Extract and transform the EDAM ontology data

OpenChallenges

### As a user, I want

_No response_

### Description

Update the script `src/main.py` of the `openchallenges-edam-etl` to perform the following tasks:

### Acceptance criteria

Running the following commands download, transform and print information about the EDAM data:

- `nx serve openchallenges-edam-etl`
- `nx serve-detach openchallenges-edam-etl`

### Tasks

- Download a specified version of the EDAM ontology from https://github.com/Sage-Bionetworks/edamontology
- Transform the raw data into a Pandas dataframe that match the content of [this file](https://github.com/Sage-Bionetworks/sage-monorepo/blob/2958c7453a8bee178317e8d6811400b8e2521627/apps/openchallenges/challenge-service/src/main/resources/db/edam_concept.csv#L4)
  - Start id values from 1 to mimic the behavior of SQL `AUTO_INCREMENT`.
- Print info and statistic about the data to the stdout
  - Version of EDAM processed
  - Number of concepts transformed (overall, operation, data, etc.)

### Anything else?

See #2524 and its PR to get familiar with the environment of the project `openchallenges-edam-etl`.

### LINKED ISSUE #2563: [Task] Start EDAM concept ID from 1 instead of 0

OpenChallenges

### As a user, I want

_No response_

### Description

This is to align with all the other IDs that are generated with AUTO INCREMENT.

References to existing EDAM IDs need to be updated too in `input_data_type.csv` and `challenges.csv` (`operation_id` property).

Depends on #2557

### Acceptance criteria

_No response_

### Tasks

_No response_

### Anything else?

_No response_

mdsage1:
## Description
EDAM ETL processes need to be developed to incorporate ETAM ontology in the Maria DB linking the ontology to existing data. This PR will address the extract and transform portion.

## Related Issue
Contribute to #2524
Contribute to #2548

Fixes #2547 
Fixes #2563

## Changelog

- Add 

1. Download a specified version of the EDAM ontology from https://github.com/Sage-Bionetworks/edamontology
2. Transform the raw data into a Pandas dataframe that match the content of [this file](https://github.com/Sage-Bionetworks/sage-monorepo/blob/2958c7453a8bee178317e8d6811400b8e2521627/apps/openchallenges/challenge-service/src/main/resources/db/edam_concept.csv#L4)
3. Start id values from 1 to mimic the behavior of SQL AUTO_INCREMENT.
4. Print info and statistic about the data to the stdout
5. Version of EDAM processed
6. Number of concepts transformed (overall, operation, data, etc.)

## Preview
image


mdsage1:
@tschaffter Quality Gate doesn't seem to performing checks for this PR.

mdsage1:
@tschaffter This is ready for review. Version is now an environment variable and the Description includes a preview.

apps/openchallenges/edam-etl/Dockerfile
```diff
@@ -11,10 +11,11 @@ RUN pip install poetry \
...
+
 WORKDIR /
 COPY docker-entrypoint.sh ./
 RUN chmod +x docker-entrypoint.sh
 
 ENTRYPOINT ["/docker-entrypoint.sh"]
 
-CMD python "${APP_DIR}/src/main.py"
\ No newline at end of file
+CMD python "${APP_DIR}/src/main.py"
```
vpchung, line 21: @tschaffter I missed this the first time around, but should this be in exec form?  Or does it not matter?

From the [docs](https://docs.docker.com/reference/dockerfile/#cmd):
> If CMD is used to provide default arguments for the ENTRYPOINT instruction, both the CMD and ENTRYPOINT instructions should be specified in the [exec form](https://docs.docker.com/reference/dockerfile/#exec-form).

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,68 @@
-"""Extract, transform and load EDAM concepts"""
-
+import pandas as pd
+import requests
 from os import getenv
 
 # Get config from the environment variables
 oc_db_url = getenv("OC_DB_URL")
```
vpchung, line 6: nit: [PEP 8](https://peps.python.org/pep-0008/#constants) prefers constant variables to be in all caps, e.g.

```suggestion
OC_DB_URL = getenv("OC_DB_URL")
```

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,68 @@
...
+version = getenv("OC_DB_VERSION")
```
vpchung, line 7: nit: same comment as above re: constant variable naming

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,68 @@
...
+
+# Transform the CSV to a DataFrame with indices starting from 0
+def transform_to_dataframe(version):
+    print("Processing the EDAM concepts...")
+    try:
+        df = pd.read_csv(f"EDAM_{version}.csv")
+        df["id"] = df.index
+        # Select and rename specific columns
+        df = df[["id", "Class ID", "Preferred Label"]]
+        df = df.rename(
```
vpchung, line 33: I'm a big fan of method chaining (which comes with the benefits of streamlining the data processing as well as being more memory-efficient), so we could alternatively do the following instead:

```python
df = (
   pd.read_csv(f"EDAM_{version}.csv", usecols=['Class ID', 'Preferred Label'])
   .rename(columns={"Class ID": "class_id", "Preferred Label": "preferred_label"})
   .assign(id=df.index + 1)
)
```

Here, I added `usecols` to the `read_csv()` method, so that we subset which columns we need rather than load the entire file into the disk space first.  I additionally added `assign()` to the chain, which will add a new column to the data frame called `id` (therefore, replacing line 30).  I also added 1 to `df.index` since I believe "id" needs to start at 1 rather than 0.  @tschaffter to confirm.

We could alternatively do the following as well:

```python
df2 = (
   pd.read_csv("Downloads/EDAM.csv", usecols=["Class ID", "Preferred Label"])
   .reset_index()
   .rename(columns={"index": "id", "Class ID": "class_id", "Preferred Label": "preferred_label"})
   .assign(id=df.index + 1)
)
```

The difference in these two approaches is that `id` would be listed first in `df2`, rather than listed last like in `df`:

```
>>> df.columns
Index(['class_id', 'preferred_label', 'id'], dtype='object')
>>> df2.columns
Index(['id', 'class_id', 'preferred_label'], dtype='object')
```

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,68 @@
...
+        print("Statistics:")
+        # Set the display options to show only 2 decimal places
+        pd.set_option("display.float_format", "{:.0f}".format)
+        print(df.describe())
+    else:
+        print("No data available.")
+
+
+def main():
+    url = "https://github.com/edamontology/edamontology/raw/main/releases/EDAM_1.25.csv"
```
vpchung, line 58: Perhaps interpolate `VERSION` into this url as well?

```suggestion
    url = f"https://github.com/edamontology/edamontology/raw/main/releases/EDAM_{VERSION}.csv"
```

That way, we won't need to manually update this when the time comes!

vpchung:
@mdsage1 thanks for working on this!!  You didn't ask me to, but I added some comments to the PR. Feel free to use them or ignore 😄

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,68 @@
-"""Extract, transform and load EDAM concepts"""
-
+import pandas as pd
```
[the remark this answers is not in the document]
mdsage1: Added back in docstring that was removed accidentally

apps/openchallenges/edam-etl/src/main.py
mdsage1: updated constant variable naming

apps/openchallenges/edam-etl/src/main.py
mdsage1: interpolated version into the url

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,68 @@
...
+    print_info_statistics(version, df)
+
+
+if __name__ == "__main__":
+    main()
 
-# TODO Push the EDAM concept to the OC challenge service DB
-print("[TODO] Push the EDAM concept to the OC challenge service DB")
+# TODO: Push the EDAM concepts to the OC challenge service DB
+print("\n[TODO] Push the EDAM concepts to the OC challenge service DB")
```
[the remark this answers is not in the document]
mdsage1: Removed the ToDo

apps/openchallenges/edam-etl/src/main.py
mdsage1: updated variable name

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,68 @@
...
 # Get config from the environment variables
 oc_db_url = getenv("OC_DB_URL")
 print(f"OC DB URL: {oc_db_url}")
 
-# TODO Download EDAM concepts from GitHub or S3 bucket (CSV file)
-print("[TODO] Download EDAM concepts from GitHub or S3 bucket (CSV file)")
 
-# TODO Process the EDAM concepts
-print("[TODO] Process the EDAM concepts")
+# Download EDAM concepts from GitHub or S3 bucket (CSV file)
```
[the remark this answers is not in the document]
mdsage1: comment is now a docstring

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,68 @@
...
+        response = requests.get(url)
+        response.raise_for_status()  # Raise an exception for bad response
+        with open(f"EDAM_{version}.csv", "wb") as f:
+            f.write(response.content)
+        print("EDAM concepts downloaded successfully.")
+    except requests.RequestException as e:
+        print(f"Error downloading EDAM concepts: {e}")
+
+
+# Transform the CSV to a DataFrame with indices starting from 0
```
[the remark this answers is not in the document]
mdsage1: comment is now a docstring

apps/openchallenges/edam-etl/src/main.py
mdsage1: used the first method and made improvements to ensure code executed as it resulted in an erorr related to assigning the id `Error processing EDAM concepts: cannot access local variable 'df' where it is not associated with a value`

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,68 @@
...
+        )
+        print("EDAM concepts processed successfully.")
+        return df
+    except FileNotFoundError:
+        print(f"File EDAM_{version}.csv not found.")
+    except Exception as e:
+        print(f"Error processing EDAM concepts: {e}")
+
+
+# Gather data about the EDAM ontology
```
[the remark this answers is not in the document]
mdsage1: changed comment to a docstring

tschaffter:
> Fixes #2546

This PR is part of #2546, so this PR should not be configured to close this ticket.

tschaffter (commented):
@mdsage1 Why do you show the std, mean and other metrics for the id column?

tschaffter (changes requested):
The config parameters should be validated before using them, otherwise the following behavior will occur:

```
$ nx serve-detach openchallenges-edam-etl

> nx run openchallenges-edam-etl:serve-detach

 Container openchallenges-mariadb  Recreate
 Container openchallenges-mariadb  Recreated
 Container openchallenges-edam-etl  Recreate
 Container openchallenges-edam-etl  Recreated
 Container openchallenges-mariadb  Starting
 Container openchallenges-mariadb  Started
 Container openchallenges-mariadb  Waiting
 Container openchallenges-mariadb  Healthy
 Container openchallenges-edam-etl  Starting
 Container openchallenges-edam-etl  Started

 ————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 >  NX   Successfully ran target serve-detach for project openchallenges-edam-etl (33s)
 
   View logs and investigate cache misses at https://cloud.nx.app/runs/Ivfcx2Zd35

vscode@dee30b82cf44:/workspaces/sage-monorepo$ docker logs openchallenges-edam-etl
EDAM Version: None
OC DB URL: jdbc:mysql://openchallenges-mariadb:3306/challenge_service
Downloading the EDAM concepts from GitHub (CSV file)...
Error downloading EDAM concepts: 404 Client Error: Not Found for url: https://github.com/edamontology/edamontology/raw/main/releases/EDAM_None.csv
Processing the EDAM concepts...
File EDAM_None.csv not found.
No data available.
```

apps/openchallenges/edam-etl/Dockerfile
tschaffter: This improvement is also suggested by Hadolint, which will be added to our development workflow soon. For now it should still be addressed.

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,75 @@
...
+def download_edam_csv(url: str, version: str) -> Optional[bool]:
+    """Download EDAM concepts from GitHub or S3 bucket (CSV file)"""
+    print("Downloading the EDAM concepts from GitHub (CSV file)...")
+    try:
+        response = requests.get(url)
+        response.raise_for_status()  # Raise an exception for bad response
+        with open(f"EDAM_{version}.csv", "wb") as f:
+            f.write(response.content)
+        print("EDAM concepts downloaded successfully.")
+        return True
```
vpchung, line 26: Is a return statement needed for this function?  I don't see the return value being used.

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,75 @@
 """Extract, transform and load EDAM concepts"""
 
+import pandas as pd
+import requests
 from os import getenv
+from typing import Optional
+from typing import NoReturn
```
vpchung, line 7: nit: can be joined into a single statement:

```python
from typing import Optional, NoReturn
```

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,75 @@
...
+        with open(f"EDAM_{version}.csv", "wb") as f:
+            f.write(response.content)
+        print("EDAM concepts downloaded successfully.")
+        return True
+    except requests.RequestException as e:
+        print(f"Error downloading EDAM concepts: {e}")
+        return None
+
+
+def transform_to_dataframe(version: str) -> pd.DataFrame:
```
vpchung, line 32:
```suggestion
def transform_to_dataframe(version: str) -> pd.DataFrame | None:
```

Since `None` will be returned if there is an exception.

Note: using the pipe is supported by Python 3.10+.  If we want to be backwards-compatible, then use `Optional[pd.DataFrame]`

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,75 @@
...
+    """Transform the CSV to a DataFrame with indices starting from 1"""
+    print("Processing the EDAM concepts...")
+    try:
+        df = (
+            pd.read_csv(f"EDAM_{version}.csv", usecols=["Class ID", "Preferred Label"])
+            .rename(
+                columns={"Class ID": "class_id", "Preferred Label": "preferred_label"}
+            )
+            .assign(id=lambda x: x.reset_index(drop=True).index + 1)
```
vpchung, line 41: Interesting that using `.assign(id=df.index + 1)` didn't work, but I'm glad you found a solution!

vpchung: Oh, nevermind - I get the reason why 🤦‍♀️😅

apps/openchallenges/edam-etl/src/main.py
vpchung: Actually, maybe remove `NoReturn`?  Since it's not being used.

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,75 @@
...
+import pandas as pd
+import requests
 from os import getenv
+from typing import Optional
+from typing import NoReturn
 
-# Get config from the environment variables
-oc_db_url = getenv("OC_DB_URL")
-print(f"OC DB URL: {oc_db_url}")
+"""Get config from the environment variables"""
```
vpchung, line 9: nit: this doesn't need to be a docstring - it can remain as a comment 😁

apps/openchallenges/edam-etl/src/main.py
mdsage1: thanks I forgot I had it there then didn't need it.

apps/openchallenges/edam-etl/src/main.py
mdsage1: changed to comment

apps/openchallenges/edam-etl/src/main.py
mdsage1: Added `| None`

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,75 @@
...
+        print("No data available.")
+
+
+def main() -> None:
+    """Main function to excute preceeding functions"""
+    url: str = (
+        f"https://github.com/edamontology/edamontology/raw/main/releases/EDAM_{VERSION}.csv"
+    )
+    download_edam_csv(url, VERSION)
+    df: pd.DataFrame = transform_to_dataframe(VERSION)
```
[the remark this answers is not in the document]
mdsage1: I think this was added at the request of Sonar's quality check. I'll take it out and see if the issue arises again.

mdsage1:
closed in error

mdsage1:
> @mdsage1 ~Why do you show the std, mean and other metrics for the id column?~
> 
> Update: here are the information that the script should print
> 
> * Version of EDAM processed
> * Number of concepts that will be added to the table
>   
>   * Total number of concepts
>   * Number of concepts for the following category
>     
>     * Data concepts
>     * Operation concepts
>     * Format concepts
>     * Operation concepts
>     * Other concepts


@tschaffter Where would I find this data in the edam csv downloaded from GitHub. I do not see a column that indicates a concept name/category only a preferred label that may include some of the concepts that you've listed

apps/openchallenges/edam-etl/src/main.py
mdsage1: The return value is used to return all of the content that is associated w/the function. Without it only `EDAM concepts downloaded successfully.` will be printed.

tschaffter:
> Where would I find this data in the edam csv downloaded from GitHub.

You could get this information from the column `class_id` and a regex.

mdsage1:
> > Where would I find this data in the edam csv downloaded from GitHub.
> 
> You could get this information from the column `class_id` and a regex.

@tschaffter Please see an example of the class_id column below: 
image
This column is a link that only differs in the sequence of numbers at the end of the link for each entry.

mdsage1:
@tschaffter I have updated the concept counts using the preferred_label column and removed the statistics like mean etc.

tschaffter:
> You could get this information from the column class_id and a regex.

See the suggestion I made above.

vpchung:
> You could get this information from the column class_id and a regex.

@mdsage1 alternatively, you can also do a `replace()` to remove the substring you don't need from `class_id`, if you're not comfortable with regex.

tschaffter:
Prefer exact match instead of `contains`: `contains` would not work if the ontology were to have the concept `Data` and `DataFormat`, e.g.

vpchung:
> if the ontology were to have the concept Data and DataFormat, for example.

Good point.  Just shooting my shot here, but this can be overcome by using `data_` (assuming they use "dataformat_").  Also, you can use regex with [`contains()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.str.contains.html).

mdsage1:
@tschaffter I've updated the concept counts to use the class_id column and regex. The case has been ignored to avoid any future issues. I didn't use contains but used search() function from the regex module. I have prevented future issues with data, and any other concept name, listing as a match when there is an additional word following the word of interest by adding the underscore to the regex as @vpchung suggested.

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,100 @@
...
+        format_pattern = re.compile(r"format_", re.IGNORECASE)
+
+        # Initialize counts
+        data_count = 0
+        operation_count = 0
+        format_count = 0
+        other_count = 0
+
+        # Loop through the class_id column and count occurrences
+        for value in df["class_id"]:
```
vpchung, line 68: A loop works!  It's very clear and straightforward on what the code is trying to do.

That being said, I think it could still be handy to take advantage of the functional programming that pandas offer, especially since `df` is a data frame.  Here's how I would use `contains()` in lieu of looping through the rows:

```python
>>> # Approach 1
>>> df["Class ID"].str.contains(r"data_\d+$", regex=True).value_counts()[True]
1493
>>>
>>> # Approach 2
>>> len(df[df["Class ID"].str.contains(r"data_\d+$", regex=True)])
1493
```

vpchung: Here are the performance differences of the three approaches:

```python
>>> # For loop
>>> timeit(
... 'for value in df["Class ID"]: data_count += 1 if data_pattern.search(value) else 0', 
... setup='import pandas as pd; import re; df = pd.read_csv("~/Downloads/EDAM.csv", usecols=["Class ID", "Preferred Label"]).fillna(""); data_pattern = re.compile(r"data_", re.IGNORECASE); data_count = 0',
... number=1000)
1.7666597969946451
>>>
>>> # using contains with value_counts
>>> timeit(
... 'df["Class ID"].str.contains(r"data_\d+$", regex=True).value_counts()[True]',
... setup='import pandas as pd; df = pd.read_csv("~/Downloads/EDAM.csv", usecols=["Class ID", "Preferred Label"]).fillna("")',
... number=1000)
1.2751378359971568
>>>
>>> # using contains with len
>>> timeit(
... 'len(df[df["Class ID"].str.contains(r"data_\d+$", regex=True)])',
... setup='import pandas as pd; df = pd.read_csv("~/Downloads/EDAM.csv", usecols=["Class ID", "Preferred Label"]).fillna("")',
... number=1000)
1.2614918049948756
```

So, not a _huge_ improvement on time, but worth noting.

vpchung: Ultimately, I say go with what makes the most sense to you!

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,100 @@
...
+    try:
+        response = requests.get(url)
+        response.raise_for_status()  # Raise an exception for bad response
+        with open(f"EDAM_{version}.csv", "wb") as f:
+            f.write(response.content)
+        print("EDAM concepts downloaded successfully.")
+        return True
+    except requests.RequestException as e:
+        print(f"Error downloading EDAM concepts: {e}")
+        return None
```
vpchung, line 27: This doesn't matter, but I'm just curious: why return `None` instead of `False`?

vpchung (approved):
I have one final suggestion, but otherwise, the script looks good on my end!

apps/openchallenges/edam-etl/src/main.py
mdsage1: Honestly, I was just trying my best to incorporate both of your feedback w/o causing any additional collisions based on how I implemented the logic.

apps/openchallenges/edam-etl/src/main.py
mdsage1: False indicates a failure, i.e. the download failed because the operation didn't do what it was supposed to do properly and there is a boolean checking to see if the operation fails, while None just says that there is no boolean since there's no meaningful output. Sonar suggested that I add an exception many pushes to this pr ago so I just added that since IMO the use of an exception wasn't really necessary since we know that either you're getting a download when executing the function or not. My original function had no exceptions until the Sonar Issue was flagged.

mdsage1 (commented):
tested the link and updated it

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,100 @@
...
+
+        # Define regex patterns for each concept
+        data_pattern = re.compile(r"data_", re.IGNORECASE)
+        operation_pattern = re.compile(r"operation_", re.IGNORECASE)
+        format_pattern = re.compile(r"format_", re.IGNORECASE)
+
+        # Initialize counts
+        data_count = 0
+        operation_count = 0
+        format_count = 0
```
tschaffter, line 64: You missed concept types. See https://edamontology.org/page#Documentation

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,100 @@
...
+        # Loop through the class_id column and count occurrences
+        for value in df["class_id"]:
+            if data_pattern.search(value):
+                data_count += 1
+            elif operation_pattern.search(value):
+                operation_count += 1
+            elif format_pattern.search(value):
+                format_count += 1
+            else:
+                other_count += 1
```
tschaffter, line 76: Aside the missing concept, this implementation is OK.

However, "for" loops are usually inefficient. Could you please try the following implementation proposed by ChatGPT?

> Certainly! To implement a solution without an explicit for loop, we can leverage pandas' vectorized string operations. These operations allow us to apply regular expressions over the entire column efficiently. However, note that this approach may be more suitable for certain types of patterns and data structures. In this case, we'll use the str.contains() method provided by pandas, which supports regex and can be used to create boolean masks for filtering or counting.

```python
import pandas as pd

# Assuming df is your DataFrame and 'class_id' is the column of interest

# Create regex patterns for each concept
data_pattern = r"data_"
operation_pattern = r"operation_"
format_pattern = r"format_"

# Use pandas' vectorized string operations to count occurrences
data_count = df["class_id"].str.contains(data_pattern, case=False, na=False).sum()
operation_count = df["class_id"].str.contains(operation_pattern, case=False, na=False).sum()
format_count = df["class_id"].str.contains(format_pattern, case=False, na=False).sum()

# Calculate 'other' count by subtracting the specific counts from the total
other_count = len(df) - (data_count + operation_count + format_count)

# Example of how to print the counts
print(f"data: {data_count}")
print(f"operation: {operation_count}")
print(f"format: {format_count}")
print(f"other: {other_count}")
```

tschaffter (changes requested):

apps/openchallenges/edam-etl/src/main.py
tschaffter: The solution introduced by ChatGPT relies on `contains`. There is a way to use `contains` without bumping into the issue I described previously. Can you see how to achieve that?

tschaffter: Note how this vectorized implementation is easier to read.

apps/openchallenges/edam-etl/src/main.py
mdsage1: I was relying on your instructions given in this pr. 
image
I will add the topic and identifier concepts.

apps/openchallenges/edam-etl/src/main.py
mdsage1: The cconcepts were used in accordance to this comment:

image
I have added the Topic and Identifier concepts.

I avoided contains() due to this response to @vpchung comment:
image
I have updated the function to include contains()

apps/openchallenges/edam-etl/src/main.py
tschaffter: My bad. Thanks!

apps/openchallenges/edam-etl/src/main.py
vpchung: `contains()` makes a comeback! :-)

mdsage1: The contains() will not bump into an issue if the regex is appropriately written the way @vpchung suggested to include a hyphen after the word.

vpchung: Re: the comment about exact matches, this can be done with `contains()` by flagging `regex=True` which will allow searching with a pattern. I showed an usage example in my latest code review, but feel free to ping me for some guidance!

tschaffter: @mdsage1 Do you mean the underscore character? A constraint should be placed on both sides of the string, not just at the end. For example (the forward slash may need to be escaped)

```
data_pattern = r"/data_"
```

mdsage1: Sorry about that I thought I typed underscore.  I meant underscore not hyphen. I've added the regex=True (@vpchung) great catch because I thought I typed that in! I added a forward slash before the concept name @tschaffter

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,117 @@
...
+        # Create regex patterns for each concept
+        data_pattern = r"/data_"
+        operation_pattern = r"/operation_"
+        format_pattern = r"/format_"
+        topic_pattern = r"/topic_"
+        identifier_pattern = r"/identifier_"
+
+        # Use pandas' vectorized string operations to count occurrences
+        data_count = (
+            df["class_id"]
```
vpchung, line 65: Sorry, one more final thought:

Since this code is reused multiple times, maybe extract it into a function?  That way, you would only need to update the logic once instead of 5 times.

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,117 @@
...
+
+
+def print_info_statistics(df: pd.DataFrame) -> None:
+    """Gather data about the EDAM ontology"""
+    if df is not None:
+        print(f"Number of Concepts Transformed: {len(df)}")
+        print(f"Column names: {df.columns.tolist()}")
+
+        # Create regex patterns for each concept
+        data_pattern = r"/data_"
```
vpchung, line 57: For more specificity, you can also add that it should end with digits? e.g.

```suggestion
        data_pattern = r"/data_\d+$"
```

apps/openchallenges/edam-etl/src/main.py
mdsage1: extracted this into a function.

apps/openchallenges/edam-etl/src/main.py
mdsage1: added `\d+$`

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,107 @@
...
+            .rename(columns={"index": "id"})
+        )
+        print("EDAM concepts processed successfully.")
+        return df
+    except FileNotFoundError as e:
+        print(f"Error processing EDAM concepts: {e}")
+        return None
+
+
+def count_occurrences(identifier_pattern: str, df) -> int:
```
vpchung, line 49:
```suggestion
def count_occurrences(identifier_pattern: str, df: pd.DataFrame) -> np.int64:
```

^^ IIRC.  May need to double-check whether it is a numpy type being returned.

apps/openchallenges/edam-etl/src/main.py
```diff
@@ -1,16 +1,107 @@
...
+    """Count the number of pattern occurrences"""
+    return (
+        df["class_id"]
```
vpchung, line 52: Maybe have the column name be a function parameter too?

Or, instead of passing the entire dataframe, pass only the series?  So:

```python
def count_occurrences(identifier_pattern: str, column: pd.Series) -> int:
   return (
      column
      ...
   )

...

data_count = count_occurrences(data_pattern, df["class_id"])
```

apps/openchallenges/edam-etl/src/main.py
mdsage1: Numpy int is being returned and made changes.

apps/openchallenges/edam-etl/src/main.py
mdsage1: Changed function to pass to column

tschaffter (approved):

apps/openchallenges/edam-etl/src/main.py
vpchung: Cool cool. May need to add `import numpy as np` in this case!

tschaffter: Nice catch! I need to submit and patch for this project in #2594 and will fix this issue at the same time.

PR #2612, opened 2024-04-01
fix(openchallenges): add functionality to retrieve label for challenge property from mapping file

### LINKED ISSUE #2580: [Bug] Fix MLCube casing typo

OpenChallenges

### Current behavior

The official casing is MLCube but is displayed as "Mlcube" on the challenge page. The casing is different on the challenge page:

See the challenge page for one of those: https://openchallenges.io/challenge?submissionTypes=mlcube

I fixed the casing last week for the filter. Ideally, we should define the mapping from the constant to human-friendly name once and then use it for the filter and the challenge page.

> Submission Type	Prediction file, Mlcube

### Expected behavior

_No response_

### Anything else?

_No response_

### Commit ID

_No response_

### Code of Conduct

rrchai:
- fixes #2580 

## Changelogs
- Map the label for a challenge property based on a mapping file. To minimize changes, use the existing filter file, which consists of the value and label for each filterable challenge property, as the mapping file.
- Add a generic function to retrieve the label from the filter object array, allowing it to be called in other components, like `challenge-overview`

### DIFF
--- a/libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.html
+++ b/libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.html
@@ -26,7 +26,7 @@
         </tr>
         <tr>
           <td class="text-right">Status</td>
-          <td>{{ prettify(challenge.status) }}</td>
+          <td>{{ getStatusLabel(challenge.status) }}</td>
         </tr>
         <tr>
           <td class="text-right">Platform</td>
@@ -58,7 +58,7 @@
               nowrap
               *ngFor="let submissionType of challenge.submissionTypes; let isLast = last"
             >
-              {{ prettify(submissionType) }}{{ isLast ? '' : ', ' }}</span
+              {{ getSubmissionTypeLabel(submissionType) }}{{ isLast ? '' : ', ' }}</span
             >
           </td>
           <ng-template #na_subs>
@@ -69,7 +69,7 @@
           <td class="text-right">Incentive(s)</td>
           <td *ngIf="challenge.incentives && challenge.incentives.length > 0; else na_prize">
             <span nowrap *ngFor="let incentive of challenge.incentives; let isLast = last">
-              {{ prettify(incentive) }}{{ isLast ? '' : ', ' }}</span
+              {{ getIncentiveLabel(incentive) }}{{ isLast ? '' : ', ' }}</span
             >
           </td>
           <ng-template #na_prize>
--- a/libs/openchallenges/challenge-search/src/index.ts
+++ b/libs/openchallenges/challenge-search/src/index.ts
@@ -1,1 +1,3 @@
 export * from './lib/challenge-search.routes';
+export * from './lib/challenge-search-filters';
+export * from './lib/get-label-by-filter-value';
--- a/libs/openchallenges/challenge-search/src/lib/get-label-by-filter-value.ts
+++ b/libs/openchallenges/challenge-search/src/lib/get-label-by-filter-value.ts
@@ -1,0 +1,13 @@
+import { Filter } from '@sagebionetworks/openchallenges/ui';
+
+/**
+ * A generic function used to retrieve the label for values of challenge properties
+ */
+
+export function getLabelByFilterValue(
+  filter: Filter[],
+  value: any
+): string | undefined {
+  const filterItem = filter.find((item) => item.value === value);
+  return filterItem ? filterItem.label : undefined;
+}
--- a/libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.ts
+++ b/libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.ts
@@ -1,10 +1,22 @@
 import { CommonModule } from '@angular/common';
 import { Component, Input } from '@angular/core';
-import { Challenge } from '@sagebionetworks/openchallenges/api-client-angular';
+import {
+  Challenge,
+  ChallengeIncentive,
+  ChallengeStatus,
+  ChallengeSubmissionType,
+} from '@sagebionetworks/openchallenges/api-client-angular';
 import {
   MOCK_ORGANIZATION_CARDS,
   OrganizationCard,
 } from '@sagebionetworks/openchallenges/ui';
+import {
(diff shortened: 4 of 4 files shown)


rrchai:
I realize the challenge card has incentive information. To avoid calling the function hundreds of times to get incentive label for each challenge card, I will change to create a separate mapping file for label-value.

rrchai:
When I was trying to update the incentive in the challenge card component, all pages displayed blank with an error indicating that loading the footer failed. Here is the error message.

```
statistics-viewer.component.ts:13 Uncaught TypeError: Cannot read properties of undefined (reading 'FooterComponent')
    at Module.FooterComponent (statistics-viewer.component.ts:13:72)
    at 72716 (challenge-search.component.ts:83:5)
    at __webpack_require__ (bootstrap:19:1)
    at 17844 (challenge-search.component.ts:87:34)
    at __webpack_require__ (bootstrap:19:1)
    at 15392 (variables.ts:9:2)
    at __webpack_require__ (bootstrap:19:1)
    at 71604 (mock-avatars.ts:14:2)
    at __webpack_require__ (bootstrap:19:1)
    at 65488 (statistics-viewer.component.ts:13:72)
```

I noticed this issue when I shared the mapping file in the `challenge-search` library via `index.ts`. However, the error disappears if I share it in the `ui` library. I'm not sure if it's related to a recurring module loading issue.

libs/openchallenges/ui/src/index.ts
```diff
@@ -4,6 +4,7 @@ export * from './lib/avatar/mock-avatars';
 export * from './lib/challenge-card/challenge-card.component';
 export * from './lib/challenge-card/mock-challenges';
 export * from './lib/challenge-card/mock-platforms';
+export * from './lib/challenge-card/challenge-property-labels';
```
tschaffter, line 7: Placing the labels in the folder of the `challenge-card` component seems arbitrary. Where could these Records be moved to?

libs/openchallenges/challenge/src/lib/challenge-overview/challenge-overview.component.html
```diff
@@ -26,7 +26,7 @@ <h3 class="section-title">Challenge Details</h3>
         </tr>
         <tr>
           <td class="text-right">Status</td>
-          <td>{{ prettify(challenge.status) }}</td>
+          <td>{{ challenge.status | titlecase }}</td>
```
tschaffter, line 29: Does this pipe make assumption about the value of the status? I'm asking because this very PR is about fixing an assumption about the casing of challenge input data format that was no longer valid with "MLCube". :-)

tschaffter: I get what the pipe does and it's OK given the existing set of status.

libs/openchallenges/ui/src/index.ts
rrchai: yea, I was trying to put it in the `challenge` and `challenge-search` libs, but I got the blank pages. As long as  I call the records is not stored in `ui` lib and the record is called in `challenge-card` component, it will raise error with blank pages. 

 I'm not sure if it's due to some sort of circular dependencies, but I was not able to identify the issue yesterday.

tschaffter: What about using a `constants.ts` file like we do in other projects? E.g. in `./lib/constants.ts` (usually data only) and/or `./util.ts` (reusable utility functions)?

rrchai:
>I noticed this issue when I shared the mapping file in the challenge-search library via index.ts. However, the error disappears if I share it in the ui library. I'm not sure if it's related to a recurring module loading issue.

### Potential circular dependency, if creating a new mapping file stored in `challenge-search`/`challenge`:

- `challenge-search`/`challenge` depends on `ui` (which includes `challenge-card`).
- `challenge-card` depends on `challenge-search`/`challenge` for records.

The new mapping file cannot be in the `challenge`/`challenge-search` lib to avoid the error.

### Potential circular dependency, If using a function to retrieve the labels from an existing mapping file in `challenge-search`:
- the reusable function depends on `challenge-search` for mappings.
- `challenge-search` depends on `ui` (which includes `challenge-card`).
- `challenge-card` depends on  the reusable function for mappings.

This approach is not doable, as the reusable function needs to be isolated from `challenge-search` to avoid the error, unless creating/moving a mapping file. Or directly import the mapping file in `challenge-search`, see [this comment](https://github.com/nrwl/nx/issues/9894#issuecomment-1174797217).

libs/openchallenges/util/src/lib/pipe/challenge-property-label.pipe.ts
tschaffter, line 1: The pipe approach is an elegant solution for converting types from `@sagebionetworks/openchallenges/api-client-angular` to human-readable English.

tschaffter, line 1: The current implementation makes the library `libs/openchallenges/util` unnecessarily depending on `../../../../challenge-search/src/lib/challenge-search-filters`. The util library is aimed to be consumed by other library and should remain independent as much as possible.

With the current implementation, any libraries that import the util library will also import the `challenge-search` project, which is an entire page of OC.

How would you solve this issue?

tschaffter (changes requested):

libs/openchallenges/util/src/lib/pipe/challenge-property-label.pipe.ts
tschaffter: Your proposal makes sense to me, i.e. moving the localization data and related functions to `util`, then importing them in the projects that need them. In the future, I can envision creating a `localization` library on its own but the first implementation can be included in `util`.

I recommend moving the minimum localization data to the `util` library, i.e. not the entire filters that include information not related to localization like whether the filter is enabled or not.

tschaffter (approved):
This fix works as expected, thanks! I like how developer-friendly using Angular pipes to convert constant to human-readable text is.

There are two improvements I can see for the future:
- use the constant from the API client as keys in a dictionary and set the localization strings as the values. This approach has the benefit of defining the content of the dictionary once and they have one "implementation" for each languages.
- extract the localization code from the util library and place it into a "localization" library.

PR #2619, opened 2024-04-05
fix(schematic): updated schematic api python version to 3.10.14

linglp:
## Problem
Related to https://sagebionetworks.jira.com/browse/FDS-1942
Got an error message when trying to build docker image: 
```
#14 ERROR: process "/bin/sh -c pip install --no-cache-dir poetry==1.6.1   && poetry config --local virtualenvs.create false   && poetry run pip install \"cython<3.0.0\"   && poetry run pip install --no-build-isolation pyyaml==5.4.1   && poetry install --with prod --no-root --no-interaction --no-ansi   && mkdir /root/.synapseCache /app/app/manifests   && echo \"This is a test file.\" > /root/.synapseCache/test.txt   && chmod -R 777 /root /app" did not complete successfully: exit code: 1
```
and also: 
```
#14 11.64 The currently activated Python version 3.10.14 is not supported by the project (3.10.13).
```
The cause is because our project is using python version 3.10.13 but 3.10.14 is required. 

The docker file that we are using is using`FROM tiangolo/uwsgi-nginx-flask:python3.10`. When I looked into the base image of [that docker file](https://github.com/tiangolo/uwsgi-nginx-flask-docker/blob/master/docker-images/python3.10.dockerfile), I saw: `FROM tiangolo/uwsgi-nginx:python3.10`. And in [this docker file](https://github.com/tiangolo/uwsgi-nginx-docker/blob/master/docker-images/python3.10.dockerfile), I saw: `FROM python:3.10-bullseye`. I then checked the docker file related to `python:3.10-bulleye` [here](https://github.com/docker-library/python/blob/1b7a1106674a21e699b155cbd53bf39387284cca/3.10/bullseye/Dockerfile) and saw that `ENV PYTHON_VERSION 3.10.14`. 

I also checked the commit history of `python:3.10-bulleye` docker file and found that two weeks ago, they updated the image to use Python 3.10.14 instead of Python 3.10.13. (See commit [here](https://github.com/docker-library/python/commit/397357517f9570687adbae192f56406dd641831e)) 

## Temporary solution 
Updated our project to use python 3.10.14

[FDS-1942]

[FDS-1942]: https://sagebionetworks.jira.com/browse/FDS-1942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andrewelamb:
@linglp This seems to have failed

apps/schematic/api/prepare-python.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
-PYTHON_VERSION="3.10.13"
+PYTHON_VERSION="3.10.14"
+
+pyenv update
```
linglp, line 5: @tschaffter Can I add this line to update pyenv to the latest? Otherwise, 3.10.14 can't be downloaded because the older version of pyenv doesn't have 3.10.14 python version available.

tschaffter: The version of Python should be pinned to avoid unexpected behavior, unless a strong case can be made to using `latest`. A better solution is to update pyenv in https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/tools/devcontainers/sage/.devcontainer/Dockerfile.

I can update pyenv now as part of the bi-annual tool update.

andrewelamb (approved):

apps/schematic/api/prepare-python.sh
tschaffter: pyenv has been updated

linglp: Thank you @tschaffter !!

linglp:
@tschaffter I looked at the security issues raised by trivy, and here's a summary of how I plan to address them: 
* `python-cryptography`: this has to be addressed by turning pyopenssl package optional in schematic. See PR here: https://github.com/Sage-Bionetworks/schematic/pull/1413 because cryptography is a dependency of pyopenssl. _**This means that we could update cryptography and pyopenssl after another release of schematic**_ 

* flask: updating flask to a later version requires updating connexion. The latest version of connexion is 3.0.6 but the current version that we are using is 2.14.1. I tried updating connexion to the latest but got an error message:: 
```
Thus, connexion (>=3.0.6,<4.0.0) requires MarkupSafe (>=2.1.1).
And because schematicpy (24.2.1) depends on MarkupSafe (2.1.0), connexion (>=3.0.6,<4.0.0) is incompatible with schematicpy (24.2.1).
So, because schematic-api depends on both schematicpy (24.2.1) and connexion (^3.0.6), version solving failed.
```
I then double checked schematicpy version 24.2.1 and found out that in the lock file, MarkupSafe version is indeed 2.1.0. But the current schematic in develop branch is already using MarkupSafe 2.1.5. _**This means that we could update connexion and flask after another release of schematic**_ 

* updating werkzeug
This is a package related to flask. I think after updating flask to a newer version, this error should go away. 


Note: How about making MarkupSafe optional?
I did a `poetry show markupsafe` in schematic repo, and I could see: 
required by
 - jinja2 >=2.0
 - nbconvert >=2.0
 - pdoc *
 - werkzeug >=2.1.1

And nbconvert is required by jupyter-server.. which means that to turn it optional, we have to turn jupyter-server optional.. that’s possible, but should be its own PR and ticket.

andrewelamb (approved):

PR #2627, opened 2024-04-10
fix(schematic): mock os environ to force use of local creds when testing

andrewelamb:


linglp (approved):
Note: approve for now to unblock @andrewelamb . But this might not be needed after commenting out .env and rebuilding the dev container.

PR #2629, opened 2024-04-11
chore(sage-monorepo): update ubuntu and Python tools in devcontainer

tschaffter:
## Changelog

- update ubuntu to the latest stable version, but still using a pinned version
- update Python tools installed in the devcontainer (Poetry, pyenv)
- update pre-installed Python environments


linglp (approved):
Looks good to me. Thank you Thomas!

PR #2637, opened 2024-04-12
feat(openchallenges): implement lazy loading in searchable dropdown filters on search pages

### LINKED ISSUE #2634: [Feature] List items dynamically in searchable dropdown filter as scrolling down

OpenChallenges

### Description

With the current searchable dropdown filters, users can only see the first 100 items when they open the option box and need to use the search bar to find items not displayed.

The goal is to list a certain number of items initially when opening the dropdown, and then list more items as the user scrolls down. This can also improve data query efficiency on the web app

I believe it's called lazyLoading or visualScorll in `primeng`.

### Anything else?

_No response_

### Code of Conduct

rrchai:
- closes #2634 

## Changelogs
- Enable to dynamically load the data for the searchable dropdown filters ("Input Data Types", "Organization", "Platforms")
  - By default, it will load 25 data initially, and then load next pages as user scroll down
  - The loader animation will be displayed while loading (the data with avatar will have circle loader to match the styles)
- Refactor some codes in challenge search page to make codebase easier to fit in the logics for lazy loading

## Preview

Recording 2024-05-14 at 01 15 34

---
The bug #2648 is not fixed in this PR, as I think it needs extra backend support, i.e. filter input data types by its ids, which will be saved for another PRs.


rrchai:
The lazyLoad is enabled in all search dropdown filters, but I need more time to test the behavior when the search term is applied. I will also try to consolidate the multiple lazyLoad functions for different filters into one generic function to reduce codes.

rrchai:
Currently, when user scroll down the results, it will trigger to query next page results (0...N) with constant page size. The new scrolled results will be appended to the existing results. To avoid duplicated queries, the page number is checked to make sure the same page is only loaded once. 

However, when the search term is applied, the pagination of query results are different, which could lead to miss or duplicate results. I am checking ...

rrchai:
### Notes on Managing BehaviorSubject with Object Types

When using a `BehaviorSubject` to emit _objects_ in RxJS (like `debounceTime()`), ensure to create a new object when updating the value emitted by a BehaviorSubject: 

```ts
private platformSearchQuery: BehaviorSubject<ChallengePlatformSearchQuery> =
    new BehaviorSubject<ChallengePlatformSearchQuery>({});

setPlatformSearch(searchQuery: ChallengePlatformSearchQuery) {
    this.platformSearchQuery.next(searchQuery); // No
    this.platformSearchQuery.next({ ...searchQuery }); // Yes
}

**Comment by tschaffter** [comment]
> ensure to create a new object when updating the value emitted by a BehaviorSubject:

Why?

Note that using the spread operator results in a shallow copy of the object. For deep copy, I am using cloneDeep from [lodash](https://lodash.com/docs/4.17.4#cloneDeep).

**Comment by rrchai** [comment]
> Why?
> 
> Note that using the spread operator results in a shallow copy of the object. For deep copy, I am using [cloneDeep](https://lodash.com/docs/4.17.4#cloneDeep) from lodash.

If I don't copy or force the change on the object in `BehaviorSubject`, the `debounceTime(400)` will not work as expected. Instead of waiting for 400ms, it will trigger the change as soon as I type something in search bar. I didn't look into the exactly reason, but using the spread operator seem to work on my end. I am open to change to deep copy if that's necessary.

**Comment by rrchai** [comment]
I will try to address two more issue for lazy loading before making this pr open:
- [ ] #2648 
- [ ] As I change to let viewport of dropdown to determine the query size, it will not be triggered until the dropdown open up. As a result, it will show "No results found" initially. (not a big deal, but I will try to improve)
Recording 2024-05-13 at 09 40 28

File: libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html [hunk 1]
```diff
@@ -8,6 +8,11 @@
   [overlayOptions]="overlayOptions"
   [filter]="filter"
   (onFilter)="onSearch($event)"
+  [virtualScroll]="true"
+  [virtualScrollItemSize]="50"
```
tschaffter, line 12: Is this the number of new item fetch during scrolling? The PR changelogs mentions that the initial number of items loaded is 25. Where is this value defined?

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.scss
```diff
@@ -54,10 +58,43 @@
   }
 }
 
+.loader-container {
```
tschaffter, line 61: Do these styles defined the placeholders displayed during loading? They look great in the preview you shared in the PR description.

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
```diff
@@ -8,6 +8,11 @@
...
+  [virtualScrollOptions]="scrollOptions"
+  [lazy]="lazy"
+  (onLazyLoad)="onLazyLoad($event)"
 >
   <!-- custom search bar to avoid conflicts against default filter func -->
   <ng-template *ngIf="filterByApiClient" pTemplate="header">
```
tschaffter, line 18: In VS Code, `webhint` shows a warning for the line "<input>" element. Could this be addressed as part of this PR? The same warning is shown for another "<input>" in the other HTML file modified by this PR.

image

tschaffter (commented):
The PR works as described. This is awesome!

I noticed that there is a flickering between the placeholder and the real content states when scrolling. More precisely,

1. Scroll
2. Placeholder is displayed
3. Real data is shown
4. Placeholder is shown again
5. Same real data as previously visible are shown again

Can you reproduce this behavior?

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
rrchai: This is the number to define the item height in px in the virtual viewpoint, which used to calculate how many items to render based on the scroll position and the viewport size.

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.scss
rrchai: yes, PrimeNg has default placeholder for loaders called [skeleton](https://primeng.org/skeleton). We just need to figure out how to properly layout the loaders.

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
rrchai: Is there any other element you found having this warning, other than the screenshot one?

tschaffter: Just the two `<input>` elements mentioned. Are they also visible for you in VS Code?

tschaffter:
@rrchai Converting the PR back to draft since I saw that you made new commits. Please turn in back to Ready for review when ready.

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
```diff
@@ -8,13 +8,18 @@
...
     <div class="custom-filter-container p-multiselect-filter-container">
       <input
```
tschaffter, line 20: This is the second `<input>`

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
tschaffter: See below

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
tschaffter: Nevermind, this is the same as the first `<input>`.

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
tschaffter: This PR is modifying two HTML files. Each of them have an `<input>` elements.

rrchai: Oh okay, but I don't see the warning on the other html file~

Screen Shot 2024-05-16 at 12 23 13 PM

tschaffter: Then it's all good.

tschaffter:
Further exploration is needed about the flickering. The issue will be fixed in another PR.

tschaffter (approved):
See comment above

PR #2660, opened 2024-05-02
feat(schematic): added synapse cache purging fds-1445

andrewelamb:
- fixes [FDS-1445]
- Adds synapse cache purging
- Adds `deault_config.yaml` for globals
- `deault_config.yaml` can be overridden by `config.yaml` file for settable globals
- Synapse Cache path is setable (set to `/var/tmp/synapse`)
- Synapse Cache purging is setable (set to true)
- Updated readme

[FDS-1445]: https://sagebionetworks.jira.com/browse/FDS-1445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andrewelamb:
@jesusaurus This is the new Schematic API we are hoping to deploy soon, that would replace the old one.

I undertsand that it was suggested by you in this [PR](https://github.com/Sage-Bionetworks/schematic/pull/1202), that the synapse cache be moved to "/var/tmp/synapse". Is that still valid?

linglp (commented):
Have you tried executing one of the endpoints and see if everything is working as expected? I tried executing: http://localhost:7443/api/v1/ui/#/Storage/get_manifest_csv,
and I got an error: 
```
{
  "detail": "local variable 'byte_size' referenced before assignment",
  "status": 500,
  "title": "Internal error"
}
```
But on schematic side, I tried: 
```
store = SynapseStorage(
        access_token="<my token>", synapse_cache_path="/var/tmp/synapse"
)
```
and it worked... 
Do you have similar errors?

apps/schematic/api/schematic_api/controllers/utils.py
```diff
@@ -184,3 +193,110 @@ def download_schema_file_as_jsonld(schema_url: str) -> str:
...
+        minute_buffer (int): All files created this amount of time or older will be deleted
+    """
+    # try clearing the cache
+    # scan a directory and check size of files
+    if os.path.exists(store.root_synapse_cache):
+        maximum_storage_allowed_cache_bytes = maximum_storage_allowed_cache_gb * (
+            1024**3
+        )
+        dir_size_bytes = check_synapse_cache_size(directory=store.root_synapse_cache)
+        # if 1 GB has already been taken, purge cache before 15 min
```
linglp, line 217: could you also edit this comment? since minute_buffer is a parameter, and don't always clear cache before 15min

apps/schematic/api/schematic_api/controllers/utils.py
```diff
@@ -184,3 +193,110 @@ def download_schema_file_as_jsonld(schema_url: str) -> str:
...
+                f"{num_of_deleted_files}  files have been deleted from {store.root_synapse_cache}"
+            )
+        else:
+            # on AWS, OS takes around 14-17% of our ephemeral storage (20GiB)
+            # instead of guessing how much space that we left, print out .synapseCache here
+            LOGGER.info(f"the total size of .synapseCache is: {dir_size_bytes} bytes")
+
+
+def check_synapse_cache_size(
+    directory: str = "/root/.synapseCache",
```
linglp, line 231: shouldn't this match AWS_SYNAPSE_CACHE_PATH?

apps/schematic/api/schematic_api/controllers/utils.py
```diff
@@ -184,3 +193,110 @@ def download_schema_file_as_jsonld(schema_url: str) -> str:
...
+) -> float:
+    """use du --sh command to calculate size of .synapseCache.
+
+    Args:
+        directory (str, optional): .synapseCache directory. Defaults to '/root/.synapseCache'
```
linglp, line 236: please also edit this commnet if using AWS_SYNAPSE_CACHE_PATH

apps/schematic/api/schematic_api/controllers/utils.py
```diff
@@ -184,3 +193,110 @@ def download_schema_file_as_jsonld(schema_url: str) -> str:
...
+    int_size_match = re.match(f"^[0-9]+[{size_letter_string}]$", size_string)
+    float_size_match = re.match(f"^[0-9]+\.[0-9]+[{size_letter_string}]$", size_string)
+    if not (int_size_match or float_size_match):
+        LOGGER.error("Cannot recognize the file size unit")
+        raise ValueError("The size string doesn't match the allowed type:", size_string)
+
+    size_letter = size_string[-1]
+    size = float(size_string[:-1])
+    multiple = 1024 ** size_dict[size_letter]
+    byte_size: int = ceil(size * multiple)
```
linglp, line 280: byte size is now float right?

apps/schematic/api/schematic_api/controllers/utils.py
```diff
@@ -184,3 +193,110 @@ def download_schema_file_as_jsonld(schema_url: str) -> str:
...
+    output = subprocess.run(command, capture_output=True, check=False).stdout.decode(
+        "utf-8"
+    )
+
+    # Parsing the output to extract the directory size
+    size = output.split("\t")[0]
+    return calculate_byte_size(size)
+
+
+def calculate_byte_size(size_string: str) -> int:
```
linglp, line 253: this should be float?

apps/schematic/api/schematic_api/controllers/utils.py
andrewelamb: It's an int, thats what the ceil function does.

apps/schematic/api/schematic_api/controllers/storage_controller_impl.py
```diff
@@ -22,9 +22,15 @@
...
 from schematic_api.models.file_metadata_page import FileMetadataPage
-from schematic_api.controllers.utils import handle_exceptions, get_access_token
+from schematic_api.controllers.utils import (
+    handle_exceptions,
+    get_access_token,
+    purge_synapse_cache,
+)
 from schematic_api.controllers.paging import Page
 
+AWS_SYNAPSE_CACHE_PATH = "/var/tmp/synapse"
```
linglp, line 32: Based on the code, this is not just the cache path when schematic APIs run on AWS, but this will also becomes the cache path when schematic APIs run locally? If so, I suggest changing the variable name to something else

apps/schematic/api/schematic_api/test/test_synapse_endpoints.py
```diff
@@ -330,3 +339,22 @@ def test_get_project_manifest_metadata_page(self) -> None:
...
+
+@pytest.mark.synapse
+@pytest.mark.secrets
+class TestPurgeSynapseCache:  # pylint: disable=too-few-public-methods
+    """Tests purge_synapse_cache"""
+
+    def test_success(self) -> None:
+        """Tests for a successful purge"""
+        synapse_store = SynapseStorage(
+            access_token=SYNAPSE_TOKEN, synapse_cache_path=AWS_SYNAPSE_CACHE_PATH
```
linglp, line 352: if when running the api locally, we are also using AWS_SYNAPSE_CACHE_PATH as cache path, I think the test should be run using a different cache path.. otherwise it will clear all the cache every time the test gets run localy..

apps/schematic/api/schematic_api/controllers/utils.py
linglp: Yeah sorry for the previous comment. I just realized that this will be `int`
But do we do inline typing like this? I think having int as return type is sufficient?

jesusaurus:
@andrewelamb yeah, I still have a mild preference for `/var/tmp/synapse`. It's a standard location for linux services to store a persistent cache, and it preemptively moves the cache to a location that a service user can own, potentially making a transition to running as a service user a little simpler.

apps/schematic/api/schematic_api/controllers/utils.py
andrewelamb: Yep!

apps/schematic/api/schematic_api/controllers/utils.py
andrewelamb: Yes, looks like I forgot to go back after we went with "/var/tmp/synapse"

apps/schematic/api/schematic_api/controllers/utils.py
andrewelamb: Yep!

apps/schematic/api/schematic_api/controllers/storage_controller_impl.py
andrewelamb: I beleive this only happens in the dev contianer and not anywhere else, but I cna change it just in case.

andrewelamb:
> Have you tried executing one of the endpoints and see if everything is working as expected? I tried executing: http://localhost:7443/api/v1/ui/#/Storage/get_manifest_csv, and I got an error:
> 
> ```
> {
>   "detail": "local variable 'byte_size' referenced before assignment",
>   "status": 500,
>   "title": "Internal error"
> }
> ```
> 
> But on schematic side, I tried:
> 
> ```
> store = SynapseStorage(
>         access_token="<my token>", synapse_cache_path="/var/tmp/synapse"
> )
> ```
> 
> and it worked... Do you have similar errors?

I can't replicate this error

apps/schematic/api/schematic_api/controllers/storage_controller_impl.py
```diff
@@ -49,9 +64,26 @@ def get_asset_storage_class(asset_type: str) -> Callable:
...
+        asset_type (str): The type of storage class (will be used in the future)
+
+    Returns:
+        SynapseStorage: A synapse storage class
+    """
+    access_token = get_access_token()
+    store = SynapseStorage(
+        access_token=access_token, synapse_cache_path=SYNAPSE_CACHE_PATH
+    )
+    purge_synapse_cache(store)
```
linglp, line 82: I think I understand why I ran into an error when running endpoint: 
```
ERROR:schematic.utils.general:Cannot recognize the file size unit
INFO:werkzeug:127.0.0.1 - - [08/May/2024 18:41:44] "GET /api/v1/assetTypes/synapse/manifests/syn51078535/csv HTTP/1.1" 500 -
```
Without updating SYNAPSE_CACHE_PATH to /var/tmp/synapse, by default, the .synapseCache was saved to: /home/vscode/.synapseCache. But after the cache path update, for some reasons, when doing: 
`du -sh /var/tmp/synapse/`, it is showing 0: 

```
vscode@0b85e8ada458:/workspaces/sage-monorepo/apps/schematic/api$ du -sh /var/tmp/synapse/
0       /var/tmp/synapse/
```
I suspect similar problem will also show on AWS in deployment. It is up to you to decide if you want to have this fully figure out

linglp:
Also, I think it worths noting some other changes that were made in the PR: 
* when running schematic api locally, `purge_synapse_cache` function will now purge the cache of users. In the current schematic code base, I think it only purge caches on AWS. When users run schematic APIs locally, we don't touch their cache. I am not sure if this is a "desired" behavior since "users" will most likely to be developers and probably aware of what's going on.
* And also worth noting that we now clear cache before calling manifest download endpoint.

apps/schematic/api/schematic_api/controllers/utils.py
```diff
@@ -8,17 +8,40 @@
...
 
+# Config for various settable global values
+# Will use config.yaml if it exists, otherwise uses the example file
+# config.yaml is ignored by git so can be changed locally without accidentaly commiting it
+# To do so copy example_cofnig.yaml to config.yaml and make changes there
+if os.path.exists("config.yaml"):
+    with open("config.yaml", "r", encoding="utf-8") as file:
+        API_CONFIG = yaml.safe_load(file)
+else:
+    with open("example_config.yaml", "r", encoding="utf-8") as file:
```
linglp, line 37: this here should be "default_config.yaml" instead of "example_config.yaml"?

andrewelamb: OH yep, for soem reaosn that change didn't get commited

linglp (approved):
Thanks. I think it looks good to me.

PR #2688, opened 2024-05-17
fix(openchallenges): fix bugs related search dropdown filters on search page

### LINKED ISSUE #2648: [Bug] search dropdown filter UI shows "empty"

OpenChallenges

### Current behavior

The current search dropdown filter by default will only query first 100 results. When a selected item is not in the first 100 results + no search term applied,  it will show "empty" in the UI.

Recording 2024-04-24 at 15 26 18

### Expected behavior

The selected item labels/N should be indicated in the UI box.

### Anything else?

_No response_

### Commit ID

_No response_

### Code of Conduct

rrchai:
- fixes #2648 

## Changelogs

- Add flag to enable/disable loader animation
   - Disable the loader animation by default to avoid showing placeholders when scrolling to the data already loaded.
- Enable to dynamic set height based on two inputs: # of options (pageSize), and height of each option, so it will always show all options from the first page.
- Fixed the "empty" bug: #2648
  - If there are selected values for dropdown filter in URL params, retrieve and append the selected values to the top of items
   - If the selected value(s) in URL params is invalid, it will be ignored.
- Fixed a bug that the selected values are not shown in the UI when the search terms change, instead it will show "empty".
 
### Disable loader animation

While the loading animation looks nice, a flickering behavior has been observed (see @tschaffter's message below), which is why this PR hides the animation.

| Before  | After |
| ------------- | ------------- |
| v0 | v1  |

### Fix empty labels

| Before  | After |
| ------------- | ------------- |
| v0 2 | v1 2 |

#### b) Invalid selected values
| Before  | After |
| ------------- | ------------- |
|  | v1 3  |

### Fix selected values not shown in UI

The selected values, if present,  will always sit on the top of options, unless the values are unselected and reset by changing search term.
| Before  | After |
| ------------- | ------------- |
|  |  |


rrchai:
@tschaffter I am not sure if the changes in this pr will fix the flicking issue you encountered 🍀

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
```diff
@@ -44,32 +42,17 @@
     </ng-template>
   </ng-template>
 
-  <!-- Empty results templates -->
+  <!-- custom message for empty results -->
```
tschaffter, line 45: What is a "message"?

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
```diff
@@ -29,7 +27,7 @@
     </div>
   </ng-template>
 
-  <!-- loader template with skeleton -->
+  <!-- custom loader with skeleton -->
   <ng-template pTemplate="loader">
     <div *ngIf="showAvatar; else noIconLoader" class="loader-with-icon-container">
```
tschaffter, line 32: Do "avatar" and "icon" refer to the same concept?

libs/openchallenges/challenge-search/src/lib/challenge-search.component.ts
```diff
@@ -133,6 +135,7 @@ export class ChallengeSearchComponent
   // set default values
   defaultPageNumber = 0;
   defaultPageSize = 24;
+  defaultDropdownOptionSize = 5;
```
tschaffter, line 138: I didn't understand what this variable meant until I played with its value. I initially thought that it was related to the size of a UI component, though a size of 5 pixels wasn't making much sense. It's actually referring to a number of elements/options visible in the dropdown container.

I also noticed that this value control the number of elements/options displayed in the UI component at any time. The initial value of 5 makes the UI component feel a bit cramped. 10 looks like a better value. What do you think?

tschaffter, line 138:
```suggestion
  defaultDropdownOptionSize = 10;
```

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
```diff
@@ -3,14 +3,12 @@
...
   [placeholder]="placeholder"
-  [maxSelectedLabels]="2"
-  selectedItemsLabel="{0} items selected"
   [overlayOptions]="overlayOptions"
   [filter]="filter"
   (onFilter)="onSearch($event)"
   [virtualScroll]="true"
-  [virtualScrollItemSize]="50"
-  [virtualScrollOptions]="scrollOptions"
+  [virtualScrollItemSize]="itemHeight"
```
tschaffter, line 10: Another potential confusion around size. Use "height" instead of "size" here as it's more descriptive.

tschaffter:
Yes, enabling the loading animation still leads to a flickering. See another video below where I scroll 4 times. The first scroll is fine, but the next three scrolls feature a flickering. Tested on Firefox.

Recording 2024-05-23 at 14 26 55

libs/openchallenges/challenge-search/src/lib/challenge-search.component.ts
tschaffter: > I also noticed that this value control the number of elements/options displayed in the UI component at any time.

Does that come from our own design or from the design of the component from the library? Can we consider decoupling the two behavior? For example, displaying 10 elements but preloading 50 could make the scrolling smoother (less "bumps").

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
rrchai: yes, I'll change icon to avatar for consistency.

libs/openchallenges/challenge-search/src/lib/challenge-search.component.ts
rrchai: `defaultDropdownOptionSize` (in challenge search component)) is the page size used to query how many items for dropdown items. It also feeds the input `optionsPerPage` for the dropdown component.

[virtualScrollItemSize](https://primeng.org/multiselect#api.multiselect.props.virtualScrollItemSize) (in dropdown component) is from default setting from the component, which is the height of item displayed. That's the only value you could control how many items displayed, but not great in my experience because we would need to adjust the item height to fit N items in a fixed-height scroller viewpoint.

`optionsPerPage` (in dropdown component) is I add to manually set how many items to load per page, as this component does not have the page size option. The `optionsPerPage` input will be used to calculate the page number for lazy loading, as well as the height of scroller viewpoint. 

With the dynamical change in scroller height now, your suggestion (preload 50 items, but display 10 items in scrollbar) is easy to achieve. I'll make the adjustment for review.

rrchai: see comment below

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
rrchai: See comment above - the `virtualScrollItemSize` is default setting name from the component. I agree it's confusion, so I change the input name to `itemHeight`.

libs/openchallenges/ui/src/lib/search-dropdown-filter/search-dropdown-filter.component.html
rrchai: The message for empty results is "No results found".

rrchai:
> Yes, enabling the loading animation still leads to a flickering. See another video below where I scroll 4 times. The first scroll is fine, but the next three scrolls feature a flickering. Tested on Firefox.

Just to clarify, if we disable loading animation, will it still have flickering?

tschaffter (approved):

PR #2700, opened 2024-06-04
feat(model-ad): externalize the app config

tschaffter:
## Changelogs

- create the library `model-ad-config`
- add config file for the development server in `src/config/config.json`
- add template config file used when the app is containerized in `src/config/config.json.template`
- load the config during the initialization of the app
- inject the config server into the constructor of the not-found page

## References

- [The Twelve-Factor App](https://www.12factor.net/)

## Notes

Today, VS Code became unresponsive. After closing and reopening VS Code, we noticed that the port 4200 was in use. I assumed that it was the development server of MODEL-AD that was still running in the background. One way to kill a process that uses a specific port is using the workspace command `workspace-kill-port {port number}`. This command failed because my assumption was incorrect: it was actually the Docker container of the OpenChallenges (OC) app that was running, and the kill command can't stop a container that is using a port. One way I'm trying to prevent this specific issue (port conflicts) is to assign different ports to different projects in the monorepo. We currently have multiple projects that use the default Angular port 4200. We will change this first next time we meet.

### DIFF
--- a/apps/model-ad/app/src/config/config.json
+++ b/apps/model-ad/app/src/config/config.json
@@ -1,0 +1,9 @@
+{
+  "apiDocsUrl": "http://localhost:8000/api-docs",
+  "appVersion": "1.0.0-beta",
+  "csrApiUrl": "http://localhost:8082/api/v1",
+  "dataUpdatedOn": "yyyy-mm-dd",
+  "environment": "development",
+  "googleTagManagerId": "",
+  "ssrApiUrl": "http://openchallenges-api-gateway:8082/api/v1"
+}
--- a/apps/model-ad/app/src/config/config.json
+++ b/apps/model-ad/app/src/config/config.json
@@ -5,5 +5,5 @@
   "dataUpdatedOn": "yyyy-mm-dd",
   "environment": "development",
   "googleTagManagerId": "",
-  "ssrApiUrl": "http://openchallenges-api-gateway:8082/api/v1"
+  "ssrApiUrl": "http://model-ad-api-gateway:8082/api/v1"
 }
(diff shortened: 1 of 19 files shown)


apps/model-ad/app/src/config/config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "apiDocsUrl": "http://localhost:8000/api-docs",
+  "appVersion": "1.0.0-beta",
+  "csrApiUrl": "http://localhost:8082/api/v1",
+  "dataUpdatedOn": "yyyy-mm-dd",
+  "environment": "development",
+  "googleTagManagerId": "",
+  "ssrApiUrl": "http://openchallenges-api-gateway:8082/api/v1"
```
hallieswan, line 8:
```suggestion
  "ssrApiUrl": "http://model-ad-api-gateway:8082/api/v1"
```

libs/model-ad/not-found/src/lib/not-found.component.ts
```diff
@@ -5,6 +5,7 @@ import { RouterModule } from '@angular/router';
 import { MatCardModule } from '@angular/material/card';
 // import { ConfigService } from '@sagebionetworks/openchallenges/config';
 import { FooterComponent } from '@sagebionetworks/model-ad/ui';
+import { ConfigService } from '@sagebionetworks/model-ad/config';
```
hallieswan, line 8:
```suggestion
import { FooterComponent } from '@sagebionetworks/model-ad/ui';
import { ConfigService } from '@sagebionetworks/model-ad/config';
```

hallieswan (approved):
Looks great to me!

sagely1 (approved):
Looks good!

PR #2713, opened 2024-06-14
feat(openchallenges): add script for pulling GC challenges

vpchung:
## Changelog

* Add Python script currently used to pull data from Grand Challenge
* Add example of CSV results

> [!IMPORTANT]
> As noted during one of our standups, the pulled data cannot directly be put into the DB dump files / OC Data sheet - manual QA/QC will need to be done first.  For example, some results will return null start/end dates, despite there being dates listed on the Grand Challenge website.  For the most part, our OC Data sheet already reflects the dates found on the website, so they should ideally not be over-written.

## How to test

The API endpoint provided by Grand Challenge does not require any authentication, ergo, the script can readily be executed with:

```
python pull_grand_challenge_data.py
```


tschaffter (changes requested):
Can you please try to package this script as an Nx project to make the script more accessible. A perfect template for this is the project `apps/openchallenges/edam-etl`.

tschaffter:
@vpchung Here are a template project for Python scripts that you could use for reference. This will make the project portable, easier to develop, and enable the CI/CD workflow to capture it.

- https://github.com/Sage-Bionetworks/sage-monorepo/tree/main/apps/openchallenges/edam-etl

tschaffter:
## Update

This PR is parked until Verena's return.

tschaffter:
> [!WARNING]  
> This branch was created before the repo migrated to pnpm as the Node.js package manager. Consider creating a new branch from `main`. Alternatively, we could meet in a call to update this branch together.

tschaffter:
This PR hasn't been updated in a while so I'll go ahead and close it. Since this branch still relies on the Yarn package manager, it would be easier to create it again from `main` rather than update it.

PR #2719, opened 2024-06-20
feat(model-ad): add a reverse proxy to consolidate the stack

tschaffter:
## Changelog

- Add a reverse proxy to forward the traffic to the web app and API server

## Features

- Using Caddie instead of Nginx for better DX
- Externalized configuration using environment variables

### Build the images

```
model-ad-build-images
```

### Start the stack

```
nx serve-detach model-ad-apex
```

### Access the app and the API server

- App: `localhost:8000`
- API server (list of genes): `localhost:8000/api/v1/genes`


hallieswan (approved):
Great addition! I was able to access the app and API via the reverse proxy

PR #2722, opened 2024-06-24
refactor(openchallenges): overideable config-server spring import config

zaro0508:
spring boot [issue 3346](https://github.com/spring-projects/spring-boot/issues/33446)
says it's not possible to override Spring Import with SPRING_CONFIG_IMPORT env var.  The suggestion is to use confighost instead which can be overriden with a CONFIGHOST env var.

### DIFF
--- a/apps/openchallenges/api-gateway/src/main/resources/application.yml
+++ b/apps/openchallenges/api-gateway/src/main/resources/application.yml
@@ -4,8 +4,9 @@ spring:
   profiles:
     active: dev,local
   config:
-    import: 'configserver:http://openchallenges-config-server:8090'
+    import: 'configserver:'
   cloud:
     config:
+      uri: http://openchallenges-config-server:8090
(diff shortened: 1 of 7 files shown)


tschaffter:
## Solution

### Before

```yml
spring:
  config:
    import: 'configserver:http://openchallenges-config-server:8090'
```

### After

```
spring:
  config:
    import: 'configserver:'
  cloud:
    config:
      uri: http://openchallenges-config-server:8090
```

The value of `spring.cloud.config.uri` can be overridden with the environment variable `SPRING_CLOUD_CONFIG_URI`

### Notes

- It's not possible to use placeholder or override the value of `spring.config.import` unlike other config properties. This behavior was tested and confirmed. The solution is to rely on the property `spring.cloud.config.uri`, which can be overridden with an environment variable (see above).
- This solution does not require other developers to update their configuration files (`.env`).

PR #2723, opened 2024-06-24
refactor(openchallenges): overideable config-server spring import config

zaro0508:
spring boot [issue
3346](https://github.com/spring-projects/spring-boot/issues/33446) says it's not possible to override Spring Import with SPRING_CONFIG_IMPORT env var.  The suggestion is to use confighost instead which can be overriden with a CONFIGHOST env var.


tschaffter:
Duplicate of #2722

PR #2728, opened 2024-06-25
refactor(openchallenges): add init script to MariaDB image

tschaffter:
Closes #2727

## Changelog

- Copy the folder that include init scripts to the MariaDB image
- Remove unused Flyway config

### DIFF
--- a/docker/openchallenges/services/mariadb.yml
+++ b/docker/openchallenges/services/mariadb.yml
@@ -5,9 +5,9 @@ services:
     restart: always
     env_file:
       - ../../../apps/openchallenges/mariadb/.env
-    volumes:
-      # - openchallenges-mariadb:/data/db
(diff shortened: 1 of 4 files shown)


docker/openchallenges/services/mariadb.yml
```diff
@@ -5,9 +5,9 @@ services:
     restart: always
     env_file:
       - ../../../apps/openchallenges/mariadb/.env
-    volumes:
-      # - openchallenges-mariadb:/data/db
-      - ../../../apps/openchallenges/mariadb/docker-entrypoint-initdb.d:/docker-entrypoint-initdb.d
+    # volumes:
```
zaro0508, line 8: shouldn't we at least mount and empty volume to persist data?

zaro0508 (approved):

docker/openchallenges/services/mariadb.yml
tschaffter: It's not needed during development.

PR #2729, opened 2024-06-26
refactor(openchallenges): allow DB connection from other hosts

zaro0508:
The current MariaDb configurations does not allow remote hosts to access the DB. Now that we are deploying with AWS ECS we need to allow remote hosts to connect to the DB[1] which can be running in a different VPC subnet.  This change should allow that to happen.

[1] https://mariadb.com/kb/en/configuring-mariadb-for-remote-client-access/

### DIFF
--- a/apps/openchallenges/mariadb/my.cnf
+++ b/apps/openchallenges/mariadb/my.cnf
@@ -1,0 +1,36 @@
+# The MariaDB configuration file
+#
+# The MariaDB/MySQL tools read configuration files in the following order:
+# 0. "/etc/mysql/my.cnf" symlinks to this file, reason why all the rest is read.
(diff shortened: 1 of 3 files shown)


apps/openchallenges/mariadb/my.cnf
```diff
@@ -0,0 +1,36 @@
...
+# port = 3306
+socket = /run/mysqld/mysqld.sock
+
+# Import all .cnf files from configuration directory
+[mariadbd]
+# skip-host-cache
+# skip-name-resolve
+
+skip-networking=0
+skip-bind-address
```
zaro0508, line 33: i also tried
```
# skip-networking=0
# skip-bind-address
bind-address=0.0.0.0
```

zaro0508:
This did not resolve the DB connection issue in our AWS ECS with service connect environment.

PR #2743, opened 2024-07-17
feat(): setup workspace automatically when the dev container starts

tschaffter:
## Changelog

- Add `postStartCommand` to the dev container definition


tschaffter:
There are two issues:

- the command `workspace-install` is not available when the container starts (requires a bash session to have been created first)
- another problem is that the CI/CD workflow runs the "-affected" version of the command to save time. Running the normal + the affected version of the command would waste a bit of time. Ultimately that something that should be fixable.

PR #2748, opened 2024-07-23
feat(schematic): fds-1605 excel endpoint

andrewelamb:
- fixes [FDS-1605]
- Adds excel manifest generation endpoint

[FDS-1605]: https://sagebionetworks.jira.com/browse/FDS-1605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tschaffter:
@andrewelamb The branch is out-of-date

apps/schematic/api/schematic_api/openapi/openapi.yaml
```diff
@@ -1435,6 +1435,106 @@ paths:
...
+        required: false
+        schema:
+          default: false
+          type: boolean
+        style: form
+      - description: The ID of a dataset.
+        explode: true
+        in: query
+        name: datasetId
+        required: true
```
linglp, line 1464: The dataset id shouldn't be required. Users should be able to generate a a manifest in excel format without providing a dataset id

apps/schematic/api/schematic_api/openapi/openapi.yaml
```diff
@@ -1435,6 +1435,106 @@ paths:
...
+        name: manifestTitle
+        required: false
+        schema:
+          type: string
+        style: form
+      - description: A data type
+        explode: true
+        in: query
+        name: dataType
+        required: false
```
linglp, line 1481: here the data type should be required. Without knowing what is the data type that user wants to generate, schematic can't generate a manifest.

apps/schematic/api/schematic_api/openapi/openapi.yaml
```diff
@@ -1435,6 +1435,106 @@ paths:
...
+          enum:
+          - class_label
+          - display_label
+          type: string
+        style: form
+      - description: "If true, users are blocked from entering incorrect values. If\
+          \ false, users will get a warning when using incorrect values."
+        explode: true
+        in: query
+        name: useStrictValidation
```
linglp, line 1501: I think useStrictValidation parameter is only applicable to google sheet manifest generation. You could remove it here.

linglp (changes requested):

linglp:
@andrewelamb is it possible to make authentication not required when user just wants to generate a new manifest with data type and no dataset id and asset view? The authentication should only be required when user is interacting with synapse. Since users don't need to interact with synapse when generating a new manifest, authentication shouldn't be required.

andrewelamb:
@linglp I'm not sure if you can or not, but I'm not aware of one since currently it's a setting at the endpoint level.

andrewelamb:
@linglp I don't think so, at least as far as I can tell its required or isn't for each endpoint. One saolution could be to split it up into two different endpoints.

GiaJordan:
> @andrewelamb is it possible to make authentication not required when user just wants to generate a new manifest with data type and no dataset id and asset view? The authentication should only be required when user is interacting with synapse. Since users don't need to interact with synapse when generating a new manifest, authentication shouldn't be required.

I agree with @linglp . The auth token should be made optional (not conditionally required) for the endpoint to match the existing behavior of the manifest generation endpoint

andrewelamb:
@GiaJordan We could make it optional, do we want to do that for both the excel and google sheet version?

linglp:
@andrewelamb I think on DCA, manifests are always generated with a dataset id and asset view. I am not sure if anyone is using schematic APIs directly to generate a new manifest. I think it makes sense to double check with @AmyHeiser and @milen-sage on: 
do we expect anyone interacting with schematic APIs to generate a manifest be able to provide a synapse token?
Currently, the people who are not on synapse can still use a schema url and a data type to generate a manifest. If we change this behavior in mono repo and always require synapse token to be present, is it going to impact anyone? 

I could approve for now. Otherwise it looks good to me.

linglp (approved):

PR #2760, opened 2024-08-16
fix(openchallenges): correct the function name to trigger full text search on filters

### LINKED ISSUE #2757: [Bug] Full-text search for platforms and orgs does not work on the challenge search page

OpenChallenges

### Current behavior

Using the full-text search to filter the results in the platforms and organizations filters do not affect the content of the platforms and orgs list.

### Expected behavior

_No response_

### Anything else?

_No response_

### Commit ID

_No response_

### Code of Conduct

rrchai:
- fixes #2757 

The cause of the bug was using the wrong function name to trigger the full-text search on the corresponding filter.

## Preview

Recording 2024-08-16 at 12 44 53


tschaffter (commented):
The fix works but I noticed that there may be a race condition that leads the list to include incorrect results.

Screencast 1:

Recording 2024-08-16 at 13 37 24

Screencast 2:

Recording 2024-08-16 at 13 38 48

Screencast 3:

Recording 2024-08-16 at 13 40 36

@rrchai Is a race condition possible based on your knowledge of the implementation? If there's an easy fix we could package it with this PR. Otherwise, this PR contributes to increasing the quality of OC so I'm OK shipping it as is.

tschaffter:
Merging this PR despite the race condition issue because it significantly reduces the impact of the original bug.

@rrchai Please create a new ticket for the race condition issue I described above.

PR #2785, opened 2024-08-30
feat(agora): initial migration of packages and services (AG-1532)

sagely1:
## Description
Initial migration of Agora services

## Changelog
**Added packages for services:**
- sanitize-html package
- rollbar

**Updated package:**
- @types/node

**Added Agora services:**
- ErrorService
- HelperService
- RollbarService
- SynapseApiService

## Testing / Validation
**Run all unit tests:**
`nx run-many -t test --parallel=5`

**Run agora service unit tests:**
`nx test agora-services`


libs/agora/services/src/lib/synapse-api.service.spec.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-template-curly-in-string */
```
sagely1, line 1: Disabling this linting rule since we are ingesting wiki content from Synapse with this format.

libs/agora/services/tsconfig.lib.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "../../../dist/out-tsc",
+    "declaration": true,
+    "declarationMap": true,
+    "inlineSources": true,
+    "types": ["node"]
```
sagely1, line 8: `"types": ["node"] `
had to be specified due to linting error.

apps/agora/app/src/config/config.json
```diff
@@ -5,5 +5,6 @@
   "dataUpdatedOn": "yyyy-mm-dd",
   "environment": "development",
   "googleTagManagerId": "",
-  "ssrApiUrl": "http://agora-api:3333/v1"
-}
\ No newline at end of file
+  "ssrApiUrl": "http://agora-api:3333/v1",
+  "rollbarToken": "e788198867474855a996485580b08d03"
```
sagely1, line 9: Should this be in a gitignore?

tsconfig.base.json
```diff
@@ -10,122 +10,63 @@
     "importHelpers": true,
     "target": "es2015",
     "module": "esnext",
-    "lib": [
```
sagely1, line 13: Are we aligned on formatting?  My settings have a different line break.

apps/agora/app/src/config/config.json
tschaffter: This is the file that the dev server uses, so it shouldn't be added to `.gitignore`.

libs/agora/services/src/models/SynapseWiki.ts
tschaffter, line 1: The naming convention for TS files in the monorepo is to use kebab case. The file could be renamed to `synapse-wiki.ts`

libs/agora/services/src/models/index.ts
```diff
@@ -0,0 +1 @@
+export * from './SynapseWiki';
```
tschaffter, line 1: Rename file

libs/agora/services/jest.config.ts
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+  displayName: 'web-services',
```
tschaffter, line 2: Set to the name of the project, i.e. agora-services?

libs/agora/services/project.json
```diff
@@ -0,0 +1,21 @@
...
+  "$schema": "../../../node_modules/nx/schemas/project-schema.json",
+  "projectType": "library",
+  "sourceRoot": "libs/agora/services/src",
+  "prefix": "agora",
+  "targets": {
+    "test": {
+      "executor": "@nx/jest:jest",
+      "outputs": ["{workspaceRoot}/coverage/libs/agora/services"],
+      "options": {
+        "jestConfig": "libs/agora/services/jest.config.ts"
```
tschaffter, line 12:
```suggestion
        "jestConfig": "{projectRoot}/jest.config.ts"
```

tschaffter: This makes the file more portable.

package.json
```diff
@@ -48,7 +48,9 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "regenerator-runtime": "0.13.11",
+    "rollbar": "2.26.4",
     "rxjs": "7.8.1",
+    "sanitize-html": "2.13.0",
     "tslib": "2.4.1",
```
tschaffter, line 54: Is Agora compatible with the latest version (2.7.0)?

tsconfig.base.json
tschaffter: I'm aware of this issue where the formatting of this files and other JSON files may go back and forth between two different formattings. I think that I have a ticket that I will resurface now that we are more to use the monorepo.

tschaffter: @sagely1 This issue is now tracked in #2786

package.json
sagely1: I think 2.13.0 is the latest according to:  [link](https://www.npmjs.com/package/sanitize-html)

libs/agora/services/project.json
sagely1: great suggestion!  I copied this from somewhere else so this pattern could be applied in more places in the monorepo

libs/agora/services/jest.config.ts
sagely1: good catch!  I will update this.

libs/agora/services/README.md
```diff
@@ -0,0 +1,7 @@
+# web-services
+
+This library was generated with [Nx](https://nx.dev).
+
+## Running unit tests
+
+Run `nx test web-util` to execute the unit tests.
```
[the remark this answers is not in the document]
sagely1: I'll update this!

sagely1:
> > Run all unit tests:
> 
> @sagely1 Documented command should be passed via `nx` as much as possible to benefit from it (e.g. caching). Instead of suggesting to run `jest`, you could suggest `nx test agora-services`.

Will do!  I've updated the PR and README.md

package.json
tschaffter: You are right.

tschaffter (approved):

PR #2805, opened 2024-09-12
feat(sage-monorepo): standardize the linting of Python files (ARCH-273)

tschaffter:
Closes https://sagebionetworks.jira.com/browse/ARCH-273

## Changelog

- Update lint-staged config to lint all the projects affected by the staged files
- Update Pylint to its latest version for `schematic-api`

## Notes

The initial attempt was to install Pylint at the workspace-level. However, Pylint needs to access the dependencies of a project, for instance when checking `import-error` (see below). The simplest solution is to keep Pylint installed at the project level.

### Lint the projects affected by the staged files

image


tschaffter:
## Error when lintint and Pylint is installed in a separate virtualenv

The issue you're encountering with `pylint` is related to how it's being executed in relation to the virtual environments of your project and `pylint` itself. The errors you see (`E0401: Unable to import 'schematic' (import-error)`) occur because `pylint` is not running within the context of your project's virtual environment, where the `schematic` package is installed.

Here’s what’s likely happening and how to fix it:

### Cause of the Issue:
1. **Different Virtual Environments:** You have two separate virtual environments—one for `pylint` and one for your Python project. When you run `pylint`, it's using its own environment where it might not have access to the libraries installed in your project's virtual environment.
   
2. **Import Errors:** Since `pylint` is not running in the same environment as your project, it cannot find the dependencies of your project, resulting in the `import-error` messages.

### How to Fix It:
To resolve this, you should ensure that `pylint` runs within the context of your project's virtual environment. Here are some steps you can take:

1. **Install `pylint` in Your Project's Virtual Environment:**
   - First, activate your project's virtual environment:
     ```bash
     source /path/to/your/project/venv/bin/activate
     ```
   - Install `pylint` inside your project's virtual environment:
     ```bash
     pip install pylint
     ```

2. **Run `pylint` Inside the Project’s Virtual Environment:**
   - After activating the virtual environment, run `pylint` as you normally would:
     ```bash
     pylint schematic_api/controllers/*impl.py schematic_api/test/test*.py
     ```

3. **Check Your Path Configuration:**
   - Ensure that when you run `pylint`, it's actually using the version installed in your project’s virtual environment, not the global one or the one in another virtual environment.

4. **Avoid Path Conflicts:**
   - Make sure your `PYTHONPATH` and other environment variables are set correctly when running `pylint`, so it picks up the correct environment settings. 

By running `pylint` from within the virtual environment that contains all your project dependencies, it will be able to resolve the imports correctly, and the errors should be resolved.

tschaffter:
## Is there a way to run pylint, which is installed globally, without installing it as part of the project virtualenv?

ChatGPT:

Yes, you can run `pylint` installed globally without installing it in your project's virtual environment by explicitly specifying the Python interpreter of your project's virtual environment when running `pylint`. This allows `pylint` to recognize and use the dependencies installed in your project’s environment. Here’s how you can do it:

### Solution: Use the Virtual Environment's Python Path

1. **Use the `--init-hook` Option:**
   The `--init-hook` option allows you to specify Python commands to be executed before `pylint` starts. You can use it to set the `PYTHONPATH` to include your project’s virtual environment.

   ```bash
   pylint --init-hook "import sys; sys.path.insert(0, '/path/to/your/project/venv/lib/pythonX.X/site-packages')" schematic_api/controllers/*impl.py schematic_api/test/test*.py
   ```

   Replace `/path/to/your/project/venv/lib/pythonX.X/site-packages` with the path to the `site-packages` directory of your virtual environment.

2. **Specify the Python Interpreter:**
   You can specify the Python interpreter of your project’s virtual environment when running `pylint` to ensure it uses the correct environment.

   ```bash
   pylint --python-exec=/path/to/your/project/venv/bin/python schematic_api/controllers/*impl.py schematic_api/test/test*.py
   ```

   This tells `pylint` to use the Python interpreter from your virtual environment, which has access to your project’s dependencies.

3. **Set the PYTHONPATH Environment Variable:**
   Temporarily set the `PYTHONPATH` to include your virtual environment’s `site-packages` directory before running `pylint`:

   ```bash
   export PYTHONPATH="/path/to/your/project/venv/lib/pythonX.X/site-packages:$PYTHONPATH"
   pylint schematic_api/controllers/*impl.py schematic_api/test/test*.py
   ```

   This ensures that `pylint` can find the necessary modules when running.

### Explanation:
- **`--init-hook`:** This hook runs the specified command at the initialization of `pylint`, allowing you to modify the Python environment `pylint` uses.
- **`--python-exec`:** Directs `pylint` to use the specified Python interpreter, effectively running in the context of the specified environment.
- **`PYTHONPATH`:** Modifies the path Python uses to search for modules, ensuring it includes your project’s dependencies.

These methods allow you to run `pylint` without having to install it in your project's virtual environment, leveraging the globally installed `pylint` while ensuring it can correctly resolve imports from your project's dependencies.

andrewelamb (approved):

PR #2808, opened 2024-09-12
feat(schematic): update python packages

andrewelamb:
This PR:

- updates some python packages including schematic to 24.7.2
- fixes a typo in a test


apps/schematic/api/schematic_api/test/test_synapse_endpoints.py
```diff
@@ -140,7 +140,7 @@ def test_submit_manifest_csv(self) -> None:
         )
         body = csv_to_bytes(CORRECT_MANIFEST_PATH)
         response = self.client.open(url, method="POST", headers=HEADERS, data=body)
-        self.assert500(response, f"Response body is : {response.data.decode('utf-8')}")
+        self.assert200(response, f"Response body is : {response.data.decode('utf-8')}")
```
linglp, line 143: why this gets changed to `assert200`? is it because of a change in recent schematic?

linglp (approved):
looks good to me overall.

apps/schematic/api/schematic_api/test/test_synapse_endpoints.py
linglp: oh i see. that was a typo before?

PR #2810, opened 2024-09-12
feat(schematic): create app and lib environment with `@nxlv/python` (ARCH-275)

tschaffter:
## Description

- TODO

## Changelog

- Add the Nx plugin `@nxlv/python`
- Create a lib and app project with the NX plugin
- Add the lib as a dependency of the app project
- Remove the need for Black to be added to the PATH

## TODO

- Identify how to prepare the Python projects to build and test them in the CI workflow
- Identify how to prepare the Python projects to develop them in VS Code
- Identify how to manage the Python version of the projects

Currently these two tasks are failing in the CI workflow:

- schematic-core:lint
- schematic-api-next:lint

## References

- https://betterprogramming.pub/poetry-python-nx-monorepo-5750d8627024
- https://github.com/lucasvieirasilva/python-poetry-monorepo
- https://www.npmjs.com/package/@nxlv/python

### Project dependency

`nx graph`:

graph(1)

### Create the app project

```console
$ nx generate @nxlv/python:poetry-project schematic-api-next --directory=apps/schematic/api-next

 NX  Generating @nxlv/python:poetry-project

CREATE apps/schematic/api-next/project.json
CREATE apps/schematic/api-next/README.md
CREATE apps/schematic/api-next/.python-version
CREATE apps/schematic/api-next/schematic_api_next/__init__.py
CREATE apps/schematic/api-next/schematic_api_next/hello.py
CREATE apps/schematic/api-next/poetry.toml
CREATE apps/schematic/api-next/pyproject.toml
CREATE apps/schematic/api-next/tests/__init__.py
CREATE apps/schematic/api-next/tests/conftest.py
CREATE apps/schematic/api-next/tests/test_hello.py
CREATE apps/schematic/api-next/.flake8
UPDATE pyproject.toml
  Updating root poetry.lock...
Resolving dependencies... (0.4s)

Writing lock file
Running command: poetry install 

Installing dependencies from lock file

Package operations: 1 install, 0 updates, 0 removals

  - Installing schematic-api-next (1.0.0 /workspaces/sage-monorepo/apps/schematic/api-next)

  poetry.lock updated.
```

### Create the library project

```console
$ nx generate @nxlv/python:poetry-project schematic-core --directory=libs/schematic/core --projectType=library

 NX  Generating @nxlv/python:poetry-project

CREATE libs/schematic/core/project.json
CREATE libs/schematic/core/README.md
CREATE libs/schematic/core/.python-version
CREATE libs/schematic/core/schematic_core/__init__.py
[...]

PR #2835, opened 2024-09-19
feat(sage-monorepo): setup Python monorepo v3 (ARCH-275)

tschaffter:
## Description

This iteration takes inspiration from the lessons learned and documented in #2810.

Main contributions:
- This PR demonstrates how to import a local Python project (lib) into another Python project (app).
- This PR uses a Poetry project in the root folder to install "global" tools like Black

## Changelog

- Add the project `sandbox-py-app`
- Add the project `sandbox-py-lib`
- Add `sandbox-py-lib` as a dependency of `sandbox-py-lib`
- Add `numpy` to `sandbox-py-lib`
- Add the command `workspace-nuke-venv`
- Document a use case where a feature is developed across a Python lib and an app
- Run Black with Poetry

## Notes

- [nxlv/python](https://www.npmjs.com/package/@nxlv/python) was used to create the two "individual" projects. Since then, the root folder includes a Poetry project to install global tools. Because of this root Poetry files, the [nxlv/python](https://www.npmjs.com/package/@nxlv/python) believes that ALL the Python projects are sharing a single virtual environment (`.venv`), which is incorrect. For that reason, we will only rely on a subset of the methods offered by the [nxlv/python](https://www.npmjs.com/package/@nxlv/python) from now on (mainly the ones used in `project.json`). In particular, future Python projects should be created by copying existing projects.

- Projects are linted with `flake8`. This could be changed to `pylint` in another PR.

## Create projects

The two projects `sandbox-py-app` and `sandbox-py-lib` were created using [nxlv/python](https://www.npmjs.com/package/@nxlv/python) when there were no Poetry project in the root folder. This note is important because the projects were then created as "individual" projects that do not share a venv. Now that we have a Poetry project in the root for another reason, creating projects with [nxlv/python](https://www.npmjs.com/package/@nxlv/python) would result in non-individual projects, which is currently not supported/recommended. For now, new projects could be created by copying the projects `sandbox-py-app` and `sandbox-py-lib`.

Here is how these projects were created:

```console
nx generate @nxlv/python:poetry-project sandbox-py-app --directory=apps/sandbox/py-app
nx generate @nxlv/python:poetry-project sandbox-py-lib --directory=libs/sandbox/py-lib --projectType=library
```

### Add the lib as a dependency of the app project

This was done using [nxlv/python](https://www.npmjs.com/package/@nxlv/python). Because we currently only support individual Python projects but [nxlv/python](https://www.npmjs.com/package/@nxlv/python) thinks that we are in a shared virtual environment, it's recommended to look at the `pyproject.toml` of `sandbox-py-app` to learn how to import other Python projects developed in this monorepo.

Here is how the lib was added as a dependency of the app:

```console
nx run sandbox-py-app:add --name sandbox-py-lib --local
```

### Install global Python tools

A Python venv is created in the root folder when running `workspace-install`.

### Preparing a Python project

Preparing a Python project involves creating a venv for it and install its dependencies. This is currently done as part of `workspace-install` for all the Python projects, but this may change in the future.

To prepare a specific project, e.g.:

```console
nx prepare sandbox-py-app
```

> [!TIP]
> If this project depends on other local projects (libs), it is not required to prepare these local projects.

### Use Case: Add a feature to a library and use it in an app

Let's assume that we are implementing a new feature in `sandbox-py-app` that has `sandbox-py-lib` as a dependency.

First, we create the venv for the app and install its dependencies:

```console
nx prepare sandbox-py-app
```

In VS Code, we open the Command Palette (Cmd + Shift + P) and select the interpreter by specifying the path below. This will enable VS Code to enable IntelliSense and auto-completion. 

```console
apps/sandbox/py-app/.venv/bin/python
```

image

We decide that part of the business logic of this feature should be implemented in the project `sandbox-py-lib` and that we need to add the package `numpy` to `sandbox-py-lib`.

```console
$ cd libs/sandbox/py-lib
$ poetry add numpy
```

Then we open a Python file of the library `sandbox-py-lib` and try to import numpy. We get the following error:

image

This happens because we are using the Python interpreter of the app (`sandbox-py-app`), which doesn't know anything about numpy. The solution is to switch to the Python interpreter of the library before contributing to it.

image

The error should then immediately disappear. If not, try restarting VS Code.

image

After implementing the feature in the library project, let's use it in the app project. First, we need to switch back to the interpreter of the app project. Because we have previously used it, it should already be listed in the "Select Python Interpreter" menu. If may even be marked as Recommended as shown below.

image

After updating the code in the app to use the new feature from the library, we could try to run the project. Given the current state of the app project, we can run the existing script with:

```console
$ poetry run python sandbox_py_app/hello.py 
Traceback (most recent call last):
  File "/workspaces/sage-monorepo/apps/sandbox/py-app/sandbox_py_app/hello.py", line 1, in <module>
    import sandbox_py_lib.hello
  File "/workspaces/sage-monorepo/libs/sandbox/py-lib/sandbox_py_lib/hello.py", line 4, in <module>
    import numpy as np
ModuleNotFoundError: No module named 'numpy'
```

The same error would occur if we were to run tests with `nx test sandbox-py-app`. The reason is because the venv of the app has been created before we added numpy as a dependency of the library project `sandbox-py-lib`.

To update the app project dependencies and hence install the numpy package, we run:

```console
vscode@7d06102369c4:/workspaces/sage-monorepo/apps/sandbox/py-app$ poetry update
Updating dependencies
Resolving dependencies... (2.1s)

Package operations: 1 install, 0 updates, 0 removals

  - Installing numpy (2.1.1)

Writing lock file
```

or by running the task `update` that applies the above command:

```console
nx update sandbox-py-app
```

The new feature should now be working in the app! 🚀


tschaffter:
## Error linting app and lib in CI workflow

```console
> nx run sandbox-py-app:lint


  Running flake8 linting on project  sandbox-py-app ...

The currently activated Python version 3.10.12 is not supported by the project (3.10.14).
Trying to find and use a compatible version. 

Poetry was unable to find a compatible version. If you have one, you can explicitly use it via the "env use" command.

   ERROR  ENOENT: no such file or directory, open '/workspaces/sage-monorepo/reports/apps/sandbox/py-app/pylint.txt'
```

tschaffter:
## Weird Poetry error

> The current project's supported Python range (>=3.9,<3.12) is not compatible with some of the required packages Python requirement:
>  - numpy requires Python >=3.10, so it will not be satisfied for Python >=3.9,<3.10

PR #2836, opened 2024-09-20
feat(sage-monorepo): adopt JS for definining OAS validation ruleset

tschaffter:
## Description

Switch OAS validation ruleset from YAML to JS. This will enable us to further customize rules beyond just setting them to `off` and `level`. For example, we will be able to specify that path segments should be camel case with a level set to warn, then later error.

## Changelog

- Switch OAS validation ruleset from YAML to JS
- Update the task `agora-api-description:lint` to use the monorepo-wide OAS validation config

### DIFF
--- a/libs/agora/api-description/project.json
+++ b/libs/agora/api-description/project.json
@@ -14,8 +14,7 @@
     "lint": {
       "executor": "nx:run-commands",
       "options": {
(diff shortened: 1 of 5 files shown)


sagely1 (approved):
LGTM!

PR #2847, opened 2024-09-23
fix(sage-monorepo): fix `schematic-api:build-image` issue by using a non-moving tag (FDS-2453)

tschaffter:
Fixes https://sagebionetworks.jira.com/browse/FDS-2453

## Description

The task `schematic-api:build-image` failed 3-4 hours ago. I realized that this is because 1) the base image has been updated around that time and 2) the project is using a moving tag (`python3.10`).

The underlying issue seems to be that the base image comes with a newer version of Python (3.10.15) not compatible with the version specified by the project (3.10.14). Ideally, we should be able to specify the version of Python to use in the Dockerfile, e.g. using pyenv. This would provide some flexibility so that the version used by the project is not imposed by the Python version running in the Docker image. For instance, it's preferable to use one of the Python version pre-installed in the Sage monorepo dev container for running tasks faster locally and in the CI workflow.

image

## Changelog

- Specify a non-moving tag for the base Docker image of `schematic-api`

Cc: @andrewelamb @linglp

### DIFF
--- a/apps/schematic/api/Dockerfile
+++ b/apps/schematic/api/Dockerfile
@@ -1,4 +1,4 @@
-FROM tiangolo/uwsgi-nginx-flask:python3.10
+FROM tiangolo/uwsgi-nginx-flask:python3.10-2024-09-16
 
 # add label
 LABEL org.opencontainers.image.authors='Milen Nikolov <milen.nikolov@sagebase.org>, Andrew Lamb <andrew.lamb@sagebase.org>, Mialy DeFelice <mialy.defelice@sagebase.org>, Gianna Jordan <gianna.jordan@sagebase.org>, Lingling Peng <lingling.peng@sagebase.org>'


apps/schematic/api/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM tiangolo/uwsgi-nginx-flask:python3.10
+FROM tiangolo/uwsgi-nginx-flask:python3.10-2024-09-16
```
linglp, line 1: Thanks for the change @tschaffter . I could pre-approve for now. I am curious why you selected `python3.10-2024-09-16`. Is it because this keeps the underlying Python version the same? 

Also, I am still seeing that in the base image of `tiangolo/uwsgi-nginx-flask:python3.10` [here](https://github.com/tiangolo/uwsgi-nginx-flask-docker/blob/master/docker-images/python3.10.dockerfile), they are pulling from the docker image [here](https://github.com/tiangolo/uwsgi-nginx-docker/blob/master/docker-images/python3.10.dockerfile) which is still using `python:3.10-bullseye` as the base Python image.  Does that mean python updated `python:3.10-bullseye` to a different python version?

linglp (approved):
Thanks for noticing this problem @tschaffter !! It looks good to me. But I am curious

apps/schematic/api/Dockerfile
tschaffter: These are great questions. Regarding the first one, my hypothesis was that the issue stemmed from a recent update to `tiangolo/uwsgi-nginx-flask:python3.10` and the fact that Schematic uses a moving tag. I then searched for a non-moving tag for `python3.10`, such as those that typically include a date. I selected the second most recent tag, which was built around a week ago (`python3.10-2024-09-16`). After rebuilding the image using this tag, the `schematic-api:build-image` task succeeded.

Yes, it's likely that `python:3.10-bullseye` is also a moving tag, which introduced changes to `tiangolo/uwsgi-nginx-flask:python3.10`.

linglp: ooooh I see. Thank you so much! Using a moving tag seems to be a bad idea then. They are using a moving tag which caused their python version to change, and since we are also using a moving tag, that ends up affecting us.

PR #2854, opened 2024-09-25
feat(sage-monorepo): format HTML files with Prettier

tschaffter:
Closes https://sagebionetworks.jira.com/browse/ARCH-290

## Changelog

- Remove webhint (packages, VS Code extension and config)
- Configure lint-staged to format HTML files
- Format all the HTML files with Prettier


tschaffter:
Here is the ESLint config I wanted to add to the root config file:

```json
    {
      "files": ["*.html"],
      "parser": "@html-eslint/parser",
      "extends": ["plugin:@html-eslint/recommended"],
      "plugins": ["@html-eslint", "html"],
      "rules": {
        "@html-eslint/indent": ["error", 2],
        "@html-eslint/require-closing-tags": "off"
      }
    }
```

Note: The two rules were to silence errors based on a single HTML file used for testing.

An issue is that Angular projects come with this local ESLint rule that breaks the config for the HTML files in the root ESLint config file.

```json
    {
      "files": ["*.html"],
      "extends": ["plugin:@nx/angular-template"],
      "rules": {}
    }
```

Note: It doesn't even look like the later config overrides the previous.

This error has been previously reported here:

- https://github.com/eslint/eslint/issues/14286
- https://github.com/eslint/eslint/discussions/18808

tschaffter:
This [comment](https://stackoverflow.com/a/76867576) include a reference about the Angular ESLint plugin not wanting to implement formatting rules, which makes sense. I'm still looking for a way to enable "basic" rules like checking that images have an alternative text, which doesn't seem available in the Angular ESLint plugin.

tschaffter:
Based on this [comment](https://github.com/nrwl/nx/issues/5259#issuecomment-814904289), what `"plugin:@nx/angular-template"` currently does it extend from `@angular-eslint/eslint-plugin:recommended`.

This seems to be confirmed on this Nx docs page:
- https://nx.dev/nx-api/eslint-plugin#angular

tschaffter:
My feeling is that we should continue to use `@nx/angular-template` in case Nx updates `@nx/angular-template` to does one day more than extending `plugin:@angular-eslint/template/recommended`. Later on, we can customize the rules, e.g. by using `plugin:@angular-eslint/template/accessibility`.

```json
    {
      "files": ["*.html"],
      "extends": ["plugin:@nx/angular-template", "plugin:@angular-eslint/template/accessibility"]
    }
```

PR #2864, opened 2024-09-29
feat(agora): migrate Nominated Targets (AG-1551)

sagely1:
## Description

- Adding Nominated Targets Component
** Note: table on page will be implemented in a future PR
- Removing teams dependency on legacy ApiService.  Moving to OpenApi implementation.
- Site-wide CSS updates (also using normalize.css to serve as a css reset)
- Added shared components - SvgIcon and ModalLink
   
## Testing
`nx test agora-nominated-targets`

## Expected output
Navigate to:
http://localhost:4200/genes/nominated-targets

image


apps/agora/api/.eslintrc.json
```diff
@@ -12,7 +12,9 @@
     },
     {
       "files": ["*.js", "*.jsx"],
-      "rules": {}
+      "rules": {
+        "tailwindcss/no-custom-classname": "off"
```
sagely1, line 16: Added rule as there was a linting error in webpack.config.js

tschaffter: We could ignore `webpack.config.js` with:

```
"files": ["*.js", "*.jsx", "!webpack.config.js"],
```

tschaffter: We should either ignore `webpack.config.js` or disable the rule just for `webpack.config.js`. The former is either. Note that `webpack.config.js` may be removed in the near future when we switch to esbuild.

sagely1: sounds good!

tschaffter (approved):

PR #2896, opened 2024-10-13
feat(sandbox): prototype locust integration to the monorepo

tschaffter:
## Changelog

- Add the project `sandbox-locust`
- Add locust file examples for Agora and OpenChallenges
- Configure the task `sandbox-locust:serve` for Agora and OpenChallenges
- Add GH workflow to perform a load test

## Development workflow

The project `sandbox-locust` provides a place to develop load tests as locust files (Python project). The base configuration of locust is defined in `pyproject.toml` under the section `[tool.locust]`.

The `project.json` file of this project defines the `serve` task, which can be customized with multiple configurations. For example, this PR adds two configuration examples for Agora and OpenChallenges:

```json
    "serve": {
      "executor": "nx:run-commands",
      "options": {
        "command": "poetry run locust",
        "cwd": "{projectRoot}"
      },
      "configurations": {
        "agora": {
          "locustfile": "sandbox_locust/locustfile_agora.py",
          "host": "http://localhost:8000"
        },
        "openchallenges": {
          "locustfile": "sandbox_locust/locustfile_openchallenges.py",
          "host": "https://openchallenges.io"
        }
      },
      "defaultConfiguration": "agora"
    }
```

Run the load test for the specified configuration with:

```console
nx serve sandbox-locust --configuration=agora  # OR
nx run sandbox-locust:serve:agora
```

A developer can run the load test from their development environment using the commands above after starting their stack. Additionally, they could pass extra CLI options to the task, allowing them to apply a subset of [Locust tasks](https://docs.locust.io/en/stable/writing-a-locustfile.html) to test specific components of the stack (e.g., the REST API).

This PR includes a draft of a GitHub workflow to run the load test from GitHub. If the workflow is intended to test a local deployment, an additional step is needed to configure the stack before serving it. The load test should work for OpenChallenges, as the host is set to its production instance.

## Future work

In the future, we could consider the following ideas:

- Run a variant of this workflow as part of the deployment of the stage environment.
- Analyze the content of the Locust reports and make the workflow fail if the response time of the endpoints do not meet specific requirements.

### Perform a load test on Agora

```console
agora-build-image
nx serve-detach agora-apex

nx run sandbox-locust:serve:agora
```

Set `headless` to false in `pyproject.toml` and run the load test with the previous command. The locust web app will open in your browser.

image

image


apps/sandbox/locust/project.json
```diff
@@ -0,0 +1,70 @@
...
+      }
+    },
+    "serve": {
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "poetry run locust",
+        "cwd": "{projectRoot}"
+      },
+      "configurations": {
+        "agora": {
```
thomasyu888, line 52: Since this is a PoC, this looks great! In the future, we may want to think about how we support folders or locust.py files so that the load testing py scripts can be more modular.

.github/workflows/locust-load-test.yml
```diff
@@ -0,0 +1,53 @@
...
+      #   run: |
+      #     devcontainer exec --workspace-folder ../sage-monorepo bash -c ". ./dev-env.sh \
+      #       && nx run-many --target=build-image --projects=${{ inputs.product }}-*" \
+      #       && nx serve-detach ${{ inputs.product }}-apex \
+      #       && nx serve sandbox-locust --configuration=${{ inputs.product }}
+
+      - name: Perform a load test
+        run: |
+          devcontainer exec --workspace-folder ../sage-monorepo bash -c ". ./dev-env.sh \
+            && nx run sandbox-locust:serve:${{ inputs.product }}
```
thomasyu888, line 47: I like this a lot!  Do you think the endpoint parameter should be parametrized here?

thomasyu888 (approved):
LGTM!  Minor comments, but I like the idea of adding load testing framework here.

apps/sandbox/locust/pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+autopep8 = "2.0.2"
+flake8 = "6.0.0"
+pytest = "7.3.1"
+pytest-sugar = "0.9.7"
+pytest-cov = "4.1.0"
+pytest-html = "3.2.0"
+
+[tool.locust]
+# locustfile = ""
+headless = true
```
sagely1, line 34: consider having a setting in the future to allow UI to be used without risking accidental check-in of changes to this variable.

sagely1 (approved):
Overall looks great!  Love the load-testing integration!

apps/sandbox/locust/pyproject.toml
sagely1: `--headful` and `--headless` can be used to override

.github/workflows/locust-load-test.yml
tschaffter: Which endpoint parameter? The only URL part parametrized for now is the host, which is define in the task configurations `sandbox-locust:serve:{product}`.

apps/sandbox/locust/pyproject.toml
tschaffter: The task `sandbox-locust:serve` now runs Locust headful. `--headless` is now specified in the CI workflow.

.github/workflows/locust-load-test.yml
thomasyu888: I see. That makes sense, you can have multiple tasks per different host like dev, staging, prod.

PR #2941, opened 2024-12-10
feat(iatlas): import github.com/CRI-iAtlas/iatlas-app (ARCH-362)

tschaffter:
## Changelog

- Create the project `iatlas-app`.


tschaffter:
I chose not to import the iAtlas Shiny app into the monorepo due to the excessive time required to install its dependencies. This is primarily because 1) the project has an exceptionally large number of dependencies, and 2) many R packages must be built and installed from source.

"Adding the Shiny app was a stretch goal to consolidate all components in one place. However, it is not essential, as Sage is only responsible for maintaining the iAtlas backend.

Cc: @andrewelamb

PR #2967, opened 2025-01-15
feat(agora): change agora-api to construct mongo URI from host, username, password, port, and DB name env variables (AG-1592)

hallieswan:
## Description

Changes agora-api to construct the mongo URI with host, username, password, port, and DB name env variables, rather than specifying the URI directly in the env file. Future PR in `agora-infra-v3` will set the environment variables to secrets allowing connection to existing DocumentDB instances, as described [here](https://github.com/Sage-Bionetworks-IT/agora-infra-v3?tab=readme-ov-file#secrets).

## Related Issue

- [AG-1592](https://sagebionetworks.jira.com/browse/AG-1592)

## Changelog

- Changes environment variables for agora-api
- Updates agora_version in agora-app/config to align with latest deployment tag
- Adds env variables missing from agora-app .env.example, which are used to populate agora-app config.json
- Fixes healthcheck for agora-apex -- container does not include `curl`, so use `wget` instead
   
## Validation

1. Update `apps/agora/api/.env` and `apps/agora/app/.env` to create new environment variables specified in `.env.example` files
2. Stop any running containers: `workspace-docker-stop`.
3. Start agora stack: `nx run agora-apex:serve-detach`.
4. Navigate to teams page: localhost:8000/teams.
5. Confirm that teams and user images are displayed

[AG-1592]: https://sagebionetworks.jira.com/browse/AG-1592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/agora/api/src/api.ts
```diff
@@ -14,13 +14,27 @@ import {
...
+  process.exit(1);
+}
+
+if (!mongodbHost) {
+  console.error('No MONGODB_HOST environment variable has been defined.');
+  process.exit(1);
+}
+
+const mongoUri = `mongodb://${mongodbUser}:${mongodbPass}@${mongodbHost}:27017/agora?authSource=admin`;
+
```
tschaffter, line 37: An alternative would be to define a JSON Schema that describes the config file, then use it to validate the configuration object. Out of scope for this PR but if you are interested, I provide some pointers towards libraries that can be used to achieve that.

apps/agora/api/src/api.ts
```diff
@@ -14,13 +14,27 @@ import {
...
+  console.error('No MONGODB_PASS environment variable has been defined.');
+  process.exit(1);
+}
+
+if (!mongodbHost) {
+  console.error('No MONGODB_HOST environment variable has been defined.');
+  process.exit(1);
+}
+
+const mongoUri = `mongodb://${mongodbUser}:${mongodbPass}@${mongodbHost}:27017/agora?authSource=admin`;
```
tschaffter, line 36: The port and DB table could be externalized too.

apps/agora/api/src/api.ts
hallieswan: Oh interesting! I'd love to learn more -- I created a ticket [here](https://sagebionetworks.jira.com/browse/AG-1626), feel free to add any notes you'd like!

apps/agora/api/src/api.ts
hallieswan: Done in the next commit

sagely1 (approved):
LGTM! We can further optimize as per Thomas's suggestions in another PR if you'd like.

apps/agora/api/src/api.ts
```diff
@@ -14,13 +14,39 @@ import {
...
+  console.error('No MONGODB_PORT environment variable has been defined.');
+  process.exit(1);
+}
+
+if (!mongodbName) {
+  console.error('No MONGODB_NAME environment variable has been defined.');
+  process.exit(1);
+}
+
+const mongoUri = `mongodb://${mongodbUser}:${mongodbPass}@${mongodbHost}:${mongodbPort}/${mongodbName}?authSource=admin`;
```
zaro0508, line 48: This is fine when mongo is running in ECS because API containers have access to `mongodbHost` via [ECS service connect](https://github.com/Sage-Bionetworks-IT/agora-infra-v3/blob/dev/src/ecs_stack.py#L32) (or service discovery).  It will probably not work when connecting to the agora AWS document db though.  You'll additionally need to [update the role or policy](https://github.com/Sage-Bionetworks-IT/agora-infra-v3/blob/dev/src/service_stack.py#L36-L59) to allow the API container running api services to access document db.

hallieswan: Oh good point, thanks! Would [`AmazonDocDBFullAccess`](https://docs.aws.amazon.com/documentdb/latest/developerguide/docdb-managed-policies.html#AmazonDocDBFullAccess) be an appropriate managed policy to use? Will we also need to update the DocumentDB security group to allow the API container to connect?

apps/agora/api/src/api.ts
tschaffter: @hallieswan I've shared some guidelines in the Jira ticket. Thank you for taking the time to review this suggestion!

tschaffter (approved):
I followed the testing instructions, including updating the config files. I confirm that the team members are visible on the team page.

apps/agora/api/src/api.ts
tschaffter: Resolving this comment before merging, as it seems the discussion is continuing on Slack, as noted by Hallie.

zaro0508: > Oh good point, thanks! Would [AmazonDocDBFullAccess](https://docs.aws.amazon.com/documentdb/latest/developerguide/docdb-managed-policies.html#AmazonDocDBFullAccess) be an appropriate managed policy to use?

not sure, it probably depends on what the API needs to do.  Best practice is to only allow what's needed (least privileged access).

> Will we also need to update the DocumentDB security group to allow the API container to connect?

yes, you'll definitely need a security that allows db port access.

PR #2985, opened 2025-01-28
chore(openchallenges): remove OC API client for Python

tschaffter:
## Description

We are removing the OC API client for Python, as it is currently unused. This change aligns with our ongoing migration of Python projects in the monorepo to the `uv` package manager.


tschaffter:
images

PR #3003, opened 2025-02-08
fix(agora): reset boxplot before updating data (AG-1673)

hallieswan:
## Description

Fixes bug where boxplot data would persist after changing the model. 

## Related issue

- [AG-1673](https://sagebionetworks.jira.com/browse/AG-1673)

## Changelog

- Resets boxplot chart data when input data is updated
- Adds types in RNA and Protein component methods

## Validation

Build agora images: `agora-build-images`
Run app: `nx run agora-apex:serve-detach`
Navigate to PLEC RNA tab: http://localhost:8000/genes/ENSG00000178209/evidence/rna --

https://github.com/user-attachments/assets/5540c223-767b-470b-9763-5c3d7c6744be

Navigate to PLEC Protein page: http://localhost:8000/genes/ENSG00000178209/evidence/protein -- 

https://github.com/user-attachments/assets/8b62f17a-cf64-4f64-ac98-3937f0fc604d

[AG-1673]: https://sagebionetworks.jira.com/browse/AG-1673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/agora/api-client-angular/src/lib/model/proteomicsDistribution.ts
+++ b/libs/agora/api-client-angular/src/lib/model/proteomicsDistribution.ts
@@ -11,9 +11,16 @@
  */
 
 /**
- * Distributions
+ * ProteomicsDistribution
  */
 export interface ProteomicsDistribution {
+  _id: string;
+  tissue: string;
+  min: number;
+  max: number;
+  median: number;
+  first_quartile: number;
+  third_quartile: number;
   /**
    * Type of proteomics distribution (e.g., LFQ, SRM, TMT)
    */
--- a/libs/agora/charts/src/lib/box-plot-chart/box-plot-chart.component.ts
+++ b/libs/agora/charts/src/lib/box-plot-chart/box-plot-chart.component.ts
@@ -45,7 +45,17 @@ export class BoxPlotComponent {
   @Input() yAxisMin: number | undefined;
   @Input() yAxisMax: number | undefined;
 
+  reset() {
+    this.points = [];
+    this.summaries = [];
+    this.xAxisCategoryToTooltipText = {};
+    this.isInitialized = false;
+    this.pointTooltipFormatter = undefined;
+  }
+
   init() {
+    this.reset();
+
     if (!this._data?.length) {
       return;
     }
--- a/libs/agora/genes/src/lib/components/gene-evidence-proteomics/gene-evidence-proteomics.component.ts
+++ b/libs/agora/genes/src/lib/components/gene-evidence-proteomics/gene-evidence-proteomics.component.ts
@@ -1,12 +1,17 @@
 import { Component, inject, Input } from '@angular/core';
 
-import { DistributionService, Gene } from '@sagebionetworks/agora/api-client-angular';
-import { ChartRange } from '@sagebionetworks/agora/models';
-import { HelperService } from '@sagebionetworks/agora/services';
-import { GeneProteinSelectorComponent } from '../gene-protein-selector/gene-protein-selector.component';
+import {
+  DistributionService,
+  Gene,
+  ProteinDifferentialExpression,
+  ProteomicsDistribution,
+} from '@sagebionetworks/agora/api-client-angular';
 import { BoxPlotComponent } from '@sagebionetworks/agora/charts';
-import { DownloadDomImageComponent } from '../download-dom-image/download-dom-image.component';
+import { boxPlotChartItem, ChartRange } from '@sagebionetworks/agora/models';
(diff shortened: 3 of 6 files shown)


libs/agora/api-client-angular/src/lib/model/proteomicsDistribution.ts
```diff
@@ -11,9 +11,16 @@
  */
 
 /**
- * Distributions
+ * ProteomicsDistribution
```
hallieswan, line 14: Updated this type to match [the model](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/agora/models/distribution.ts#L12-L21).

libs/agora/charts/src/lib/box-plot-chart/box-plot-chart.component.ts
```diff
@@ -45,7 +45,17 @@ export class BoxPlotComponent {
...
+    this.points = [];
+    this.summaries = [];
+    this.xAxisCategoryToTooltipText = {};
+    this.isInitialized = false;
+    this.pointTooltipFormatter = undefined;
+  }
+
   init() {
+    this.reset();
+
```
hallieswan, line 58: Fixes AG-1673.

libs/agora/genes/src/lib/components/gene-evidence-proteomics/gene-evidence-proteomics.component.ts
```diff
@@ -36,26 +41,26 @@ export class GeneEvidenceProteomicsComponent {
   uniProtIds: string[] = [];
   selectedUniProtId = '';
 
-  LFQData: any = undefined;
+  LFQData: boxPlotChartItem[] = [];
```
hallieswan, line 44: The remaining changes just add types for easier troubleshooting in the future

sagely1 (changes requested):
I'm still seeing scaling issues.  The y-axis scale seems to creep up as before.

hallieswan:
> I'm still seeing scaling issues. The y-axis scale seems to creep up as before.

@sagely1 The max log2 fold change value changes depending on the selected model, so the max y-axis tick location does change accordingly. However, the y-axis should be consistent for each model. That behavior is similar to the current [dev site](https://agora-develop.adknowledgeportal.org/genes/ENSG00000171862/evidence/rna) -- 

https://github.com/user-attachments/assets/1cc3d0b7-a0a0-4516-94d0-c53862604a81

Are you seeing something else "creeping up" on the y-axis? Just trying to understand what the expected y-axis should look like.

hallieswan:
> > I'm still seeing scaling issues. The y-axis scale seems to creep up as before.
> 
> @sagely1 The max log2 fold change value changes depending on the selected model, so the max y-axis tick location does change accordingly. However, the y-axis should be consistent for each model. That behavior is similar to the current [dev site](https://agora-develop.adknowledgeportal.org/genes/ENSG00000171862/evidence/rna) --
> 
>  AG-1673_current_develop.mov 
> Are you seeing something else "creeping up" on the y-axis? Just trying to understand what the expected y-axis should look like.

I was able to repro the issue by switching between two models repeatedly. It looks like the y-axis min/max weren't reset when the model was changed. Ensuring that the min/max are always reset fixes the bug -- 

http://localhost:8000/genes/ENSG00000178209/evidence/rna

https://github.com/user-attachments/assets/f185d58f-40de-4652-87c9-d42c3c52fe2b

sagely1:
> > > I'm still seeing scaling issues. The y-axis scale seems to creep up as before.
> > 
> > 
> > @sagely1 The max log2 fold change value changes depending on the selected model, so the max y-axis tick location does change accordingly. However, the y-axis should be consistent for each model. That behavior is similar to the current [dev site](https://agora-develop.adknowledgeportal.org/genes/ENSG00000171862/evidence/rna) --
> > AG-1673_current_develop.mov
> > Are you seeing something else "creeping up" on the y-axis? Just trying to understand what the expected y-axis should look like.
> 
> I was able to repro the issue by switching between two models repeatedly. It looks like the y-axis min/max weren't reset when the model was changed. Ensuring that the min/max are always reset fixes the bug --
> 
> http://localhost:8000/genes/ENSG00000178209/evidence/rna
> 
>  AG-1673_yaxis.mov

Looks like the y-axis scaling issue is fixed now.  Looks great!

sagely1 (approved):
Looks good!

PR #3004, opened 2025-02-10
fix(agora): fix routing (AG-1645)

sagely1:
## Description
/genes route not redirecting to home page.

## Changelog
/genes and /genes/gene-router have been migrated


hallieswan (approved):
Looks good! Would it be worth adding a few e2e tests demonstrating the routing works as expected?

PR #3006, opened 2025-02-10
fix(agora): fixing team member link styling (AG-1646)

sagely1:
## Description
Team member link styling was using browser defaults.

This change fixes the styling and co-locates the hyperlink styling logic.  

Before:
image

After:
image

### DIFF
--- a/libs/agora/teams/src/lib/team-member-list/team-member-list.component.scss
+++ b/libs/agora/teams/src/lib/team-member-list/team-member-list.component.scss
@@ -1,5 +1,5 @@
-@import 'libs/agora/styles/src/lib/variables';
-@import 'libs/agora/styles/src/lib/mixins';
+@use 'libs/agora/styles/src/lib/variables';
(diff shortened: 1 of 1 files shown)


libs/agora/teams/src/lib/team-member-list/team-member-list.component.scss
```diff
@@ -42,8 +42,15 @@
...
+    a {
+      color: var(--color-link);
+      cursor: pointer;
+      text-decoration: none;
+
+      &:hover {
+        color: var(--color-link);
+        text-decoration: underline;
+      }
+    }
```
hallieswan, line 54: Could we use the imported `mixins` instead so the shared styles are used?

```suggestion
    a {
      @include mixins.link;

      &:hover {
        @include mixins.link-hover;
      }
    }
```

hallieswan (changes requested):

hallieswan (approved):
Looks great!

PR #3012, opened 2025-02-14
fix(agora): fix team members sorting (AG-1650)

sagely1:
### Description
Fixed bug in team member names not sorting correctly due to a property not matching the casing in MongoDB (isprimaryinvestigator).

Note: In the future, for consistency we should consider coordinating with @JessterB to use snake casing to match other properties in the database.


hallieswan (approved):
Looks good to me! I was able to confirm the teams sort for AMP-AD: Columbia University - Rush University matches current Agora dev -- 

image

PR #3017, opened 2025-02-20
fix(agora): nominated targets layout bug (AG-1652)

sagely1:
### Description
Fixes layout styling above the gene table (alignment, missing icon)
Removed Material styling
Adds a custom preset based on Lara for PrimeNG styling
Updates PrimeNG to the latest version (v19 final)

Helpful information on theme customization:
https://primeng.org/theming
https://medium.com/@abdulkhader_95798/primeng-theme-customization-v19-in-angular-f371c31dc601

Before:
image

After:
image


hallieswan:
I ran the "Run e2e tests for Agora" job to demonstrate that #3016 was working as expected. However, the job failed because this branch does not contain the fix added by #3013. This branch could be rebased on main and then the job rerun, if desired, but this failing run should not prevent this PR from being merged since the failure is unrelated to these changes.

apps/agora/app/src/app/app.config.ts
```diff
@@ -7,7 +7,7 @@ import {
 import { provideHttpClient, withInterceptors } from '@angular/common/http';
 import { provideAnimationsAsync } from '@angular/platform-browser/animations/async';
 import { providePrimeNG } from 'primeng/config';
-import Lara from '@primeng/themes/lara';
+import { MyPreset } from './myPreset';
```
hallieswan, line 10: What do you think about moving this into the existing `agora-themes` library to colocate with other theme related code? We could also more easily incorporate into Storybook stories

hallieswan: Also, what do you think about using a more descriptive name, e.g. `AgoraPreset`?

apps/agora/app/src/app/myPreset.ts
```diff
@@ -0,0 +1,50 @@
+import { definePreset } from '@primeng/themes';
+import Lara from '@primeng/themes/lara';
+
+export const MyPreset = definePreset(Lara, {
```
hallieswan, line 4: Should this theme and the colors in [`_variables.scss`](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/agora/styles/src/lib/_variables.scss#L44-L97) be aligned?

libs/agora/styles/src/lib/components/_button.scss
```diff
@@ -1,19 +1,21 @@
...
   text-decoration: none;
   box-shadow: none;
   user-select: none;
   transition: $transition-duration;
   cursor: pointer;
 }
 
+.p-button-label {
+  font-weight: normal;
+}
```
hallieswan, line 17: The theme [docs](https://primeng.org/theming#component) state that 
> Overriding components tokens is not the recommended approach if you are building your own style, building your own preset should be preferred instead.

What do you think about including this in the preset instead? I think something like this should work -- 

```
// myPreset.ts
export const MyPreset = definePreset(Lara, {
  semantic: {...},
  components: {
    button: {
      label: {
        fontWeight: 'normal',
      }
    }
  }
});
```

Alternately, we could open another ticket to address moving the component overrides in styles into the preset theme?

hallieswan (changes requested):

libs/agora/styles/src/lib/components/_button.scss
sagely1: I created a preset to handle this.  It wasn't documented but there is a wizard to create a custom preset.

apps/agora/app/src/app/myPreset.ts
sagely1: I've incorporated colors into a custom preset.  I will leave _variables.scss and similar files for now with the goal of simplifying these styles in the future.

apps/agora/app/src/app/app.config.ts
sagely1: You're right.  I'll use myPrimeNGPreset

apps/agora/app/src/app/myPrimeNGPreset.ts
```diff
@@ -0,0 +1,5422 @@
+import { ApplicationConfig } from '@angular/core';
+import { provideAnimationsAsync } from '@angular/platform-browser/animations/async';
+import { providePrimeNG } from 'primeng/config';
+import Lara from '@primeng/themes/lara';
+import { definePreset } from '@primeng/themes';
+
+export const MyPreset = definePreset(Lara, {
```
hallieswan, line 7: Does this file include the entire PrimeNG theme? To improve readability and maintainability, what do you think about only including the fields where we've set custom values? That would help highlight the specific styles we're managing.

apps/agora/app/src/app/myPreset.ts
hallieswan: Would you mind opening a ticket to capture the work to align the `_variables.scss` colors with the preset colors?

apps/agora/app/src/_app-theme.scss
```diff
@@ -1,25 +1,7 @@
...
-  (
-    color: (
-      primary: $primary,
-      accent: $accent,
-    ),
-    typography: fonts.$lato,
-    density: 0,
-    is-dark: false,
-  )
-);
```
hallieswan, line 22: Do you mind pointing me to where these new values are managed in the new preset?

sagely1: These values don't exist in the original Agora codebase.  These were OC defaults as this is a part of Material.

apps/agora/app/src/app/myPrimeNGPreset.ts
sagely1: Yes, this is actually the recommended approach of creating a preset.

apps/agora/app/src/app/myPreset.ts
sagely1: Sure.  It is being tracked by [AG-1524](https://sagebionetworks.jira.com/browse/AG-1524).  The goal is to remove all PrimeNG related style overrides in the libs/styles folder

apps/agora/app/src/app/myPrimeNGPreset.ts
hallieswan: We discussed together and @sagely1 made the point that keeping the entire theme will make it easier to migrate the overrides from the SCSS files into the preset. If needed, we can diff `MyPreset` and the base preset generated by the PrimeNG theme designer to see which values have been customized.

hallieswan (approved):

PR #3021, opened 2025-02-26
fix(agora): only show agora's e2e workflow as waiting on PRs that affect agora (SMR-19)

hallieswan:
## Description

We should only show Agora's e2e workflow as "waiting" on PRs that affect Agora.

## Validation

Here is a run in [my fork](https://github.com/hallieswan/sage-monorepo/actions/runs/13554103650/attempts/1) that shows the two jobs and that Agora e2e tests run when Agora is affected.

### DIFF
--- a/.github/workflows/e2e-agora.yaml
+++ b/.github/workflows/e2e-agora.yaml
@@ -8,9 +8,50 @@ on:
     types: [opened, synchronize, reopened, labeled]
 
 jobs:
-  run-agora-e2e-tests:
+  check-agora-affected:
     # Run in Sage repo on main branch and on all branches in user-owned forks
     if: ${{ github.ref_name == 'main' || github.actor == github.repository_owner }}
+    runs-on: ubuntu-22.04
+    outputs:
+      agora_affected: ${{ steps.agora_affected.outputs.AFFECTED }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # We need to fetch all branches and commits so that Nx affected has a base to compare
+          # against.
+          fetch-depth: 0
+          persist-credentials: false
+          # By default, actions/checkout@v4 will checkout the main branch instead of the merge
+          # commit when when using pull_request_target. It is currently difficult to checkout the
+          # merge commit in this context. The current solution is to checkout the PR HEAD instead
+          # and enable the branch protection rule "Require branches to be up to date before
+          # merging".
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-agora.yaml
```diff
@@ -53,49 +47,77 @@ jobs:
           done
           echo "AFFECTED=false" >> "${GITHUB_OUTPUT}"
 
+      - name: Remove the dev container
+        run: docker rm -f sage_devcontainer
+  run-agora-e2e-tests:
```
tschaffter, line 52: Add a blank line on top.

.github/workflows/e2e-agora.yaml
```diff
@@ -8,13 +8,12 @@ on:
...
-  run-agora-e2e-tests:
+  check-agora-affected:
     # Run in Sage repo on main branch and on all branches in user-owned forks
     if: ${{ github.ref_name == 'main' || github.actor == github.repository_owner }}
-    environment: ${{ github.event_name == 'pull_request_target' &&
-      github.event.pull_request.head.repo.full_name != github.repository && 'agora-pr' || 'agora' }}
-    timeout-minutes: 60
     runs-on: ubuntu-22.04
+    outputs:
+      agora_affected: ${{ steps.agora_affected.outputs.AFFECTED }}
```
tschaffter, line 16: We should use the same naming convention / casing for all step outputs (`...outputs.xxx`). The casing should also be updated where the output is exported and wherever needed.

```suggestion
      agora_affected: ${{ steps.agora_affected.outputs.affected }}
```

.github/workflows/e2e-agora.yaml
```diff
@@ -53,49 +47,77 @@ jobs:
           done
           echo "AFFECTED=false" >> "${GITHUB_OUTPUT}"
 
+      - name: Remove the dev container
+        run: docker rm -f sage_devcontainer
+  run-agora-e2e-tests:
+    needs: check-agora-affected
+    if: needs.check-agora-affected.outputs.agora_affected == 'true'
+    environment: ${{ github.event_name == 'pull_request_target' &&
```
tschaffter, line 55: Could you please add a comment on top to describe the behavior of this `environment` expression? This should make this bit of code more understandable and hence reduce the chances of bugs.

tschaffter (changes requested):

tschaffter (approved):
LGTM!

PR #3026, opened 2025-03-06
fix(agora): remove gene-druggability component (AG-1710)

hallieswan:
## Description

The `gene-druggability` component depends on a deprecated PrimeNG component (`Accordion`). However, since we don't plan to use the `gene-druggability` component in the future, we can remove this component rather than updating to the latest PrimeNG components. 

## Related Issues 

- [AG-1710](https://sagebionetworks.jira.com/browse/AG-1710)
- [AG-1486](https://sagebionetworks.jira.com/browse/AG-1486)

[AG-1710]: https://sagebionetworks.jira.com/browse/AG-1710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AG-1486]: https://sagebionetworks.jira.com/browse/AG-1486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3032, opened 2025-03-10
fix(agora): migrate InputSwitch to ToggleSwitch (AG-1709), remove styles specific to removed gene-druggability component (AG-1710)

hallieswan:
## Description

We recently updated to PrimeNG v19 from v17. There were a number of changes to the component names and style classes from v17 to v18+, see [here](https://primeng.org/guides/migration). The `InputSwitch` component was renamed `ToggleSwitch`. This PR updates the Agora components accordingly.

We also recently removed an unused component (#3026). This PR also removes remaining styles specific to this component.

## Related Issues

- [AG-1709](https://sagebionetworks.jira.com/browse/AG-1709)
- [AG-1710](https://sagebionetworks.jira.com/browse/AG-1710)

## Changelog

- Migrates Agora components to use `ToggleSwitch` instead of `InputSwitch`
- Moves global toggleswitch styles into preset
- Removes styles specific to removed component

## Validation

Current dev: 

https://github.com/user-attachments/assets/318beea9-2246-4ecc-9106-1a8957b8e20d

New dev: 

https://github.com/user-attachments/assets/fc51ca25-e2c1-47b6-9c8e-f40d3ef3f1a0

Fix: 

https://github.com/user-attachments/assets/cb14ecaf-b143-4791-b799-1bfb6035411f

[AG-1709]: https://sagebionetworks.jira.com/browse/AG-1709?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AG-1710]: https://sagebionetworks.jira.com/browse/AG-1710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/agora/styles/src/lib/components/_shame.scss
```diff
@@ -1,303 +1,3 @@
-/* stylelint-disable no-descending-specificity */
-
-.gd-accordion {
```
hallieswan, line 3: Style specific to the removed gene-druggability component.

libs/agora/styles/src/lib/components/_shame.scss
```diff
@@ -1,303 +1,3 @@
...
-  }
-}
-
-p-inputswitch {
-  display: block;
-
-  .p-inputswitch {
-    display: block;
-    width: 36px;
-    height: 20px;
```
hallieswan, line 251: Moved into the preset. All the other inputswitch styles were redundant with the new preset styling.

sagely1 (approved):
LGTM!

tschaffter:
@hallieswan Why is the box titled "RNA-Differential Expression" shifted vertically when interacting with the toggle that lives inside this box?

sagely1 (approved):
LGTM!

hallieswan:
> @hallieswan Why is the box titled "RNA-Differential Expression" shifted vertically when interacting with the toggle that lives inside this box?

When the switch is toggled, a filter is added so that only differential expression for genes with significant p-values are displayed. Displaying the filter shifts the control panel downwards (including the RNA - Differential Expression dropdown).

https://github.com/user-attachments/assets/66496694-26a1-4a96-a195-73c7d7ca8caa

tschaffter:
@hallieswan What do you think about keeping the filter bar always displayed to avoid the vertical shifts. Ideally, the user should be able to click several times on the toggle to change its state without having to move the cursor. If no active filter are displayed, the "Clear All" button could be disabled.

hallieswan:
> @hallieswan What do you think about keeping the filter bar always displayed to avoid the vertical shifts. Ideally, the user should be able to click several times on the toggle to change its state without having to move the cursor. If no active filter are displayed, the "Clear All" button could be disabled.

@tschaffter great idea! I opened [a ticket](https://sagebionetworks.jira.com/browse/AG-1719) so that @JessterB can take a look and we can track the work. Thanks!

PR #3042, opened 2025-03-20
feat(agora): updating parameter handling (AG-1713)

sagely1:
### Description
This modernizes Angular to use the latest parameter handling introduces in Angular 16.

This was suggested by @tschaffter and leads to more readable code as you don't have to subscribe to router events.


apps/agora/app/src/app/app.component.ts
```diff
@@ -12,31 +11,9 @@ import { ToastModule } from 'primeng/toast';
...
-      });
-    });
-  }
-
-  getChildRoute(activatedRoute: ActivatedRoute): ActivatedRoute {
-    if (activatedRoute.firstChild) {
-      return this.getChildRoute(activatedRoute.firstChild);
-    } else {
-      return activatedRoute;
-    }
```
sagely1, line 40: Cleaned up this code in favor of a single service to handle setting the title and meta tags when routes change

apps/agora/app/src/app/app.config.ts
```diff
@@ -68,6 +69,7 @@ export const appConfig: ApplicationConfig = {
     },
     provideRouter(
       routes,
+      withComponentInputBinding(),
```
sagely1, line 72: This allows us to use @Input binding in our components

libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.ts
```diff
@@ -85,21 +87,28 @@ import { GeneComparisonToolLegendPanelComponent } from './components/gene-compar
...
   helperService = inject(HelperService);
   messageService = inject(MessageService);
   filterService = inject(FilterService);
 
+  /* Query Parameters ------------------------------------------------------ */
+  @Input('category') categoryParam = '';
+  @Input('subCategory') subCategoryParam = '';
+  @Input('sortField') sortFieldParam = '';
+  @Input('sortOrder') sortOrderParam = '';
+  @Input('significance') significanceParam = '';
```
sagely1, line 101: My own convention I've come up with is to add 'Param' as a suffix.  One could use it without the aliasing.  However, many of these variables (like category and subCategory already exist, are modified after initialization and are heavily used in the code).  For clarity sake, adding this aliasing should make the code a little easier to read.

libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.ts
```diff
@@ -1,14 +1,16 @@
+/* eslint-disable @angular-eslint/no-input-rename */
```
sagely1, line 1: When using aliasing for the input parameters, ESLint complains.

libs/agora/services/src/lib/meta-tag.service.ts
```diff
@@ -0,0 +1,41 @@
...
+      .pipe(
+        filter((event) => event instanceof NavigationEnd),
+        map(() => this.activatedRoute),
+        map((route) => {
+          while (route.firstChild) route = route.firstChild;
+          return route;
+        }),
+        filter((route) => route.outlet === 'primary'),
+        mergeMap((route) => route.data),
+      )
```
sagely1, line 28: essentially on successful navigation, get the data from app.routes.ts

tschaffter:
Cc @rrchai who is applying the same update to OC.

libs/agora/services/src/lib/meta-tag.service.ts
```diff
@@ -0,0 +1,41 @@
...
+      .subscribe((data) => {
+        if (data['title']) {
+          this.titleService.setTitle(data['title']);
+        }
```
hallieswan, line 32: The original code fell back to 'Agora' when 'title' didn't exist, do we want to maintain something similar in the new service or is it unnecessary?

apps/agora/app/src/app/app.component.ts
hallieswan: Nice! This is a lot cleaner

libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.ts
```diff
@@ -172,27 +181,32 @@ export class GeneComparisonToolComponent implements OnInit, AfterViewInit, OnDes
...
   @ViewChild('pinnedGenesModal') pinnedGenesModal!: PinnedGenesModalComponent;
 
-  constructor() {
-    this.category = 'RNA - Differential Expression';
-  }
-
   ngOnInit() {
-    this.urlParamsSubscription = this.route.queryParams.subscribe((params) => {
-      this.urlParams = params || {};
+    this.category = this.categoryParam ? this.categoryParam : this.categories[0].value;
```
hallieswan, line 185: What do you think about creating a `defaultCategory` variable that can be used here and to initialize `category` on line 110?

```suggestion
    this.category = this.categoryParam ? this.categoryParam : this.defaultCategory;
```

libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.ts
```diff
@@ -172,27 +181,32 @@ export class GeneComparisonToolComponent implements OnInit, AfterViewInit, OnDes
...
-      this.loadGenes();
-    });
+    if (this.significanceParam) {
+      const parsedValue = parseFloat(this.significanceParam);
+      if (!isNaN(parsedValue)) {
+        this.significanceThreshold = parsedValue;
+      }
+    }
+
+    this.significanceThresholdActive = !!this.significanceParam;
```
hallieswan, line 207: Should `significanceThreshold` continue to fall back to `this.DEFAULT_SIGNIFICANCE_THRESHOLD`?  

Should `significanceThresholdActive` depend on whether `significanceThreshold` was set using the value of `significanceParam`?

```suggestion
  // default values
   this.significanceThreshold = this.DEFAULT_SIGNIFICANCE_THRESHOLD;
   this.significanceThresholdActive = false;
   
   // update if possible using param
    if (this.significanceParam) {
      const parsedValue = parseFloat(this.significanceParam);
      if (!isNaN(parsedValue)) {
        this.significanceThreshold = parsedValue;
        this.significanceThresholdActive = true;
      }
    } 
```

hallieswan (commented):
Looks like a great improvement! Just a couple small questions.

libs/agora/services/src/lib/meta-tag.service.ts
sagely1: Adding a default fallback!

sagely1: There is technically a fallback but it is nice to have a programmatic fallback.  I noticed it was 'agora-app' defined in index.html and I will update it.

libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.ts
sagely1: In this case, the category varaiable is already initialized

libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.ts
sagely1: I think the original code used the following line to handle the active state. I do think just setting the boolean would be clearer though since I am checking for a valid number.
    this.significanceThresholdActive = !!this.significanceParam;

hallieswan (approved):
Looks great to me!

PR #3043, opened 2025-03-20
fix(agora): standardizing display of true and false values (AG-1665)

sagely1:
### Description
This PR standardizes the display of boolean values that use the shared component gene-table to capitalize the first letter of boolean values.  To accomplish this a custom pipe was created as the value may be a boolean or a string.  Some columns are booleans and some are tri-state (true, false or 'No data').

### Testing
The shared component is used on the /genes/{gene}/similar and /genes/nominated-targets routes.
Both were tested for issues.

Before:
image

After:
image

### DIFF
--- a/libs/agora/util/src/lib/pipes/capitalize.pipe.ts
+++ b/libs/agora/util/src/lib/pipes/capitalize.pipe.ts
@@ -1,0 +1,17 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'capitalize',
+})
+export class CapitalizePipe implements PipeTransform {
+  transform(value: any): string {
+    if (typeof value === 'boolean') {
+      return value.toString().charAt(0).toUpperCase() + value.toString().slice(1);
+    }
+    if (typeof value === 'string') {
+      if (value === 'true') return 'True';
+      if (value === 'false') return 'False';
+    }
+    return value;
+  }
+}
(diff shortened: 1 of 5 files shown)


libs/agora/util/src/lib/pipes/capitalize.pipe.ts
tschaffter, line 1: @sagely1 Would https://angular.dev/api/common/TitleCasePipe be an alternative solution?

sagely1: I considered this!  I would have loved to use it.  Unfortunately, since this is a shared component and the data type of the column can be of several variations (including numbers), a custom component had to be created to handle the various scenarios.  TitleCasePipe cannot unfortunately handle the number type.

libs/agora/util/src/lib/pipes/capitalize.pipe.ts
```diff
@@ -0,0 +1,17 @@
+import { Pipe, PipeTransform } from '@angular/core';
```
sagely1, line 1: Custom Angular pipe to handle booleans and strings (and ignore other types)

hallieswan: What do you think about adjusting the pipe's name to reflect that it only capitalizes the first letter of booleans and booleans stored as strings (e.g. something like `booleanCapitalizeFirst` / `BooleanCapitalizeFirst`)?

hallieswan (approved):
Looks good! I just made one suggestion on naming [here](https://github.com/Sage-Bionetworks/sage-monorepo/pull/3043#discussion_r2006594148), but up to you whether that change is worth making.

libs/agora/util/src/lib/pipes/capitalize.pipe.ts
sagely1: Good suggestion.  However, it also handles "no data" to "No data" if that is the case so I ended up with a more generic name.

hallieswan (approved):

PR #3059, opened 2025-03-31
fix(agora): fix GCT header colors

sagely1:
### Description
Fixed GCT header button colors by refactoring SVG icon component to be able to specify custom size and color.
This necessitated changes to all icon references.
Now the colors should match the container element but can be overridden if a custom color is specified.
Other components that include the custom icon component have been updated to leverage the new functionality.

Additional tweaks:
Made text "Hide insignificant" unselectable (new feature)
To standardize on tooltip locations, tooltips now appear above the control (matches the location of nominated targets table) (new feature)

See video for reference:

Original:
https://www.loom.com/share/7938b4d463e54a7abbeb19edd45e3b7b

New:
https://www.loom.com/share/f8a64f2e4f4743ea81359212aa64287f


libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.html
```diff
@@ -111,15 +111,15 @@ <h1 class="gct-heading h2">Gene Comparison Tool</h1>
...
+                  <div class="unselectable-text">Hide insignificant</div>
                   <div>
                     <p-toggleswitch
                       [pTooltip]="
                         'When on, this will filter out all statistically insignificant results. Current significance cutoff is: ' +
                         significanceThreshold
                       "
                       styleClass="gct-significance-control-toggle"
-                      tooltipPosition="left"
+                      tooltipPosition="top"
```
sagely1, line 122: These positions are new.  Now they match with the nominated targets table

hallieswan (changes requested):

libs/agora/shared/src/lib/svg-icon/svg-icon.component.ts
```diff
@@ -8,18 +10,25 @@ import { Component, Input, ViewEncapsulation } from '@angular/core';
...
-    }
+  svgContent: SafeHtml | null = null;
 
-    return classes.join(' ');
+  ngOnInit() {
+    if (this.imagePath) {
+      this.http.get(this.imagePath, { responseType: 'text' }).subscribe(
+        (svgContent) => {
+          // Sanitize the SVG content to prevent XSS attacks
+          this.svgContent = this.sanitizer.bypassSecurityTrustHtml(svgContent);
```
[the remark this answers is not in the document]
sagely1: Good question.  I think the documentation is not as clear as it could be.  My understanding is that sanitize() is not typically called but rather the other methods will call sanitize and they have return types typically used for creating exceptions.  Since our goal is to use HTML content inserted via the [innerHTML] attribute, we would use bypassSecurityTrustHTML to create an exception for HTML content with a trusted source.  By using this method, it returns HTML that will preserve HTML elements/attributes Angular will strip out by default so if anything we are telling Angular not to sanitize it since we trust this source as HTML content.  I'll create a check to ensure this is not accidentally with external URLs.  Also, I realize the comment on line 27 could be clearer so I'll update that as well.

libs/agora/shared/src/lib/svg-icon/svg-icon.component.spec.ts.off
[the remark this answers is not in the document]
sagely1: Updated!

libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.html
```diff
@@ -188,7 +188,7 @@ <h1 class="gct-heading h2">Gene Comparison Tool</h1>
                     <button
                       (click)="columnPopover.toggle($event)"
                       pTooltip="Show/hide columns"
-                      tooltipPosition="left"
+                      tooltipPosition="top"
```
hallieswan, line 191: This tooltip is squished when I test locally -- could the position have been set to `left` to prevent this? 

image

libs/agora/shared/src/lib/svg-icon/svg-icon.component.ts
```diff
@@ -8,18 +11,38 @@ import { Component, Input, ViewEncapsulation } from '@angular/core';
...
-    return classes.join(' ');
+  async loadSVG() {
+    try {
+      const svgContent = await firstValueFrom(
+        this.http.get(this.imagePath, { responseType: 'text' }),
+      );
+      this.svgContent = this.sanitizer.bypassSecurityTrustHtml(svgContent);
+    } catch (error) {
+      console.error('Error loading svg', error);
+    }
```
hallieswan, line 46: Should we consider caching or reducing the number of requests for the same icon? For example, on the GCT page, we fetch `info-circle.svg` 12 times

image

hallieswan (changes requested):

libs/agora/shared/src/lib/svg-icon/svg-icon.component.ts
sagely1: Yes, absolutely.  Since attempting to cache within this component will lead to race conditions, I've created an SVG icon caching service to handle this.

libs/agora/shared/src/lib/svg-icon/svg-icon.component.spec.ts
```diff
@@ -1,55 +1,59 @@
...
+    expect(component.isValidImagePath('/some-other-path/icon.svg')).toBe(false);
+    expect(component.isValidImagePath('')).toBe(false);
   });
 
-  describe('getClasses', () => {
-    it('should return only the default class when customClasses is not set', () => {
-      component.customClasses = '';
-      expect(component.getClasses()).toBe('svg-icon');
-    });
+  it('should load SVG and sanitize it', async () => {
```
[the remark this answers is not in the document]
sagely1: you're right that is a better name

libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.html
sagely1: I'll remove the little arrow (in the middle) in CSS so we don't have to deal with perfect alignment anymore

libs/agora/shared/src/lib/svg-icon/svg-icon.component.ts
```diff
@@ -8,18 +11,32 @@ import { Component, Input, ViewEncapsulation } from '@angular/core';
...
 
-    if (this.customClasses) {
-      classes.push(this.customClasses);
-    }
+  svgContent: SafeHtml | null = null;
+  private static svgCache = new Map<string, SafeHtml>();
 
-    return classes.join(' ');
+  ngOnInit() {
+    if (!this.isValidImagePath(this.imagePath)) return;
```
hallieswan, line 29: What do you think about moving this check into `SvgIconService`?

libs/agora/shared/src/lib/svg-icon/svg-icon.component.ts
```diff
@@ -8,18 +11,32 @@ import { Component, Input, ViewEncapsulation } from '@angular/core';
...
   encapsulation: ViewEncapsulation.None,
 })
-export class SvgIconComponent {
-  @Input() imagePath = '';
+export class SvgIconComponent implements OnInit {
+  @Input() imagePath!: string;
   @Input() altText = '';
-  @Input() customClasses = '';
+  @Input() width = 14;
+  @Input() height = 14;
```
hallieswan, line 18: The icon dimensions were previously set in CSS using the `.svg-icon` classname. Now that the component takes height and width as inputs, do other templates need to be updated to size the icon with these inputs instead? For example, see the info icon size on the quick filters -- it looks like [this CSS](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/agora/gene-comparison-tool/src/lib/components/gene-comparison-tool-filter-panel/gene-comparison-tool-filter-panel.component.scss#L199-L201) is no longer setting the icon's height, so the icon is much smaller now: 

newdev | this pr
:---:|:---:
image | image

libs/agora/shared/src/lib/svg-icon/svg-icon.component.spec.ts
```diff
@@ -1,55 +1,84 @@
...
   });
 
-  describe('getClasses', () => {
-    it('should return only the default class when customClasses is not set', () => {
-      component.customClasses = '';
-      expect(component.getClasses()).toBe('svg-icon');
-    });
+  it('should not call service if path is invalid', () => {
+    // Spy on service method
+    const serviceSpy = jest.spyOn(svgService, 'getSvg');
```
hallieswan, line 75: I think you can remove this and use `svgService` directly below.

```suggestion
```

libs/agora/shared/src/lib/svg-icon/svg-icon.component.spec.ts
```diff
@@ -1,55 +1,84 @@
...
+    // Set invalid path and trigger ngOnInit
+    component.imagePath = 'invalid/path';
+    component.ngOnInit();
 
-    it('should handle multiple custom classes', () => {
-      component.customClasses = 'custom-class another-class';
-      expect(component.getClasses()).toBe('svg-icon custom-class another-class');
-    });
+    // Verify service was not called
+    expect(serviceSpy).not.toHaveBeenCalled();
```
hallieswan, line 82:
```suggestion
    // Verify service was not called
    expect(svgService.getSvg).not.toHaveBeenCalled();
```

libs/agora/shared/src/lib/svg-icon/svg-icon.component.spec.ts
```diff
@@ -1,55 +1,84 @@
...
+    expect(component.isValidImagePath('')).toBe(false);
+  });
+
+  it('should load SVG through service and set as svgContent', async () => {
+    const mockSvg = '<svg><circle cx="50" cy="50" r="40" /></svg>';
+    const validPath = '/agora-assets/icons/cog.svg';
+    const sanitizedSvg = sanitizer.bypassSecurityTrustHtml(mockSvg);
+
+    // Spy on service method
+    jest.spyOn(svgService, 'getSvg').mockReturnValue(of(sanitizedSvg));
```
hallieswan, line 60: Might be worth clearing the mocks in a `beforeEach` to ensure isolation between tests? 

```
beforeEach(() => {
  jest.clearAllMocks();
})
```

hallieswan (changes requested):

libs/agora/shared/src/lib/svg-icon/svg-icon.component.ts
sagely1: yeah, you're right

libs/agora/shared/src/lib/svg-icon/svg-icon.component.ts
JessterB: I'm fine releasing 4.0 with a small infobutton in the Quick Filter panel. 

I created https://sagebionetworks.jira.com/browse/AG-1764 to track this issue.

libs/agora/genes/src/lib/components/popover-link/popover-link.component.spec.ts.off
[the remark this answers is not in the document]
sagely1: To be addressed in AG-1627

libs/agora/shared/src/lib/svg-icon/svg-icon.component.spec.ts
```diff
@@ -41,16 +43,6 @@ describe('SvgIconComponent', () => {
...
 
-  it('should not load SVG if imagePath is invalid', async () => {
-    // Test valid path
-    expect(component.isValidImagePath('/agora-assets/icons/cog.svg')).toBe(true);
-
-    // Test invalid paths
-    expect(component.isValidImagePath('https://external-domain.com/icon.svg')).toBe(false);
-    expect(component.isValidImagePath('/some-other-path/icon.svg')).toBe(false);
-    expect(component.isValidImagePath('')).toBe(false);
-  });
```
hallieswan: Since it was already written, could this test be moved in a new `svg-icon.service.spec.ts` file?

libs/agora/gene-comparison-tool/src/lib/components/gene-comparison-tool-filter-list/gene-comparison-tool-filter-list.component.scss
```diff
@@ -24,14 +27,16 @@
...
+      color: var(--color-action-primary);
+
       @include link-hover;
     }
+
+    &:hover agora-svg-icon {
+      path {
+        fill: var(--color-action-primary);
+      }
+    }
```
hallieswan, line 40: @sagely1 thinks this might be covered by the styles for the `agora-svg-icon`, so is going to double check whether this hover styling could be removed.

libs/agora/gene-comparison-tool/src/lib/components/gene-comparison-tool-filter-panel/gene-comparison-tool-filter-panel.component.ts
```diff
@@ -6,10 +6,18 @@ import { CheckboxModule } from 'primeng/checkbox';
...
   selector: 'agora-gene-comparison-tool-filter-panel',
-  imports: [CommonModule, FormsModule, TooltipModule, CheckboxModule, SvgIconComponent],
+  imports: [
+    CommonModule,
+    FormsModule,
+    TooltipModule,
+    ButtonModule,
+    CheckboxModule,
+    SvgIconComponent,
+  ],
```
hallieswan, line 20: We don't think this `ButtonModule` import is needed, so could be removed in the next pass.

libs/agora/styles/src/lib/_mixins.scss
```diff
@@ -53,13 +52,12 @@
...
   } @else {
-    @error 'No value found for `#{$breakpoint}`. '
-            + 'Please make sure it is defined in `$breakpoints` map.';
+    @error 'No value found for `#{$breakpoint}`. Please make sure it is defined in `$breakpoints` map.';
   }
 }
 
 @mixin link-hover {
-  color: var(--color-link);
+  // color: var(--color-link);
```
hallieswan, line 60: We think that this should be kept: 

```suggestion
  color: var(--color-link);
```

libs/agora/shared/src/lib/svg-icon/svg-icon.component.spec.ts
sagely1: moved!

libs/agora/gene-comparison-tool/src/lib/components/gene-comparison-tool-filter-list/gene-comparison-tool-filter-list.component.scss
sagely1: So this is what allows the mouseover on link OR icon to have the correct hover behavior

libs/agora/gene-comparison-tool/src/lib/components/gene-comparison-tool-filter-panel/gene-comparison-tool-filter-panel.component.ts
sagely1: removed

libs/agora/styles/src/lib/_mixins.scss
sagely1: this is usually overridden but I will leave it as a default

hallieswan (approved):
Looks good to me! The latest change to show the hover effect on the icon and the link text at the same time is great.

PR #3064, opened 2025-04-01
feat(iatlas): fixes for getting testing to work (DPE-1237)

andrewelamb:
[DPE-1237]

- Adds a command for pytest
- Adds a file for local testing env variables
- Fixes the file for env variables

[DPE-1237]: https://sagebionetworks.jira.com/browse/DPE-1237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tschaffter:
@andrewelamb The PR title needs to meet the requirement (see GitHub workflow error message).

Now I remember why I didn't add the "test" task: this task requires the DB to be accessible, which is not available when the "test" task is run by the CI/CD.

Different options:
- Update the (unit) tests to mock the responses expected from the DB. That way the DB does not need to be deployed.
- Turn the "test" task into an integration test task where the DB would be deployed. This could be done relatively easily by creating a dedicated GitHub workflow.
- Do not run the tests as part of the CI/CD workflow. This is not ideal but that may be acceptable given that the project may get shelved soon.

andrewelamb:
@tschaffter "Do not run the tests as part of the CI/CD workflow. This is not ideal but that may be acceptable given that the project may get shelved soon." What's the easiest way to do this?

tschaffter:
The easiest way is to remove the task "test" from `project.json`.

tschaffter:
Replaced by #3110

PR #3078, opened 2025-04-07
agora(feat): add Google Tag Manager

sagely1:


apps/agora/app/src/index.html
```diff
@@ -6,6 +6,18 @@
...
+    <script async src="https://www.googletagmanager.com/gtag/js?id=G-Y07DD946E4"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag() {
+        dataLayer.push(arguments);
+      }
+      gtag('js', new Date());
+
+      gtag('config', 'G-Y07DD946E4');
+    </script>
```
hallieswan, line 20: Agora is a SPA, so adding the tag to the index will only track the initial page load as a page view, but navigating to other routes will not be counted as page views. The [current Agora implementation](https://github.com/Sage-Bionetworks/Agora/blob/develop/src/app/app.module.ts#L49-L50) tracks views for each route using [`ngx-google-analytics`](https://www.npmjs.com/package/ngx-google-analytics), but this package is [not compatible with SSR](https://github.com/maxandriani/ngx-google-analytics?tab=readme-ov-file#setup-routing-module), so may not work for us anymore. 

`openchallenges` uses [`angular-google-tag-manager`](https://www.npmjs.com/package/angular-google-tag-manager) to add the tag and track route changes with its [`google-tag-manager` provider](https://github.com/Sage-Bionetworks/sage-monorepo/tree/main/apps/openchallenges/app/src/app/google-tag-manager). We may be able to follow this setup for Agora as well.

@JessterB -- do we want to continue tracking page views when navigating to individual routes within the app to part of the 4.0.0 release? If tracking the initial page is sufficient, then I can approve this PR as is!

JessterB: Yes, yes I do.

sagely1:
To be redone using GTM

PR #3080, opened 2025-04-08
feat(agora): update footer site version (AG-1140)

sagely1:
### Description
This updates the footer site version to show the Site Version

### Format
In localhost, we will see 'edge'
In dev, we use the image tag (sha) as the version
In stage and prod, we use the image tag but strip the -rcN suffix, and append the sha (via GitHub call)
In all environments, we don’t have trailing hyphens
If we fail to resolve a sha from GitHub, we can conditionally not display the hypen

Jira [link](https://sagebionetworks.jira.com/browse/AG-1140)


libs/agora/ui/src/lib/components/footer/footer.component.ts
```diff
@@ -62,9 +62,9 @@ export class FooterComponent implements OnInit {
   }
 
   getSiteVersion() {
-    return this.sha
-      ? `${this.configService.config.appVersion}-${this.sha}`
-      : this.configService.config.appVersion;
+    let appVersion = formatGitTag(this.configService.config.appVersion);
+    appVersion = appVersion.replace(/-rc\d+$/, '');
```
hallieswan, line 66: Can we remove the second line? It looks like `formatGitTag` handles replacing the rc suffix

```suggestion
    const appVersion = formatGitTag(this.configService.config.appVersion);
```

libs/agora/util/src/lib/app-helpers.ts
```diff
@@ -1,3 +1,8 @@
 export function removeParenthesis(s: string) {
   return s.replace('(', '').replace(')', '');
 }
+
+export function formatGitTag(appVersion: string) {
+  // remove the -rcX suffix
+  return appVersion.replace(/-rc\d+$/, '');
+}
```
hallieswan, line 8: `appVersion` maps to the [image tag](https://github.com/Sage-Bionetworks-IT/agora-infra-v3/blob/dev/app.py#L159-L165) (not the git tag), so what do you think about renaming this to `formatImageTag` or `formatAppVersion` for clarity?

libs/agora/ui/src/lib/components/footer/footer.component.ts
```diff
@@ -62,9 +62,9 @@ export class FooterComponent implements OnInit {
...
+    return this.sha ? `${appVersion}-${this.sha}` : `${appVersion}`;
```
hallieswan, line 67: Could we add tests for `getSiteVersion` for the expected inputs and output for dev, stage, and prod? I think that could help document the expected states for future developers. 

```
// dev
config.appVersion = 'sha-132a6d2'
config.tagName = 'agora/vsha-132a6d2'
expect(getSiteVersion()).toEqual('sha-132a6d2')

// stage, prod
config.appVersion = '4.0.0-rc3'
config.tagName = 'agora/v4.0.0-rc3'
expect(getSiteVersion()).toEqual('4.0.0-sha-132a6d2')
```

hallieswan (changes requested):
Looks great! Just a couple small tweaks

libs/agora/ui/src/lib/components/footer/footer.component.ts
sagely1: Ah yes, I had already refactored this.  Thanks!

libs/agora/util/src/lib/app-helpers.ts
sagely1: That's a better name.  I'll update it.

libs/agora/ui/src/lib/components/footer/footer.component.ts
sagely1: done

hallieswan (approved):
Looks great!

PR #3082, opened 2025-04-14
fix(agora): long urls resulting in 503 errors (AG-1754)

sagely1:
### Description
Long urls are causing errors when deployed.  

This PR updates:
- Sets a header limit (64kb) which should address the issue.

### DIFF
--- a/apps/agora/apex/Caddyfile
+++ b/apps/agora/apex/Caddyfile
@@ -1,4 +1,8 @@
 :80 {
+  limits {
+    header 64kb
+  }
+  
   handle_path /api-docs* {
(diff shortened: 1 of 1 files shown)


apps/agora/apex/Caddyfile
```diff
@@ -1,4 +1,8 @@
 :80 {
+  limits {
+    header 64kb
```
sagely1, line 3: updates Caddy request header limit to 64kb to handle more URL params

JessterB (approved):
🤞

sagely1:
Not needed

PR #3087, opened 2025-04-17
feat(openchallenges): add config to toggle the home page announcement banner (SMR-26)

rrchai:
closes https://sagebionetworks.jira.com/browse/SMR-26

## Description
Adding config (`showAnnouncement`) to toggle for the home page announcement banner as needed in the future.

## Changelog
- Add `showAnnouncement` config and disable the announcement banner on Home page

### DIFF
--- a/apps/openchallenges/app/docker-entrypoint.d/10-envsubst-on-app-config-template.sh
+++ b/apps/openchallenges/app/docker-entrypoint.d/10-envsubst-on-app-config-template.sh
@@ -2,4 +2,11 @@
 
 # Generate 'config.json' from 'config.json.template' using environment variables.
 cd "${APP_DIR}/browser/config"
-envsubst < config.json.template > config.json
+envsubst < config.json.template | jq '
+  with_entries(
+    if .value == "true" then .value = true
+    elif .value == "false" then .value = false
+    else .
+    end
+  )
+' > config.json
--- a/apps/openchallenges/app/src/config/config.json
+++ b/apps/openchallenges/app/src/config/config.json
@@ -6,5 +6,6 @@
   "environment": "development",
   "googleTagManagerId": "",
   "ssrApiUrl": "http://openchallenges-api-gateway:8082/api/v1",
-  "enableOperationFilter": false
+  "enableOperationFilter": false,
+  "showAnnouncement": false
 }
--- a/apps/openchallenges/app/src/config/config.json.template
+++ b/apps/openchallenges/app/src/config/config.json.template
@@ -6,5 +6,6 @@
   "environment": "${ENVIRONMENT}",
   "googleTagManagerId": "${GOOGLE_TAG_MANAGER_ID}",
   "ssrApiUrl": "${SSR_API_URL}",
-  "enableOperationFilter": "${ENABLE_OPERATION_FILTER}"
+  "enableOperationFilter": "${ENABLE_OPERATION_FILTER}",
+  "showAnnouncement": "${SHOW_ANNOUNCEMENT}"
 }
--- a/libs/openchallenges/config/src/lib/app.config.ts
+++ b/libs/openchallenges/config/src/lib/app.config.ts
@@ -9,7 +9,7 @@ export enum Environment {
 /* eslint-enable no-unused-vars */
 
 export interface AppConfig {
-  enableOperationFilter: boolean;
+  enableOperationFilter: string;
   appVersion: string;
   csrApiUrl: string;
   dataUpdatedOn: string;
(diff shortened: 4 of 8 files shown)


libs/openchallenges/home/src/lib/home.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="base mat-typography">
-  <div id="announcement">
+  <div *ngIf="showAnnouncement" id="announcement">
```
tschaffter, line 2: Should use the new control flow block `@if`.

apps/openchallenges/app/src/config/config.json
```diff
@@ -6,5 +6,6 @@
   "environment": "development",
   "googleTagManagerId": "",
   "ssrApiUrl": "http://openchallenges-api-gateway:8082/api/v1",
-  "enableOperationFilter": false
+  "enableOperationFilter": false,
+  "showAnnouncement": false
```
tschaffter, line 10: The template in the same folder should be updated along with the corresponding environment variable in `.env.example`.

tschaffter (changes requested):

tschaffter (commented):
LGTM! Have you tested the app when starting it with its development server and with Docker Compose?

libs/openchallenges/home/src/lib/home.component.ts
```diff
@@ -50,6 +51,7 @@ export class HomeComponent {
     this.privacyPolicyUrl = this.configService.config.privacyPolicyUrl;
     this.termsOfUseUrl = this.configService.config.termsOfUseUrl;
     this.apiDocsUrl = this.configService.config.apiDocsUrl;
+    this.showAnnouncement = this.configService.config.showAnnouncement === 'true';
```
rrchai, line 54: @tschaffter I was just testing if it works in docker, but I found the env variables we defined in the `.env/config.json` will result in string 'false', and in typescript, strings like "false" are truthy, like below:
```typescript
if ("false") {
  // it will be truthy and run
}
```
I just converted the config variable to boolean here for now, but feel free to let me know if there's a better solution.

tschaffter: Looks good. Feel free to request another review when ready.

libs/openchallenges/config/src/lib/app.config.ts
```diff
@@ -20,6 +20,7 @@ export interface AppConfig {
   ssrApiUrl: string;
   termsOfUseUrl: string;
   apiDocsUrl: string;
+  showAnnouncement: string;
```
tschaffter, line 23: Can this type be set to boolean and do the conversion type when the config file is read? It's actually not ideal if the conversion type / value check needs to be done in the Angular components.

tschaffter: If that can't be done easily, let's just remove the banner and the feature flag for the sake of time.

apps/openchallenges/app/docker-entrypoint.d/10-envsubst-on-app-config-template.sh
```diff
@@ -2,4 +2,11 @@
...
 # Generate 'config.json' from 'config.json.template' using environment variables.
 cd "${APP_DIR}/browser/config"
-envsubst < config.json.template > config.json
+envsubst < config.json.template | jq '
+  with_entries(
+    if .value == "true" then .value = true
+    elif .value == "false" then .value = false
+    else .
+    end
+  )
```
rrchai, line 11: @tschaffter the type conversion is moved to this script

apps/openchallenges/app/src/config/config.json.template
```diff
@@ -6,5 +6,6 @@
   "environment": "${ENVIRONMENT}",
   "googleTagManagerId": "${GOOGLE_TAG_MANAGER_ID}",
   "ssrApiUrl": "${SSR_API_URL}",
-  "enableOperationFilter": "${ENABLE_OPERATION_FILTER}"
+  "enableOperationFilter": "${ENABLE_OPERATION_FILTER}",
+  "showAnnouncement": "${SHOW_ANNOUNCEMENT}"
```
rrchai, line 10: TO-DO: test if removing double quote will resolve this issue

rrchai: Tested by using unquoted value e.g. `showAnnouncement: ${SHOW_ANNOUNCEMENT}` in the template, and the type is still string - see the testing console logs:
```
Raw showAnnouncement from config: false
typeof showAnnouncement: string
```

It looks like `envsubst` always performs string substitution, regardless of quoting

tschaffter (approved):

PR #3093, opened 2025-04-18
fix(agora): fix icon sizes (AG-1764)

sagely1:
### Description
Address icon issue with Nominated Targets.

### Changelog

The newly created icon component can now be configured to disable hover effects.

https://www.loom.com/share/aeff8af308a04bc6b3060aa0df1dcd1d

Now, this icon in the video does not change color when hovered over.
This icon is now the same size and aligned with the text.


libs/agora/nominated-targets/src/lib/nominated-targets/nominated-targets.component.scss
```diff
@@ -44,3 +44,14 @@
...
   }
 }
+
+.bottom-text {
+  display: flex;
+  align-items: center;
+  gap: 5px;
+
+  .bottom-text-icon {
+    color: var(--color-gray-600);
```
hallieswan, line 54: Could you pass the [icon color as an input](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/agora/shared/src/lib/svg-icon/svg-icon.component.ts#L19) to `agora-svg-icon`?

hallieswan (approved):
Nice! This is a great improvement, just one small question.

libs/agora/nominated-targets/src/lib/nominated-targets/nominated-targets.component.scss
hallieswan: Oh got it, so we preferred inheritance in this case so that we can use the color variable? Makes sense, thanks for explaining!

PR #3103, opened 2025-04-21
fix(agora): fix duplicate experimental validation filter values (AG-1758)

hallieswan:
## Description

We would like to remove the duplicate values displayed for the Experimental Validation filter in the GCT.

## Related issues

- [AG-1758](https://sagebionetworks.jira.com/browse/AG-1758)

## Changelog

- Update API to specify that `validation_study_details` is a nullable field
- Clean `validation_study_details` values before adding to set of available filters

## Validation

Confirmed that `validation_study_details` is a nullable field with this query:  

```
# data-version: 72
agora> db.geneinfo.countDocuments({  $or: [    { "target_nominations.validation_study_details": null },    { "target_nominations.validation_study_details": { $exists: false } }  ]})
35036
```

Experimental Validation filters have been updated: 

current dev | fix
:---:|:---:
experimental_validation_dev | experimental_validation_fix

And applying the single filter vs combination of filters that differ only in whitespace or capitalization results in the same number of results returned: 

https://github.com/user-attachments/assets/7e42a7ae-6297-4676-96dd-cbe88ff5e715

https://github.com/user-attachments/assets/d9fe1600-caeb-487f-a2da-6c259aa541cd

[AG-1758]: https://sagebionetworks.jira.com/browse/AG-1758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/agora/api/src/components/comparison.ts
+++ b/apps/agora/api/src/components/comparison.ts
@@ -102,8 +102,12 @@ function getComparisonGeneNominations(gene: Gene, teams: Team[]) {
     }
 
     // Validations
-    if (n.validation_study_details && !data.validations.includes(n.validation_study_details)) {
-      data.validations.push(n.validation_study_details.trim());
+    const validation_study_detail_clean = n.validation_study_details?.trim().toLowerCase();
+    if (
+      validation_study_detail_clean &&
+      !data.validations.includes(validation_study_detail_clean)
+    ) {
+      data.validations.push(validation_study_detail_clean);
     }
   });
 
--- a/apps/agora/api/src/components/comparison.ts
+++ b/apps/agora/api/src/components/comparison.ts
@@ -102,12 +102,11 @@ function getComparisonGeneNominations(gene: Gene, teams: Team[]) {
     }
 
     // Validations
-    const validation_study_detail_clean = n.validation_study_details?.trim().toLowerCase();
-    if (
-      validation_study_detail_clean &&
-      !data.validations.includes(validation_study_detail_clean)
-    ) {
-      data.validations.push(validation_study_detail_clean);
+    if (n.validation_study_details) {
+      const validation_study_detail_clean = n.validation_study_details.trim().toLowerCase();
+      if (!data.validations.includes(validation_study_detail_clean)) {
+        data.validations.push(validation_study_detail_clean);
+      }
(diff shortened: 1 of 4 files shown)


apps/agora/api/src/components/comparison.ts
```diff
@@ -102,8 +102,12 @@ function getComparisonGeneNominations(gene: Gene, teams: Team[]) {
...
 
     // Validations
-    if (n.validation_study_details && !data.validations.includes(n.validation_study_details)) {
-      data.validations.push(n.validation_study_details.trim());
+    const validation_study_detail_clean = n.validation_study_details?.trim().toLowerCase();
+    if (
+      validation_study_detail_clean &&
+      !data.validations.includes(validation_study_detail_clean)
+    ) {
+      data.validations.push(validation_study_detail_clean);
```
sagely1, line 110: Perhaps you could consider something along these lines:
Check input if defined, then perform necessary checks/formatting
```
if (n.validation_style_details) {
  if (!data.validations.includes(n.validation_study_details)) {
    const validation_study_details = n.validation_study_details?.trim().toLowerCase();
    data.validations.push(validation_study_details);
  }
}
```

sagely1 (changes requested):
I added some feedback (not tested) on code style.  Let me know if you have a different opinion.

apps/agora/api/src/components/comparison.ts
hallieswan: We can certainly check if the input is defined first! I'll make that change in the next commit.

We'll still need to `trim` and `toLowerCase` `validation_study_details` _before_ checking whether `data.validations` includes the value though, so that whitespace and capitalization differences don't cause us to add duplicate values to `data.validations`. For example, we need to ensure that `'not prioritized for experimental validation'` and `'Not prioritized for experimental validation '` are treated as the same value.

sagely1: Oh I see.  That makes sense.  We could consider creating a helper function for these string operations (with capitalization and without) in a future pass.

sagely1 (approved):
LGTM!

PR #3106, opened 2025-04-22
fix(agora): dialog, misc styling issues (AG-1751)

sagely1:
### Description
This addresses styling issues in dialogs and icons.

GCT overlay header:
Before:
image

After:
image

GCT overlay icon
Before:
image

After:
image

Legend
Before (Production Agora):
This addresses an existing bug in production with the "Color" heading missing:
image

After:
image

Nominated Targets "Learn more about Nominated Targets" link
Before:
image

After:
image

Gene Details - Evidence - RNA: Data and Methods for this study link
Before:
image

After:
image

Gene Details - Evidence - RNA: similar link
Before:
image

After:
image


hallieswan (changes requested):
The updates to the `popover-link` dialog and `gene-similar` link icon look great! 

For the `p-dialog` styling, it looks like we're hiding the PrimeNG close button and adding a custom close button in the header of each dialog, then applying similar styling in each component. Instead, could we style the PrimeNG close button and header globally or in a shared place? That way, the dialogs will stay consistent going forward and we can avoid duplicating code. 

I think we could use the existing `_dialog.scss` if we want (and tackle moving the styles into the preset in the future) -- we would just need to update class names that were changed in the PrimeNG update (`.p-dialog-header-icons` --> `.p-dialog-header-actions`, `.p-dialog-header-close-icon` --> `.p-dialog-close-button`) and tweak a couple global styles in the preset (`dialog.header.padding: '0'`, `dialog.root.borderRadius: '0'`).

sagely1:
Good point.  I wanted to duplicate the existing visual behavior so I added the ability to customize the dialog per component but this makes me feel like the right way forward is to stick with a consistent styling.  Since I wanted to clean up the shared styles, I only brought back what was actually needed and reverted PrimeNG preset changes.  Now dialogs will be using a PrimeNG button (in the header) and not the custom SVGIconComponent so any styling or customization will have to occur in modal-link component going forward.  Another side-effect is that PrimeNG buttons have a slightly different hover behavior. 
 This means buttons in the dialogs will now have a circular background around the icon which is more visually apparent the icon is clickable.  I'm personally ok with this behavior.  
image

JessterB:
@hallieswan This actually fixes https://sagebionetworks.jira.com/browse/AG-1752, so yes please!

PR #3111, opened 2025-04-23
fix(agora): update footer logic (AG-1775)

sagely1:
### Description
Update footer logic to use Angular 16+ methods for handling subscriptions.

Code involves the use of:
`.pipe(takeUntilDestroyed())`
This ensures that there won't be memory leaks.


hallieswan (approved):
Looks good!

PR #3114, opened 2025-04-25
feat(model-ad): add PrimeNG and app skeleton (MG-166)

sagely1:
### Description
This sets up Model-AD to use PrimeNG and the basic application structure of Agora

- Add PrimeNG functionality ([PR](https://sagebionetworks.jira.com/jira/software/c/projects/MG/boards/243?selectedIssue=MG-166)) and use preset file from Agora
- Add application skeleton (header - outlet - footer)
- Add home component
- Add PrimeNG button (to home component) to verify

Doesn't look like much now:
image

Note: Styles/colors specific to Model-AD will not be implemented in this PR.  This is just a copy of Agora's preset file.


apps/model-ad/app/src/app/app.component.html
```diff
@@ -1,3 +1,7 @@
-<div class="container">
-  <router-outlet></router-outlet>
+<div id="container">
+  <model-ad-header></model-ad-header>
+  <div class="container">
+    <router-outlet></router-outlet>
+  </div>
+  <!-- <model-ad-footer></model-ad-footer> -->
```
sagely1, line 6: We can reenable this after we rewrite the footer

apps/model-ad/app/src/app/app.config.ts
```diff
@@ -7,11 +7,11 @@ import {
 import { withInterceptorsFromDi, provideHttpClient } from '@angular/common/http';
 import { provideAnimationsAsync } from '@angular/platform-browser/animations/async';
 import { providePrimeNG } from 'primeng/config';
-import Lara from '@primeng/themes/lara';
 import { BASE_PATH as API_CLIENT_BASE_PATH } from '@sagebionetworks/model-ad/api-client-angular';
 import { configFactory, ConfigService } from '@sagebionetworks/model-ad/config';
 
 import { routes } from './app.routes';
+import { MyPreset } from './myPrimeNGPreset';
```
hallieswan, line 14: From the designs, Model-AD and Agora's styles look very similar. Do you think it would be worth creating a shared preset that we can use to keep base styles consistent between the two applications? Then each application could override the preset with application-specific styles, as needed.

libs/model-ad/home/src/lib/home.component.spec.ts.off
hallieswan, line 1: Would it be possible to turn this test on? I know the current Home component is a placeholder, but it would be nice for the test file to be set up so we can add meaningful unit tests when we start implementing the real component.

libs/model-ad/ui/src/_lib-theme.scss
hallieswan, line 1: I think this file was [imported](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/model-ad/themes/src/_index.scss#L2) in `libs/model-ad/themes/src/_index.scss` -- can we update that file to reflect that this `lib-theme` has been deleted?

libs/model-ad/ui/src/lib/header/header.component.spec.ts.off
hallieswan, line 1: Would it be possible to turn this test on as well?

hallieswan (changes requested):

tschaffter:
@sagely1 @hallieswan I can see this PR—and the ones that follow—becoming a valuable reference for building new Angular apps in the monorepo. I already have a couple of exploratory apps in mind that could benefit from this foundation, including a standalone chatbot to test an upcoming MCP-server that @rrchai will be developing. Thanks in advance for this PR! Including comments and basic tests will go a long way toward strengthening the robustness of future apps.

libs/model-ad/ui/src/_lib-theme.scss
sagely1: done

apps/model-ad/app/src/app/app.config.ts
sagely1: I see your point.  The designs currently look similar but they will do have some minor differences at the moment.  Most of the changes will most likely occur in shared styles. Since they won't be carbon copies of each other and we don't know what design changes are to come, we should probably allow for individual styling at this point.

libs/model-ad/ui/src/_lib-theme.scss
sagely1: done

libs/model-ad/ui/src/lib/header/header.component.spec.ts.off
sagely1: done

libs/model-ad/home/src/lib/home.component.spec.ts.off
sagely1: done

hallieswan (approved):
Looks great!

tschaffter (approved):
LGTM!

PR #3123, opened 2025-05-01
feat(model-ad): adding shared fonts, favicons, styles and svg components (MG-181) (MG-184)

sagely1:
### Description
This is foundational work before header, footer and home components can be developed.

JIRA: [MG-181](https://sagebionetworks.jira.com/browse/MG-181)
Also addresses: [MG-184](https://sagebionetworks.jira.com/browse/MG-184)

### Changelog
This PR adds ModelAD-specific assets:
- favicon
- header logo
- shared styles
- updates page title

Shared components (to be shared among the explorer apps: Agora and Model-AD) include:
- models
- services
- svg-icon and svg-image components

Package-based font using [FontSource](https://github.com/fontsource/fontsource)
The word 'Agora' was put on the Home page to visually show that the DM Sans font works.  The 'g' in Agora has a unique look when DM Sans is used.
image

### Notes
A follow-up PR will update Agora to leverage the shared code.

[MG-181]: https://sagebionetworks.jira.com/browse/MG-181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MG-184]: https://sagebionetworks.jira.com/browse/MG-184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3128, opened 2025-05-02
chore(explorers): cleanup explorers imports, tests, and styles

hallieswan:
## Description

#3123 added the `explorers` shared components library. This PR cleans up some imports that were still pointing at `agora` resources and adds `testing` and `styles` libraries so `explorers` components are independent of `agora`.


libs/explorers/services/src/index.ts
```diff
@@ -1 +1,2 @@
 export * from './lib/svg-icon.service';
+export * from './lib/synapse-api.service';
```
hallieswan, line 2: Missing export

tsconfig.base.json
```diff
@@ -44,6 +44,8 @@
       "@sagebionetworks/explorers/models": ["libs/explorers/models/index.ts"],
       "@sagebionetworks/explorers/services": ["libs/explorers/services/src/index.ts"],
       "@sagebionetworks/explorers/shared": ["libs/explorers/shared/src/index.ts"],
+      "@sagebionetworks/explorers/styles": ["libs/explorers/styles/src/index.ts"],
```
tschaffter, line 47: Is this one needed? OpenChallenges works without it.

hallieswan: Oh good call, I'll remove it.

sagely1:
We should only add styles in as we need them.  My recommendation is just add the styles necessary.  This will reduce our technical burden and identify old styles no longer needed easier.  My recommendation is to add only what is needed to variables.scss and mixins.scss as components are developed.

sagely1 (changes requested):
See comment about adding shared styles.

tschaffter (approved):

PR #3129, opened 2025-05-03
feat(agora): restructure Agora OpenAPI description to support microservice development (SMR-92)

tschaffter:
## Description

This PR restructures Agora OpenAPI description to enable language-independent backend contribution using microservices.

Even if the proposed `gene-api` service is not ultimately adopted, the restructured API description will facilitate the exploration of new backend and frontend features, and support the app's incremental modernization.

A preview of this approach and what the next step will look like is available in the PR #3130, which will be recreated once this PR is approved and merged. See [SMR-91](https://sagebionetworks.jira.com/browse/SMR-91) for more information on the motivation and roadmap of this initiative.

## Related Issues

- Closes https://sagebionetworks.jira.com/browse/SMR-92

## Changelog

- Restructure Agora OpenAPI description to support microservice development.
- Create an API description for the draft `gene-api` service (single healthcheck endpoint for demo).
- Rename the API description output folder from `build/` to `openapi/` to avoid conflicts with the workspace nuke script.

### Development Workflow

The development workflow of the API description and downstream projects is unchanged:

```console
nx build agora-api-description
nx lint agora-api-description
nx build agora-api-docs
nx run agora-api-client-angular:generate
```

### API Description File Changes

Instead of using a single `openapi.yaml` file, the API specification has been modularized into multiple files, each serving a distinct purpose:

- `openapi/openapi.yaml`: The complete, aggregated API specification. This file is used to generate the API documentation site and API clients (e.g., for Angular), and serves as a comprehensive reference. Note: The proposed `gene-api` service whose API description is listed below is currently not aggregated.

- `openapi/api.openapi.yaml`: The specification for the main API service, corresponding to the existing Agora API. This file is used to generate the primary backend service. Note: Agora is currently not generating a API stub using the OpenAPI Generator.

- `openapi/gene-api.openapi.yaml`: A separate specification for a secondary API service designed to extend the main API as part of [SMR-91](https://sagebionetworks.jira.com/browse/SMR-91). This file is used to generate the `gene-api` service, which supports the upcoming, experimental Gene Comparison Tool functionality.

> [!NOTE]  
> I hesitated to include "api" in the name of the service API descriptions, e.g. `{src,openapi}/gene-api.openapi.yaml` and `{src,openapi}/api.openapi.yaml`. The main motivation is to have an explicit mapping from the API description `openapi/api.openapi.yaml` to the agora-`api` service, and from `openapi/gene-api.openapi.yaml` to the agora-`gene-api` service.

[SMR-91]: https://sagebionetworks.jira.com/browse/SMR-91?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/agora/api-description/openapi-merge.yaml
```diff
@@ -0,0 +1,7 @@
+inputs:
+  # info is defined in base OA definition. This file must be listed first.
+  - inputFile: openapi/base.openapi.yaml
+  # list files in alphanumeric order
+  - inputFile: openapi/api.openapi.yaml
+  # - inputFile: openapi/gene-api.openapi.yaml
```
hallieswan, line 6: Should this be uncommented or are you planning to include the `gene-api` in the complete `openapi` specification in the future?

libs/agora/api-description/project.json
```diff
@@ -4,13 +4,30 @@
...
         "cwd": "{projectRoot}"
       }
     },
+    "build": {
+      "executor": "nx:run-commands",
+      "options": {
+        "commands": [
+          "openapi-merge-cli --config openapi-merge.yaml",
+          "rm openapi/base.openapi.yaml"
+        ],
```
hallieswan, line 25: Should we add a similar command to remove `openapi/base.openapi.yaml` after running `agora-api-client-angular:generate`? When I run `nx run agora-api-client-angular:generate`, `base.openapi.yaml` is added to `libs/agora/api-description/openapi`.

tschaffter: So I tried to get rid of `openapi/base.openapi.yaml` for OC last week but like you, I observed that the file was restored after running tasks like generating the API client with OpenAPI Generator. Maybe I missed something but I didn't understood why the file was being restored. I assumed it was a problem with the Nx cache. Removing the Nx cache should have solve that issue but didn't. Around that time, I realized that there was somehow some value in tracking the file `openapi/base.openapi.yaml` since it's a generated/hydrated file and the developer may be interested in looking inside it, even though its content can also be found in the aggregated API description.

If you followed me until here, I would suggest that I edit this PR to remove the command `rm openapi/base.openapi.yaml` for the reason mentioned above.

Now that I think about it, we could also use `.gitignore` to keep the file but not track it on GitHub. I'm not sure whether the slight increase in project complexity (addition of `.gitignore`) is worth it.

libs/agora/api-description/openapi-merge.yaml
hallieswan: Got it, makes sense. Thanks for the explanation!

libs/agora/api-description/project.json
hallieswan: Sounds good to remove the `rm openapi/base.openapi.yaml` command. Would we also start tracking the file in git? I would be fine with that. If we want to avoid tracking the file, then I would lean towards adding the `.gitignore`, so it's clear to future developers that we are aware of this file and that it's expected that it is untracked.

tschaffter: I just pushed a commit that restores `openapi/base.openapi.yaml`.

hallieswan (approved):

PR #3139, opened 2025-05-07
chore(synapse): upgrade Angular API clients for Synapse to OpenAPI Generator 7.11.0 (SMR-103)

tschaffter:
## Description

Update the Synapse API client for Angular generated with OpenAPI Generator. This update mirrors the update recently applied to OC API client for Angular.

> [!NOTE]
This Angular API client for Synapse is generated from an OpenAPI specification in JSON format, unlike most other projects in the monorepo, which use the YAML format.

## Related Issues

- Closes https://sagebionetworks.jira.com/browse/SMR-103

## Changelog

- Update the Synapse API client for Angular project files based off of a freshly generated Angular library.
- Convert the library to a non-buildable one to simplify the development environment.
- Update OpenGenerator from v6.1.0 to v7.11.0.
- Rename the output folder of Synapse API description from `build` to `openapi` to prevent conflict with the workspace nuke script.


tschaffter:
@hallieswan There are Agora tests that are failing after updating the Synapse API client for Angular. Could you please review them and submit a fix for them? I could also use this opportunity to update the Synapse API description, which has not been updated in a while. If you want, I could update the API description and regenerate the API client before you attempt to fix the issues. Alternatively, it may be wiser to keep the scope of this PR to updating OpenAPI Generator only and then update the Synapse API description and client for Angular in a separate PR.

hallieswan:
@tschaffter Agreed that it makes sense to focus on the Synapse API description for this PR. As for the failing tests, it looks like `OrgSagebionetworksRepoModelWikiWikiPage` and `WikiPageServicesService` were removed as part of this PR, which we import in [`synapse-api.service.ts`](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/agora/services/src/lib/synapse-api.service.ts#L9-L12). It's likely that this is what is causing the test failures (and the [build failure in CI](https://github.com/Sage-Bionetworks/sage-monorepo/actions/runs/14875121545/job/41770890561?pr=3139)). Why were these services removed?

tschaffter:
> Agreed that it makes sense to focus on the Synapse API description for this PR. 

@hallieswan To clarify, this PR currently focus on updating the OpenAPI Generator from v6 to v7. The Synapse API description included in `synapse-api-description` has not been changed.

My bad! It looks like the PR is not capturing the new files generated by the generator. I will updating shortly.

tschaffter:
@hallieswan See the note added to the description of this PR. I've built and started Agora with Docker Compose and Agora web app looks fine to me.

hallieswan (approved):
I built/started Agora and the News page (which pulls a Wiki using a service provided by the Synapse Angular API client) worked as expected. Looks good to me!

PR #3141, opened 2025-05-07
fix(synapse): fix synapse-api-client-angular build

hallieswan:
The paths in `synapse/api-client-angular` incorrectly point at the `openchallenges` project, so the build is failing since #3135 removed `ng-package.json` from `openchallenges`. This PR corrects the `synapse/api-client-angular` paths.


tschaffter (approved):
Thanks for catching that! This is currently fixed in #3139 but merging this PR first will help with upcoming PR. Thanks again!

PR #3157, opened 2025-05-11
test(openchallenges): configure unit tests for OC MCP server and add a test example (SMR-119)

tschaffter:
## Description

Configure the Gradle `test` task of the OC MCP server.

I initially considered configuring the `openchallenges-mcp-server:test` task to run `./gradlew nativeTest`. This task would test the native image (binary) that will be executed in production. One drawback is that the Mockito library is not well supported when building the tests with GraalVM.

Instead, I decided to run the unit tests with the JVM because its much faster, provides the best DX with excellent IDE support and to benefit from excellent testing tools like Mockito.

In addition, integration tests could be developed in the future as scripts that verify the behavior of the native image.

## Related Issues

- Contributes to https://sagebionetworks.jira.com/browse/SMR-119

## References

- https://java-design-patterns.com/patterns/arrange-act-assert/
- https://melix.github.io/blog/2022/05/gradle-laziness.html


apps/openchallenges/mcp-server/build.gradle.kts
```diff
@@ -29,6 +29,15 @@ graalvmNative {
       buildArgs.add("--libc=musl")
       buildArgs.add("--no-fallback")
     }
+    named("test") {
```
tschaffter, line 32: Enable the task `nativeTest`, which will not be used as part of the normal development workflow.

rrchai (approved):
LGTM

PR #3158, opened 2025-05-12
feat(openchallenges): expose OC MCP server through the reverse proxy (SMR-107) 

rrchai:
## Related Issue
- https://sagebionetworks.jira.com/browse/SMR-107

## Changelog
- Expose `openchallenges-mcp-server` in the reverse proxy, so it can be accessed via [openchallenges.io/mcp-server](http://openchallenges.io/mcp-server)
- Add `mcp.json` to configure MCP servers for VSCodes Copilot

### MCP Inspector
Screen Shot 2025-05-10 at 1 22 56 PM

### Copilot

Screen Shot 2025-05-12 at 11 01 19 AM


rrchai:
## ❓ Problem 
When running the MCP Inspector (or other MCP-compatible client, like Copilot), the backend Spring AI MCP server by default emits the following endpoint configuration:
```yaml
sse-endpoint: /sse
sse-message-endpoint: /mcp/message
```
It can be confirmed by running, and the data path in the response uses `/mcp/message`:

```console
curl -i http://localhost:8887/sse
event:endpoint
data:/mcp/message?sessionId=...
```
However, if only a single reverse proxy is set up in Caddy of `apex` like this:

```console
handle_path /mcp-server* {
  reverse_proxy {env.MCP_SERVER_HOST}:{env.MCP_SERVER_PORT}
}
```
Then, this results in the ` (HTTP 404)` error during the connection with the MCP Server.  This is because the client, like MCP Inspector sends to `http://localhost:8000/mcp/message`
as the POST endpoint, but `apex` doesn't handle `/mcp/message`, resulting in a 404.

## ✅ Solutions
### Option 1: Dual proxy rules
```console
handle_path /mcp-server* {
  reverse_proxy ...
}

handle /mcp* {
  reverse_proxy ...
}
```
SSE request: `/mcp-server/sse` → `/sse` ✅ (due to handle_path /mcp-server*`)
POST request: `/mcp/message` → preserved as `/mcp/message` ✅ (due to `handle /mcp*`)

This works, but we prefer not to maintain two separate proxy routes.

### Option 2: Adjust app endpoints to align with proxy prefix
We can change [Spring AI MCP Server configs](https://docs.spring.io/spring-ai/reference/api/mcp/mcp-server-boot-starter-docs.html#_configuration_properties).
@tschaffter - when we looked at the `sse-endpoint` yesterday, I realized it’s only supported as of the [M8 release](https://spring.io/blog/2025/04/30/spring-ai-1-0-0-m8-released). Thanks for updating to M8!

```yaml
sse-endpoint: /mcp-server/sse
sse-message-endpoint: /mcp-server/message
```
And used a single reverse proxy rule:
```console
handle /mcp-server* {
  reverse_proxy {env.MCP_SERVER_HOST}:{env.MCP_SERVER_PORT}
}
```
SSE request: /`mcp-server/sse` → backend receives `/mcp-server/sse`
Post request: `/mcp-server/message` → backend receives `/mcp-server/message`

> [!NOTE]
> Not all combinations between configs and proxy will work. For example, if we use 

| Spring Config  | Proxy | Result | Reason |
| ------------- | ------------- | ------------- | ------------- |
| sse: `/sse`, message: `/message` | `handle_path /mcp-server*`  | sse works, but message will fail  | the post request will send to `localhost:8000/message` and `/message` alone is not handled in proxy|

tschaffter (approved):
The MCP server works well as described. I was able to ask two questions via GitHub Copilot. Thanks!

PR #3180, opened 2025-05-16
feat(openchallenges): create project-specific config for the OC MCP server (SMR-112)

rrchai:
## Related Issue

closes https://sagebionetworks.jira.com/browse/SMR-112

## Changelog
- Config API base url via the environment variable for `openchallenges-mcp-server`
- Add a welcome message for `openchallenges-mcp-server`

## Preview

I tested three configuration binding approaches below in the `application.yml`, while keeping `.env` file with snake-cased variables (e.g., `WELCOME_MESSAGE`):

- Dotted case - [9dc4990](https://github.com/Sage-Bionetworks/sage-monorepo/pull/3180/commits/9dc499001afcdf071c7e2373fad60c33906914be)
- Kebab case - [11032ad](https://github.com/Sage-Bionetworks/sage-monorepo/pull/3180/commits/11032ad7aea1d7a9cb7c780f1c1b6d717704f07d)
- Camel case - [2d9050f](https://github.com/Sage-Bionetworks/sage-monorepo/pull/3180/commits/2d9050fe30cf0f3838ae4c1ad00a23a1e5449b5e)

Selected approach is ~~**Camel case**~~ **Kebab case**.

### Running in Dev Mode
```
cd apps/openchallenges/mcp-server && ./gradlew clean bootRun
```

You should see the following welcome message in the terminal logs:

```console
o.s.o.mcp.server.McpServerApplication    : Welcome to the OpenChallenges MCP Server - dev mode
```

### Running as Container

> [!NOTE]
> Make sure to run `workspace-install` or copy the contents of `.env_example` into a `.env` file in the `openchallenges-mcp-server` project for the first run.

```
cd /workspaces/sage-monorepo && \
  nx build-openchallenges-mcp-server \
  nx serve-detach openchallenges-mcp-server \
  docker logs openchallenges-mcp-server
```

You should see the following welcome message in the container logs:

```console
o.s.o.mcp.server.McpServerApplication    : Welcome to the OpenChallenges MCP Server
```

### DIFF
--- a/apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/McpServerApplication.java
+++ b/apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/McpServerApplication.java
@@ -29,7 +29,7 @@ public class McpServerApplication implements CommandLineRunner {
 
   @Override
   public void run(String... args) throws Exception {
-    logger.info(mcpServerConfigData.getApiBaseUrl());
+    logger.info(mcpServerConfigData.getWelcomeMessage());
   }
 
   @Bean
--- a/apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/McpServerApplication.java
+++ b/apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/McpServerApplication.java
@@ -29,7 +29,7 @@ public class McpServerApplication implements CommandLineRunner {
 
   @Override
   public void run(String... args) throws Exception {
-    logger.info(mcpServerConfigData.getWelcomeMessage());
+    logger.info(mcpServerConfigData.getWelcome().getMessage());
   }
 
   @Bean
--- a/apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/McpServerApplication.java
+++ b/apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/McpServerApplication.java
@@ -2,17 +2,34 @@ package org.sagebionetworks.openchallenges.mcp.server;
 
 import java.util.Arrays;
 import java.util.List;
+import org.sagebionetworks.openchallenges.mcp.server.config.McpServerConfigData;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.ai.tool.ToolCallback;
 import org.springframework.ai.tool.ToolCallbacks;
+import org.springframework.boot.CommandLineRunner;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.context.annotation.Bean;
 
 @SpringBootApplication
-public class McpServerApplication {
+public class McpServerApplication implements CommandLineRunner {
+
+  private static final Logger logger = LoggerFactory.getLogger(McpServerApplication.class);
+
+  private final McpServerConfigData mcpServerConfigData;
+
+  public McpServerApplication(McpServerConfigData mcpServerConfigData) {
+    this.mcpServerConfigData = mcpServerConfigData;
+  }
 
   public static void main(String[] args) {
     SpringApplication.run(McpServerApplication.class, args);
+  }
+
+  @Override
+  public void run(String... args) throws Exception {
+    logger.info(mcpServerConfigData.getApiBaseUrl());
   }
 
   @Bean
--- a/apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/McpServerApplication.java
+++ b/apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/McpServerApplication.java
@@ -29,7 +29,7 @@ public class McpServerApplication implements CommandLineRunner {
 
   @Override
   public void run(String... args) throws Exception {
-    logger.info(mcpServerConfigData.getWelcome().getMessage());
+    logger.info(mcpServerConfigData.getWelcomeMessage());
   }
 
   @Bean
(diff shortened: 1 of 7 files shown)


tschaffter:
@rrchai This ticket does not close [SMR-70](https://sagebionetworks.jira.com/browse/SMR-70).

tschaffter (changes requested):
While this approach works, the ticket requested the creation of a `McpServerConfigData` class based on the example provided. The motivation is to centralize all the project-specific config properties in one Java object.

apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/McpServerApplication.java
```diff
@@ -29,7 +29,7 @@ public static void main(String[] args) {
 
   @Override
   public void run(String... args) throws Exception {
-    logger.info(mcpServerConfigData.getWelcomeMessage());
+    logger.info(mcpServerConfigData.getWelcome().getMessage());
```
tschaffter: Welcome message is one indivisible entity.

apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/config/McpServerConfigData.java
```diff
@@ -6,9 +6,27 @@
...
+  @Data
+  public static class Api {
+
+    private Base base;
+
+    @Data
+    public static class Base {
+
+      private String url;
+    }
```
tschaffter: Base url is one indivisible entity.

apps/openchallenges/mcp-server/src/main/resources/application.yml
```diff
@@ -1,6 +1,5 @@
-openchallenges-mcp-server:
-  welcomeMessage: 'Welcome to the OpenChallenges MCP Server - dev mode'
-  apiBaseUrl: http://openchallenges-api-gateway:8082/api/v1
+welcome.message: 'Welcome to the OpenChallenges MCP Server - dev mode'
+api.base.url: http://openchallenges-api-gateway:8082/api/v1
```
tschaffter: These two properties should be grouped like done previously.
Keep the properties organized as the other properties in this file.

Preferred:

```yaml
openchallenges:
    mcp-server:
      welcome-message: ''
      api-base-url: ''
```

apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/config/McpServerConfigData.java
```diff
@@ -6,9 +6,27 @@
 
 @Data
 @Configuration
-@ConfigurationProperties(prefix = "openchallenges-mcp-server")
+@ConfigurationProperties
```
tschaffter, line 9: Add prefix (see comments below)

apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/config/McpServerConfigData.java
```diff
@@ -6,9 +6,27 @@
...
+  public static class Api {
+
+    private Base base;
+
+    @Data
+    public static class Base {
+
+      private String url;
+    }
+  }
```
tschaffter: This is too much code to maintain. Instead, we should rely on `@Data` more as done previously. For now keep two private properties at the same level as there is no need to nest properties at this time.

tschaffter (changes requested):

apps/openchallenges/mcp-server/src/main/resources/application.yml
rrchai: @tschaffter  this is not the latest codebase, current approach is using camel-cased.

apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/config/McpServerConfigData.java
rrchai: FYI, if we add the prefix, we need to add `OPENCHALLENGES-MCP-SERVER` to the `.env` as well, or we use the Kebab case - [dea4aee](https://github.com/Sage-Bionetworks/sage-monorepo/pull/3180/commits/dea4aee15d2c706c7d4270717e2615ab9f2d2c78)

apps/openchallenges/mcp-server/src/main/resources/application.yml
tschaffter: If I remember well, both camel and kebab case can be used in `application.yml`. I suggest to use kebab case because we use kebab case for other (Spring) properties in this file. Currently we don't a convention across projects in the monorepo, it would be nice to have one in the future. For now, we just need to be consistent within each `application.yml` file.

apps/openchallenges/mcp-server/src/main/java/org/sagebionetworks/openchallenges/mcp/server/config/McpServerConfigData.java
tschaffter: Good point! It's a bit verbose but this approach has the benefit to be descriptive when settings env vars for different dependencies (OPENCHALLENGES_MCP_SERVER_XXX, SPRING_DATA_YYY). Let's use the prefix because that's what other dependencies do like Spring.

apps/openchallenges/mcp-server/src/main/resources/application.yml
rrchai: sure, I can update to have consistent casing with other java projects.

apps/openchallenges/mcp-server/src/main/resources/application.yml
```diff
@@ -1,3 +1,7 @@
+openchallenges-mcp-server:
```
rrchai, line 1: @tschaffter to keep config data simplify, I didn't use nest property for this line.

tschaffter (approved):
The PR looks great!

PR #3187, opened 2025-05-19
feat(model-ad): 404 page (MG-90)

sagely1:
### Description
This PR creates a not-found page to match the Figma design: https://sagebionetworks.jira.com/jira/software/c/projects/MG/boards/243?selectedIssue=MG-90#:~:text=https%3A//www.figma.com/design/GtUYqZBxoOnPtjV1hLcTgw/Model%2DAD%2DExplorer%2DDesigns%3Fnode%2Did%3D273%2D6301%26t%3DQSHi57T2sYx2Wyfa%2D1

This PR also simplifies CSS in the 404 page, header and TOS pages to be mobile-friendly and responsive.

image

https://www.loom.com/share/0f1062db7fe143b08351a547556c470e
Press play to see the full video.

## Changelog
A 404 page (NotFoundComponent) was created in the explorers folder so Agora and Model-AD can reuse the same component.  It has the ability to specify an email to be used for the 404 page.


libs/explorers/ui/src/lib/components/header/header.component.scss
```diff
@@ -2,6 +2,11 @@
 
 #header {
   padding: 10px 80px;
+
+  @media (width <= 965px) {
+    padding: 10px 30px 10px 10px;
```
sagely1, line 7: ensures the proper padding when hamburger menu is displayed

libs/model-ad/util/src/lib/constants.ts
```diff
@@ -1,4 +1,4 @@
 export const HELP_URL =
   'https://help.adknowledgeportal.org/apd/Model-AD+Explorer+Resources.4077682781.html';
 
-export const MIN_DESKTOP_WIDTH = 1320;
+export const MIN_DESKTOP_WIDTH = 768;
```
sagely1, line 4: based on typical tablet resolutions, I think this value makes more sense

libs/explorers/shared/src/lib/components/not-found/not-found.component.ts
```diff
@@ -0,0 +1,18 @@
...
+  templateUrl: './not-found.component.html',
+  styleUrls: ['./not-found.component.scss'],
+})
+export class NotFoundComponent {
+  configService = inject(ConfigService);
+  backgroundImagePath = 'explorers-assets/images/background.svg';
+  email = '';
+
+  constructor() {
+    this.email = this.configService.config.supportEmail;
```
hallieswan, line 16: Do we expect the `supportEmail` to be different when running the app locally vs deploying the app? If not, I would be inclined to use an `input` rather than injecting the `ConfigService`, so that the support email is set for all environments in this repo, rather than managing the support email for the deployed stack in the infra repo.

sagely1: We can't use input without a wrapper component.  I'll use routing params and define the supportEmail in the  constants file.

libs/explorers/shared/src/lib/components/not-found/not-found.component.scss
```diff
@@ -0,0 +1,30 @@
...
+  min-height: calc(100vh - var(--header-height) - var(--footer-height) + 1px);
+  color: white;
+  background-size: cover !important;
+
+  // placeholder color until background image is loaded
+  background-color: #78789a;
+  display: flex;
+  align-items: start;
+  flex-direction: column;
+  padding: clamp(50px, 5vw, 150px);
```
hallieswan, line 13: The right / top padding in your PR screenshot looks different than the design -- should this padding be adjusted?

| pr description | design |
:--:|:--:
image | image

libs/explorers/shared/src/lib/components/not-found/not-found.component.html
```diff
@@ -0,0 +1,12 @@
+<div class="page-not-found" [style.background-image]="'url(' + backgroundImagePath + ')'">
+  <div>
+    <h1 class="page-not-found-heading">
+      This page isn't available<br />
+      right now.
+    </h1>
+    <p>The link you followed may be broken, or the page may have been removed.</p>
```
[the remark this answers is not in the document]
sagely1: done!

libs/explorers/shared/src/lib/components/not-found/not-found.component.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+      'href',
+      'mailto:support@example.com',
+    );
+  });
+
+  it('should display the background image', async () => {
+    const component = await setup();
+    const expectedPath = 'explorers-assets/images/background.svg';
+    expect(component.backgroundImagePath).toBe(expectedPath);
+  });
```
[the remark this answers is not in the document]
sagely1: This seems like a weak test.  I'll remove it.

libs/explorers/shared/src/lib/components/not-found/not-found.component.scss
sagely1: You're right.  I'll make the page scale with more padding at larger sizes.

libs/explorers/shared/src/lib/components/not-found/not-found.component.scss
```diff
@@ -0,0 +1,31 @@
+@use 'libs/explorers/styles/src/lib/mixins';
+
+.page-not-found {
+  min-height: calc(100vh - var(--header-height) - var(--footer-height) + 1px);
```
hallieswan, line 4: We use the same calculation in `terms-of-service.component.html` -- do you think this value would be useful to store in a variable, e.g. maybe something like `--content-min-height`?

libs/explorers/shared/src/lib/components/not-found/not-found.component.scss
```diff
@@ -0,0 +1,31 @@
...
+  color: white;
+  background-size: cover !important;
+
+  // placeholder color until background image is loaded
+  background-color: #78789a;
```
hallieswan, line 9: This is nice to have on a slower connection, but results in a flash of color with a faster connection. Up to you if you'd like to keep it, just something to consider!

https://github.com/user-attachments/assets/af86316c-6951-4e1f-8a69-dc78e56b7b1f

libs/explorers/shared/src/lib/components/terms-of-service/terms-of-service.component.html
```diff
@@ -1,13 +1,17 @@
...
-<div class="container-sm terms-of-service">
-  @if (loading) {
-    <div class="loading-icon">
-      <explorers-loading-icon />
+  <div class="terms-of-service">
+    @if (loading) {
+      <div class="loading-icon">
+        <explorers-loading-icon />
+      </div>
+    }
```
hallieswan, line 12: This looks great! 🎉

libs/explorers/shared/src/lib/components/terms-of-service/terms-of-service.component.scss
```diff
@@ -1,11 +1,31 @@
...
+  min-height: calc(100vh - var(--header-height) - var(--footer-height) + 1px);
+
+  .terms-of-service {
+    flex: 1 1 auto;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+
+    @media (width < 965px) {
```
hallieswan, line 15: We use a media query with the same condition (`width < 965px`) in `header.component.scss` and `hero.component.scss` as well -- would it be useful to store this width in a variable, maybe as a new breakpoint?

libs/explorers/shared/src/lib/components/terms-of-service/terms-of-service.component.scss
```diff
@@ -1,11 +1,31 @@
...
+      // adds consistent margin when the screen is smaller
+      margin: 0 20px;
+    }
+
+    .loading-icon {
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      width: min(925px, 100%);
```
hallieswan, line 24: We could use `--container-max-width-sm` instead of hardcoding 925px in this stylesheet and in `hero.component.scss`

libs/explorers/shared/src/lib/components/not-found/not-found.component.scss
hallieswan: The updated padding looks great!

hallieswan (changes requested):
This looks great! Just a couple suggestions, let me know if you have any questions!

libs/explorers/shared/src/lib/components/not-found/not-found.component.scss
sagely1: Since this is one of the few pages with a dark background, it seemed less jarring than 100% white to a darker background.

libs/explorers/shared/src/lib/components/not-found/not-found.component.scss
sagely1: A variable like content-min-height would be great but this calculates a dynamic height based on the current vertical height (100vh).  So the variable we could combine would be the heights of header and footer but I'm not sure what I would call that.

libs/explorers/shared/src/lib/components/terms-of-service/terms-of-service.component.scss
sagely1: You're right.  I wanted to use this variable but it is not a great name considering since we won't need the existing breakpoints.  I think I'll define this value as main-content-desktop-width.

libs/explorers/shared/src/lib/components/terms-of-service/terms-of-service.component.scss
sagely1: Since we won't need the existing breakpoints, I think I'll just use these variables and get rid of the container variables.

```
$main-content-desktop-width: 925px;
$main-content-margin: 20px;
$main-content-total-width: $main-content-desktop-width + ($main-content-margin * 2);
```

$main-content-total-width will represent the width including the margins.

libs/explorers/shared/src/lib/components/terms-of-service/terms-of-service.component.ts
```diff
@@ -24,16 +27,21 @@ export class TermsOfServiceComponent implements OnInit {
...
-        this.loading = false;
-      },
-    });
+    this.synapseService
+      .getTermsOfService()
+      .pipe(
+        takeUntilDestroyed(this.destroyRef),
+        finalize(() => {
+          this.loading = false;
+        }),
```
sagely1, line 36: improvment to address memory leaks.
finalize() added to DRY up code.

libs/explorers/styles/src/lib/_variables.scss
```diff
@@ -48,19 +48,17 @@ $extra-colors: (
 }
 
 // -------------------------------------------------------------------------- //
-// Container
```
sagely1, line 51: I don't think we'll need container vars going forward.

hallieswan (approved):
Looks great to me! Good catch on cleaning up the TOS memory leak!

PR #3192, opened 2025-05-20
feat(model-ad): created loading-container and configurable loading-icon (MG-225)

sagely1:
### Description
This PR creates a loading container that display the number of results.  See [Figma link](https://www.figma.com/design/GtUYqZBxoOnPtjV1hLcTgw/Model-AD-Explorer-Designs?node-id=798-11035&t=Du2yf5D1NVZ2NLEd-0)

This also makes the loading-icon component's colors configurable.

### PR
[link](https://sagebionetworks.jira.com/jira/software/c/projects/MG/boards/243?selectedIssue=MG-225)

https://www.loom.com/share/39e646724bb34b4995467a15166c5ef2

### DIFF
--- a/libs/explorers/util/src/lib/loading-container/loading-container.component.html
+++ b/libs/explorers/util/src/lib/loading-container/loading-container.component.html
@@ -4,5 +4,5 @@
     [colorCentral]="colorCentral"
     [colorOutermost]="colorOutermost"
   ></explorers-loading-icon>
-  <h1>Loading {{ count }} results</h1>
+  <p>Loading {{ count | number }} results for you</p>
 </div>
--- a/libs/explorers/util/src/lib/loading-container/loading-container.component.html
+++ b/libs/explorers/util/src/lib/loading-container/loading-container.component.html
@@ -1,0 +1,8 @@
+<div class="loading-container">
+  <explorers-loading-icon
+    [colorInnermost]="colorInnermost"
+    [colorCentral]="colorCentral"
+    [colorOutermost]="colorOutermost"
+  ></explorers-loading-icon>
+  <h1>Loading {{ count }} results</h1>
+</div>
--- a/libs/explorers/util/src/lib/loading-container/loading-container.component.ts
+++ b/libs/explorers/util/src/lib/loading-container/loading-container.component.ts
@@ -1,0 +1,17 @@
+import { CommonModule } from '@angular/common';
+import { Component, Input, ViewEncapsulation } from '@angular/core';
+import { LoadingIconComponent } from '../loading-icon/loading-icon.component';
+
+@Component({
+  selector: 'explorers-loading-container',
+  imports: [CommonModule, LoadingIconComponent],
+  templateUrl: './loading-container.component.html',
+  styleUrls: ['./loading-container.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class LoadingContainerComponent {
+  @Input() colorOutermost = '#8B8AD1';
+  @Input() colorCentral = '#8B8AD1';
+  @Input() colorInnermost = '#8B8AD1';
+  @Input() count = 0;
+}
--- a/libs/explorers/util/src/lib/loading-icon/loading-icon.component.html
+++ b/libs/explorers/util/src/lib/loading-icon/loading-icon.component.html
@@ -2,19 +2,19 @@
   <div class="loading-icon-inner">
     <div class="spinner">
       <div class="spinner-container">
-        <div class="hex0 hex0-outermost"></div>
-        <div class="hex120 hex120-outermost"></div>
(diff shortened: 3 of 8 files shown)


libs/explorers/util/src/lib/loading-icon/loading-icon.component.scss
```diff
@@ -2,7 +2,7 @@
 
 .loading-icon {
   position: relative;
-  width: 100px;
+  width: 200px;
```
sagely1, line 5: New loading icon is bigger

libs/explorers/util/src/lib/loading-container/loading-container.component.html
```diff
@@ -0,0 +1,8 @@
+<div class="loading-container">
+  <explorers-loading-icon
+    [colorInnermost]="colorInnermost"
+    [colorCentral]="colorCentral"
+    [colorOutermost]="colorOutermost"
+  ></explorers-loading-icon>
+  <h1>Loading {{ count }} results</h1>
```
hallieswan, line 7: The design has slightly different language and uses `<h2>` for this element, so the font is DM Sans Regular instead of Bold -- 

text in loom | text in design 
|:---:|:---:|
| loom | figma |

Let's try something like this? 

```suggestion
  <h2>Loading {{ count }} results for you</h2>
```

libs/explorers/util/src/lib/loading-container/loading-container.component.ts
```diff
@@ -0,0 +1,17 @@
...
+  imports: [CommonModule, LoadingIconComponent],
+  templateUrl: './loading-container.component.html',
+  styleUrls: ['./loading-container.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class LoadingContainerComponent {
+  @Input() colorOutermost = '#8B8AD1';
+  @Input() colorCentral = '#8B8AD1';
+  @Input() colorInnermost = '#8B8AD1';
+  @Input() count = 0;
```
hallieswan, line 16: In the design, the result count is formatted with commas when the value is large, e.g. 24393 is displayed as 24,393. Can we update the component to handle that formatting?

libs/explorers/util/src/lib/loading-container/loading-container.component.ts
```diff
@@ -0,0 +1,17 @@
+import { CommonModule } from '@angular/common';
+import { Component, Input, ViewEncapsulation } from '@angular/core';
+import { LoadingIconComponent } from '../loading-icon/loading-icon.component';
+
+@Component({
+  selector: 'explorers-loading-container',
+  imports: [CommonModule, LoadingIconComponent],
+  templateUrl: './loading-container.component.html',
+  styleUrls: ['./loading-container.component.scss'],
+  encapsulation: ViewEncapsulation.None,
```
hallieswan, line 10: Just curious, is this necessary for this component?

libs/explorers/util/src/lib/loading-icon/loading-icon.component.html
```diff
@@ -1,20 +1,20 @@
...
+<div class="loading-icon" [ngStyle]="customColors">
   <div class="loading-icon-inner">
     <div class="spinner">
       <div class="spinner-container">
-        <div class="hex0"></div>
-        <div class="hex120"></div>
-        <div class="hex240"></div>
+        <div class="hex0 hex0-outermost"></div>
+        <div class="hex120 hex120-outermost"></div>
+        <div class="hex240 hex240-outermost"></div>
```
hallieswan, line 7: Could these classes be simplified a bit, maybe something like this -- 

```suggestion
        <div class="hex0 outermost"></div>
        <div class="hex120 outermost"></div>
        <div class="hex240 outermost"></div>
```

hallieswan (changes requested):
Looks great, just a couple small tweaks based on the design!

libs/explorers/util/src/lib/loading-icon/loading-icon.component.html
sagely1: good idea!

libs/explorers/util/src/lib/loading-container/loading-container.component.html
sagely1: good catch.  I think I'll change this to a <p> tag since header tags should ideally be for section headings and this is just a standalone piece of text.  I'll go with a font-weight of 300 to try and approximate the Figma design as the design has normal weight and that is not what we see.

libs/explorers/util/src/lib/loading-container/loading-container.component.ts
sagely1: Good idea.  I'll add a number pipe.

hallieswan (approved):
Looks great!

PR #3194, opened 2025-05-21
feat(observability): add Pyroscope to the observability stack (SMR-153)

tschaffter:
## Architecture

```mermaid
flowchart TD
  subgraph Application Layer
    APP[Spring Boot App]
  end

  subgraph Storage
    PYRO[Pyroscope]
  end

  subgraph Visualization
    GRAFANA[Grafana]
  end

  %% Traces
  APP -->|profiles| PYRO

  %% Grafana visualizes from storage
  GRAFANA -->|view profiles| PYRO
  GRAFANA
```

## Related Issues

- https://sagebionetworks.jira.com/browse/SMR-153

## Changelog

- Create the `observability-pyroscope` project.
- Start Pyroscope along Grafana.
- Create app configuration class in `agora-gene-api` to specify Pyroscope's server address.
- Configure `agora-gene-api` to push profiles to Pyroscope.
- Visualize profiles data in Grafana.

### Build the Observability stack

```bash
observability-build-images
```

### Start the Observability stack

```bash
nx serve-detach observability-grafana
```

### Access Grafana's web UI

Navigate to `http://localhost:8501`.

- Username: `admin`
- Password: `admin`

### Data Profiles Data Source

Connection URL: `http://observability-pyroscope:8511`

### Visualize Profiles

1. Login into Grafana.
2. Select Drilldown > Profiles.
3. Specify `agora-gene-api` to filter data for this service.

image


apps/observability/otel-collector/otel-collector-config.yaml
```diff
@@ -18,7 +18,7 @@ exporters:
     endpoint: 0.0.0.0:8510
 
   otlp/tempo:
```
BryanFauble, line 20: There isn't a need at this moment, however, when/if we send data to a backend that requires a security token to be passed in this way of specifying the configuration will not work (As we do not want to expose the tokens).

It may be prudent to determine a way to inject the config.yaml without seeing it's content to support this ahead of time.

An example for signoz: https://signoz.io/opentelemetry/collector-nodejs/#creating-the-collector-configuration-file

In the example it has a section to add a header for all out-bound requests:
```
exporters:

  otlp:
    endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT}
    tls:
      insecure: false
    headers:
      "signoz-ingestion-key": ${SIGNOZ_INGESTION_KEY}
  debug:
    verbosity: normal

```

BryanFauble (commented):
LGTM!

PR #3207, opened 2025-05-28
fix(agora): upload data logs when step times out (AG-1800)

hallieswan:
## Description

The Agora e2e tests have been timing out on the "Start Explorer" step. We would like to upload logs to help us debug.

## Related ticket

- [AG-1800](https://sagebionetworks.jira.com/browse/AG-1800)

## Changelog

- Add a timeout to the "Start Explorer" step, so that subsequent steps can run before hitting the job's overall timeout
- Attach logs from data service as an artifact on failure
- Make a minimal change to an Agora library, so that Agora is affected on this PR

### DIFF
--- a/.github/workflows/e2e-explorer.yaml
+++ b/.github/workflows/e2e-explorer.yaml
@@ -126,11 +126,29 @@ jobs:
           fi
 
       - name: Start Explorer
+        id: start-explorer
+        timeout-minutes: 40
         env:
           cmd: 'nx run ${{ inputs.explorer }}-apex:serve-detach'
         run: |
           devcontainer exec --workspace-folder ../sage-monorepo bash -c ". ./dev-env.sh \
               && ${cmd}"
+
+      - name: Write out Explorer data logs
+        if: ${{ failure() }}
+        env:
+          cmd: 'docker logs ${{ inputs.explorer }}-data > data-logs.txt'
+        run: |
+          devcontainer exec --workspace-folder ../sage-monorepo bash -c ". ./dev-env.sh \
+              && ${cmd}"
+
+      - name: Upload Explorer data logs
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: 'data-logs-${{ inputs.explorer }}'
+          path: data-logs.txt
+          retention-days: 5
 
       - name: Run Explorer e2e tests
         env:
@@ -145,7 +163,7 @@ jobs:
               && workspace-docker-stop"
 
       - uses: actions/upload-artifact@v4
-        if: ${{ !cancelled() }}
+        if: ${{ !cancelled() && steps.start-explorer.conclusion != 'failure' }}
         with:
           name: 'playwright-report-${{ inputs.explorer }}'
           path: playwright-report/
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-explorer.yaml
```diff
@@ -126,12 +126,30 @@ jobs:
           fi
 
       - name: Start Explorer
+        id: start-explorer
+        timeout-minutes: 40
```
tschaffter, line 130: Is the explorer expected to take that much time to start? How was this number identified?

hallieswan: Ideally, the explorer would take much less time to start. However, the "Start Explorer" step downloads data from Synapse (~0.5 GB for [Agora](https://www.synapse.org/Synapse:syn12177492)), so if the network connection on the runner is particularly slow, then the data download could take longer. For example, see these successful runs that took [17 min](https://github.com/Sage-Bionetworks/sage-monorepo/actions/runs/14890718160/job/41821963042) and [22 min](https://github.com/Sage-Bionetworks/sage-monorepo/actions/runs/14911723651/job/41887660380).

The number was selected based on the longest successful runs on push to `main` -- [this run](https://github.com/Sage-Bionetworks/sage-monorepo/actions/runs/14736070274/job/41365532768) took ~35 minutes to start Agora, so 40 minutes was chosen to provide some buffer. We could certainly choose a shorter time, if we want to cancel jobs where the download is taking a long time and retry the job instead.

hallieswan:
@tschaffter - the new steps ("Write out Explorer data logs", "Upload Explorer data logs") aren't listed in the workflow steps for the [run for this PR](https://github.com/Sage-Bionetworks/sage-monorepo/actions/runs/15311745786/job/43078452300?pr=3207) -- I think that the base branch workflow is used instead of the PR's when using the `pull_request_target` workflow trigger. I think we'll need to merge this PR to update the workflow to upload the `agora-data` logs. 

image

.github/workflows/e2e-explorer.yaml
tschaffter: 40 minutes is a very long time. Does the e2e test need to download that much data? Out of curiosity, would downloading these data no longer be required if the REST API server returns paginated responses (#3202).

The download speed is surprisingly slow.

Improving the performance of the e2e test are out of scope of this PR, but that something that should be considered.

tschaffter (approved):

.github/workflows/e2e-explorer.yaml
hallieswan: > Does the e2e test need to download that much data? 

The e2e tests need to download data to populate the database. Since we don't navigate to every page in the app, we could use a subset of data. Each collection is stored as a JSON file in Synapse. Is there a way to filter a JSON file from Synapse without downloading the file? If not, I think the data would need to be structured differently in Synapse to support downloading a subset, e.g. provide files with fewer documents or provide files with test data. 

> would downloading these data no longer be required if the REST API server returns paginated responses

Since the issue is related to how the data is downloaded from Synapse, I don't think that paginating the responses would allow us to immediately stop downloading all of the data.

tschaffter: I don't think the JSON file can be partially downloaded, however a small file, potentially a mock file, could be used for e2e testing.

PR #3219, opened 2025-06-11
feat(model-ad): add panel navigation tabs to model details (MG-244)

hallieswan:
## Description

Updates model details to include panel navigation tabs.

## Related Issue

- [MG-244](https://sagebionetworks.jira.com/browse/MG-244)

## Changelog

- Updates model details to include panel navigation tabs
- Adds shared panel navigation component, based on Agora's [gene-details component](https://github.com/Sage-Bionetworks/sage-monorepo/tree/main/libs/agora/genes/src/lib/components/gene-details)
- Adds shared helper service to support panel navigation component, based on Agora's [helper service](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/agora/services/src/lib/helper.service.ts)
- Updates model-details routes to allow URLs with `:tab` and `:subtab` parameters
- Adds model details e2e tests
- Updates `model-ad-data`'s `.env.example` to use a newer data version
- Adds missing path for `model-ad-testing` to `tsconfig.base.json`

## Validation

```
model-ad-build-images && model-ad-docker-start
```

Navigate to a model with all tabs: http://localhost:8000/models/3xTg-AD

https://github.com/user-attachments/assets/b54470da-6b7a-4902-b04e-0054b54800e2

Navigate to a model with some tabs disabled: http://localhost:8000/models/APOE4

image

Navigate to a model that doesn't exist: http://localhost:8000/models/DOES-NOT-EXIST

image

[MG-244]: https://sagebionetworks.jira.com/browse/MG-244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/styles/src/lib/_mixins.scss
+++ b/libs/explorers/styles/src/lib/_mixins.scss
@@ -35,12 +35,6 @@
   }
 }
 
-@mixin reset-ul() {
-  list-style: none;
-  padding: 0;
-  margin: 0;
-}
-
 @mixin container($width: 'md') {
   max-width: var(--container-max-width-#{$width});
   padding-left: 0;
--- a/libs/explorers/styles/src/lib/_mixins.scss
+++ b/libs/explorers/styles/src/lib/_mixins.scss
@@ -35,6 +35,12 @@
   }
 }
 
+@mixin reset-ul() {
+  list-style: none;
+  padding: 0;
+  margin: 0;
+}
+
 @mixin container($width: 'md') {
   max-width: var(--container-max-width-#{$width});
   padding-left: 0;
--- a/libs/explorers/ui/src/lib/components/panel-navigation/panel-navigation.component.spec.ts
+++ b/libs/explorers/ui/src/lib/components/panel-navigation/panel-navigation.component.spec.ts
@@ -1,0 +1,63 @@
+import { Panel } from '@sagebionetworks/explorers/models';
+import { HelperService } from '@sagebionetworks/explorers/services';
+import { render, screen } from '@testing-library/angular';
+import userEvent from '@testing-library/user-event';
+import { PanelNavigationComponent } from './panel-navigation.component';
+
+class MockHelperService {
+  getOffset = jest.fn().mockReturnValue({ top: 100 });
+}
+
+const panelChangeSpy = jest.fn();
+
+const panelsMock: Panel[] = [
+  { name: 'panel1', label: 'Panel1', disabled: false },
+  { name: 'panel2', label: 'Panel2', disabled: false },
+  { name: 'panel3', label: 'Panel3', disabled: true },
+];
+
+async function setup(panels: Panel[] = panelsMock, activePanel = 'panel1', activeParent = '') {
+  const user = userEvent.setup();
+  const component = await render(PanelNavigationComponent, {
+    providers: [{ provide: HelperService, useClass: MockHelperService }],
+    componentInputs: {
+      panels,
+      activePanel,
+      activeParent,
+    },
+    componentOutputs: {
+      panelChange: { emit: panelChangeSpy } as any,
+    },
+  });
+  return { component, user };
+}
+
+describe('PanelNavigationComponent', () => {
+  it('should create', async () => {
+    const { component } = await setup();
+    expect(component).toBeTruthy();
+  });
+
+  it('should emit panelChange when clicking on a panel', async () => {
+    const { user } = await setup();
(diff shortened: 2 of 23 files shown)


libs/explorers/ui/src/lib/components/panel-navigation/panel-navigation.component.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  const component = await render(PanelNavigationComponent, {
+    providers: [{ provide: HelperService, useClass: MockHelperService }],
+    componentInputs: {
+      panels,
+      activePanel,
+      activeParent,
+    },
+    componentOutputs: {
+      panelChange: { emit: panelChangeSpy } as any,
+    },
```
hallieswan, line 30: Based on [this article](https://timdeschryver.dev/blog/testing-signals-with-angular-testing-library) about testing signals with testing-library

libs/model-ad/model-details/src/lib/model-details.component.ts
```diff
@@ -1,9 +1,134 @@
...
+
+  activePanel = 'omics';
+  activeParent = '';
+
+  ngOnInit() {
+    this.route.paramMap.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((params: ParamMap) => {
+      this.isLoading = true;
+      if (this.configService.config.isPlatformServer) {
+        return;
+      }
```
hallieswan, line 62: When using SSR, `ngOnInit` runs on both the server and the client. The request fails on the server, which results in the not found page being shown briefly (before the client makes the request successfully, and then displays the model's content). Add this block to exit early on the server to prevent the not found page from being shown.

libs/explorers/styles/src/lib/_mixins.scss
```diff
@@ -35,6 +35,12 @@
   }
 }
 
+@mixin reset-ul() {
+  list-style: none;
+  padding: 0;
+  margin: 0;
+}
```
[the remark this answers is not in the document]
hallieswan: Good call, done in the next commit!

libs/model-ad/model-details/src/lib/model-details.component.spec.ts
```diff
@@ -1,20 +1,71 @@
...
+import { of } from 'rxjs';
 import { ModelDetailsComponent } from './model-details.component';
 
-async function setup() {
-  const result = await render(ModelDetailsComponent, {
+async function setup(model = modelMock, tab = 'omics', subtab = null, config = configMock) {
+  const user = userEvent.setup();
+
+  const mockActivatedRoute = {
+    paramMap: of(convertToParamMap({ model: model.model, tab: tab, subtab: subtab })),
```
sagely1, line 14: `model.model` is an unfortunate naming collision.  We could request that the collection in MongoDB uses something like 'name' instead of 'model' so we end up with something more like `model.name`  We can track this in a separate PR.  https://sagebionetworks.jira.com/browse/MG-253

libs/model-ad/model-details/src/lib/model-details.component.ts
```diff
@@ -1,9 +1,134 @@
...
+    },
+  ];
+
+  activePanel = 'omics';
+  activeParent = '';
+
+  ngOnInit() {
+    this.route.paramMap.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((params: ParamMap) => {
+      this.isLoading = true;
+      if (this.configService.config.isPlatformServer) {
```
sagely1, line 60: as an enhancement for readability, you could wrap the rest of the logic in a method so the intent is clearer.  For example (pseudocode):
```
ngOnInit() {
  // only fetch data during client hydration
  if (isPlatformBrowser(this.platformId)) {
     this.loadPanels();
  }
}

loadPanels() {
  (rest of logic here)
}
```

sagely1 (changes requested):
Looks great overall!  I included a note for readability to avoid bloated ngOnInit() methods.

libs/model-ad/model-details/src/lib/model-details.component.spec.ts
hallieswan: Agreed, it is a bit confusing. Thanks for opening the ticket!

tschaffter (approved):

sagely1 (approved):
Looks great!

PR #3077, opened 2025-04-04
fix(agora): migrate agora unit tests to jest (AG-1627)

hallieswan:
## Description

Agora unit tests were originally written to be run by Jasmine and were disabled as part of the migration into the `sage-monorepo`. This PR re-enables Agora unit tests by migrating them to use Jest.

## Related Issues

- [AG-1627](https://sagebionetworks.jira.com/browse/AG-1627)

## Changelog

- Migrates Agora unit tests to Jest
- Rewrites `gene-comparison-tool-filter-list` and `gene-comparison-tool-filter-list-item` tests to use testing library
- Fixes bug in `gene-nominations` preventing `team_data` from being added to the `nominations` object
- ~~Imports `@types/dom-to-image` as types from `dom-to-image-more` and removes need for `dom-to-image-more` types file~~ - added in #3152 
- Adds `ActivatedRouteStub` and `SvgIconServiceStub` and re-enables `GenesServiceStub`

[AG-1627]: https://sagebionetworks.jira.com/browse/AG-1627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3222, opened 2025-06-12
fix(agora, model-ad): fix e2e report upload in CI, fix slow-running agora unit tests (AG-1627)

hallieswan:
## Description

This PR makes the following changes related to tests in CI:  
1. By default, Playwright HTML reports open when tests fail. However, that prevents the CI job from uploading the e2e report (see [this run](https://github.com/Sage-Bionetworks/sage-monorepo/actions/runs/15591923844/job/43913025223)). Here, we prevent the HTML report from opening in CI, so that the e2e report will be uploaded on failure. 
2. After merging #3077, the CI runner exited with error code 137 (see [this run](https://github.com/Sage-Bionetworks/sage-monorepo/actions/runs/15595922462/job/43926204347)). The `agora-services` and `agora-shared` tests were particularly slow, perhaps due to importing very large files from `synapse-api-client-angular`, which results in babel warnings. Here, we remove the unused method imported from `synapse-api-client-angular` in Agora's `SynapseApiService`. The same change was made to the Explorer's `SynapseApiService` in #3211, see related comment [here](https://github.com/Sage-Bionetworks/sage-monorepo/pull/3211/files#r2126072081).

## Related Issue

- [AG-1627](https://sagebionetworks.jira.com/browse/AG-1627)

## Changelog

- Prevents Playwright HTML reports from opening in CI
- Removes unused method `getWikiAlternate` from agora-services `SynapseApiService`
- Updates to latest Playwright version, so that all projects are "affected"

### E2E tests: 

Edit `not-found.spec.ts` so the test will fail, e.g. change expected text content to "This will fail".
Run the e2e tests as if in CI: `CI=true nx e2e model-ad-app`.
Tests should fail.
HTML report should not open.

### Unit tests: 

Run `nx test agora-shared -- --no-cache`  to skip Jest cache -- 
- Before removing dependency on `synapse-api-client-angular`, memory usage increased by 3.3GB while running tests. Tests took 50s to run.
- After removing dependency on `synapse-api-client-angular`, memory usage increased by 2.2GB while running tests. Tests took 20s to run.

I was able to successfully run all tests on my EC2. We will also validate using the CI job on this PR -- the Playwright package version was updated, so that all projects are "affected" and will run their unit tests.

[AG-1627]: https://sagebionetworks.jira.com/browse/AG-1627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/agora/app/playwright.config.ts
+++ b/apps/agora/app/playwright.config.ts
@@ -22,7 +22,7 @@ export default defineConfig({
   /* Opt out of parallel tests on CI. */
   workers: process.env['CI'] ? 2 : undefined,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: process.env['CI'] ? [['list'], ['html']] : 'html',
+  reporter: process.env['CI'] ? [['list'], ['html', { open: 'never' }]] : 'html',
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     baseURL,
--- a/apps/model-ad/app/playwright.config.ts
+++ b/apps/model-ad/app/playwright.config.ts
@@ -22,7 +22,7 @@ export default defineConfig({
   /* Opt out of parallel tests on CI. */
   workers: process.env['CI'] ? 2 : undefined,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: process.env['CI'] ? [['list'], ['html']] : 'html',
+  reporter: process.env['CI'] ? [['list'], ['html', { open: 'never' }]] : 'html',
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     baseURL,
--- a/libs/agora/services/src/lib/synapse-api.service.ts
+++ b/libs/agora/services/src/lib/synapse-api.service.ts
@@ -1,15 +1,11 @@
 // -------------------------------------------------------------------------- //
 // External
 // -------------------------------------------------------------------------- //
-import { Injectable } from '@angular/core';
 import { HttpClient } from '@angular/common/http';
-import { tap } from 'rxjs/operators';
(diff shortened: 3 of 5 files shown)


tschaffter (commented):
Thank you for this fix!

- The PR title is missing the reference to the Jira ticket.
- The Jira ticket listed under Related Issues does not have a link.

> After removing dependency on synapse-api-client-angular

Does that mean that the e2e no longer depend on the auto-generated Angular API client for Synapse? Do you know why its use increase the memory usage and slow down the tests by that much?

hallieswan:
@tschaffter I've updated the PR title and description.

> Does that mean that the e2e no longer depend on the auto-generated Angular API client for Synapse? Do you know why its use increase the memory usage and slow down the tests by that much?

Agora's `SynapseApiService` no longer include methods from the auto-generated Synapse client, so the _unit_ tests don't depend on the client.<sup>**</sup> I suspect that the slow down and memory usage is related to Babel compilation and the size of the auto-generated files. When the code includes references to the auto-generated client, many of these warnings are displayed when running the tests for the first time: 

```
[BABEL] Note: The code generator has deoptimised the styling of /workspaces/sage-monorepo/libs/synapse/api-client-angular/src/lib/api/entityServices.service.ts as it exceeds the max of 500KB.
```

Jest does some caching, so when the test suite is rerun, these message aren't displayed and the test suite runs much faster. However, you can reproduce the issue by skipping the Jest cache (and nx cache) (e.g. `nx test agora-shared --skip-nx-cache -- --no-cache`).

**The Agora `app.config.ts` imports `BASE_PATH` from the auto-generated client, so the app and _e2e_ tests still depend on the auto-generated client. However, since the issue seems to be specifically related to Jest and Babel, I don't think that is an issue.

tschaffter (approved):

PR #3226, opened 2025-06-13
feat(model-ad): add model details omics panel content (MG-215)

hallieswan:
## Description

Adds the omics panel content to the model details page.

## Related Issue

- [MG-215](https://sagebionetworks.jira.com/browse/MG-215)

## Changelog

- Adds `model-details-omics` component
- Adds shared `resource-cards` component to render a list of resource cards in a grid
- Adds shared styling for `<section>`
- Uses default loading icon colors for model details page
- Fixes navigation for internal links in `resource-card` component

## Preview

Preview shared components in storybook: 
1. Run Storybook: `nx storybook explorers-storybook`
2. View resource cards: http://localhost:4400/?path=/story/ui-cards-resourcecardscomponent--demo

Preview model details page: 
1. Build and run Model-AD: `model-ad-build-images && model-ad-docker-start`
4. Navigate to a model details page: http://localhost:8000/models/3xTg-AD

https://github.com/user-attachments/assets/03d0a4fc-5d07-4e9b-81e2-561ff03f4aae

[MG-215]: https://sagebionetworks.jira.com/browse/MG-215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/styles/src/lib/_section.scss
+++ b/libs/explorers/styles/src/lib/_section.scss
@@ -1,7 +1,7 @@
 @use 'variables' as vars;
 
 section {
-  max-width: var(--section-width);
+  max-width: var(--section-max-width);
   padding: var(--section-padding);
   margin-left: auto;
   margin-right: auto;
--- a/libs/explorers/styles/src/lib/_variables.scss
+++ b/libs/explorers/styles/src/lib/_variables.scss
@@ -145,11 +145,11 @@ $main-content-total-width: $main-content-desktop-width + ($main-content-margin *
 // Section
 // -------------------------------------------------------------------------- //
 
-$section-width: 1120px;
+$section-max-width: 1120px;
 $section-padding: 50px;
 
 :root {
-  --section-width: #{$section-width};
+  --section-max-width: #{$section-max-width};
   --section-padding: #{$section-padding};
 }
 


libs/explorers/styles/src/lib/_mixins.scss
```diff
@@ -35,14 +35,6 @@
   }
 }
 
-@mixin container($width: 'md') {
```
hallieswan, line 38: Will use new shared `SectionComponent` instead of creating a shared mixin.

libs/explorers/ui/src/lib/components/resource-cards/resource-cards.component.scss
```diff
@@ -0,0 +1,17 @@
+@use 'libs/explorers/styles/src/lib/mixins';
+
+.resource-cards {
+  display: grid;
+  gap: 30px 50px;
+  grid-template-columns: 1fr;
+
+  @include mixins.respond-to('large') {
+    grid-template-columns: 1fr 1fr;
+  }
```
hallieswan, line 10: Arrange resource cards in 1 column when the page is narrow and 2 columns when the page is wide: 

https://github.com/user-attachments/assets/d91adbba-c2f7-4073-89b3-e1e4e7b4e1b0

libs/model-ad/model-details/src/lib/components/model-details-omics/model-details-omics.component.scss
```diff
@@ -0,0 +1,3 @@
+h2 {
+  margin-bottom: 30px;
```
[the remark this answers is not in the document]
hallieswan: We are moving away from using spacing variables, so will ignore this comment.

libs/model-ad/model-details/src/lib/components/model-details-omics/model-details-omics.component.html
```diff
@@ -0,0 +1,4 @@
+<explorers-section containerSize="sm-md">
```
sagely1, line 1: If we just need wrapping, we could eliminate the need for this component and needing to specify a size.

libs/model-ad/model-details/src/lib/model-details.component.html
```diff
@@ -1,10 +1,14 @@
 <div [class]="{ 'model-details': true, loading: isLoading }">
   @if (isLoading) {
     <div id="loading">
-      <explorers-loading-icon />
+      <explorers-loading-icon
+        [colorOutermost]="loadingIconColors.colorOutermost"
+        [colorCentral]="loadingIconColors.colorCentral"
+        [colorInnermost]="loadingIconColors.colorInnermost"
+      />
```
sagely1, line 8: Something we could consider is creating a model-ad-loading-icon component so we don't have to set the loadingIconColors everywhere.  It would effectively wrap this component with the needed props.

sagely1 (changes requested):
Looks great.  I think we can eliminate the Section component.

libs/explorers/ui/src/lib/components/section/section.component.html
```diff
@@ -0,0 +1,7 @@
+<div class="section">
```
sagely1, line 1: as a suggestion, we can use <section> in places such as this to give this semantic meaning.

tschaffter: @sagely1 "<>" should be surrounded by quotes. Classic mistake :-D

libs/explorers/ui/src/lib/components/section/section.component.scss
```diff
@@ -0,0 +1,34 @@
...
+  @extend %container-base;
+
+  max-width: 1440px;
+}
+
+.container-lg {
+  @extend %container-base;
+
+  max-width: 1920px;
+}
```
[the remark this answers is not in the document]
hallieswan: These set a max-width for different size sections so the content doesn't extend past the max-width when the page is very wide. We won't use all of these section sizes in a single app, so we can certainly simplify to provide fewer presets sizes! I'll make that change in the next commit.

libs/model-ad/model-details/src/lib/components/model-details-omics/model-details-omics.component.html
hallieswan: Sounds good, I'll use `<section>` instead.

libs/explorers/ui/src/lib/components/resource-cards/resource-cards.component.scss
```diff
@@ -0,0 +1,17 @@
+@use 'libs/explorers/styles/src/lib/mixins';
+
+.resource-cards {
+  display: grid;
```
sagely1, line 4: great use of CSS grid!

sagely1 (approved):
Added one suggestion on variable naming but otherwise this is great!  Looks like I can align on your work by adding more section tags to other components!

libs/explorers/styles/src/lib/_section.scss
```diff
@@ -1,5 +1,8 @@
-@use './mixins';
+@use 'variables' as vars;
 
-.container-sm {
-  @include mixins.container('sm');
+section {
+  max-width: var(--section-width);
```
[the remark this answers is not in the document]
hallieswan: Good call, done in the next commit!

PR #3228, opened 2025-06-17
chore(openchallenges): update Spring AI in MCP server and migrate to WebFlux (SMR-184)

tschaffter:
## Description

Update the MCP server before deploying it to AWS.

## Changelog

- Update to Spring AI v1.0.0.
- Remove `spring-dependency-management`
- Migrate from webmvc to webflux.


apps/openchallenges/mcp-server/build.gradle.kts
```diff
@@ -15,6 +14,23 @@ java {
   }
 }
 
+repositories {
+	mavenCentral()
+  mavenLocal()
+}
+
+dependencies {
+	implementation(libs.spring.ai.starter.mcp.server.webflux)
```
tschaffter, line 23: Spring Web MVC is used when using synchronous endpoints. WebFlux is used when developing async endpoints. This should enable the MCP server to return/stream the response asynchronously bit by bit instead of returning the response all at once, which generates a delay. Overall, the response time is expected to be smaller when using WebFlux.

rrchai (approved):
LGTM!

PR #3248, opened 2025-06-26
fix(model-ad): hide alzforum resource card when id is missing (MG-259)

hallieswan:
## Description

Some models do not have an `alzforum_id`. We should hide the corresponding resource card in this case.

## Related Issue

- [MG-259](https://sagebionetworks.jira.com/browse/MG-259)

## Changelog

- Hides ALZFORUM resource card when id is missing

## Preview

Run `model-ad-build-images && model-ad-docker-start`
Navigate to `http://localhost:8000/models/Abca7*V1599M/resources`
Alzforum card should not be shown: 

MG-259

[MG-259]: https://sagebionetworks.jira.com/browse/MG-259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sagely1 (approved):
LGTM!

PR #3252, opened 2025-06-27
fix(explorers): ensure playwright reports are uploaded in CI

hallieswan:
## Description

The `CI` environment variable was not set in the devcontainer, so playwright HTML reports were not prevented from opening (fix in #3222 depends on setting `CI` env var). Ensure that `CI` is set, so the correct configuration is used. 

## Related Issue

- N/A

## Changelog

- Set `CI` environment variable in devcontainer when running explorers e2e tests
- Always attempt reuse existing server when running explorers e2e tests; the CI workflow starts the app before running e2e tests

## Preview

Run the app: `model-ad-build-images && model-ad-app-start`
Edit `not-found.spec.ts` so the test will fail, e.g. change expected text content to "This will fail".
Run the e2e tests as if in CI: `CI=true nx e2e model-ad-app`
Tests should fail.
HTML report should not open.

### DIFF
--- a/.github/workflows/e2e-explorer.yaml
+++ b/.github/workflows/e2e-explorer.yaml
@@ -152,7 +152,7 @@ jobs:
 
       - name: Run Explorer e2e tests
         env:
-          cmd: 'nx run ${{ inputs.explorer }}-app:e2e'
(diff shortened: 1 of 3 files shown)


sagely1 (approved):
LGTM!

PR #3237, opened 2025-06-20
feat(openchallenges): introduce auth service and enforce scoped access to organization deletion (SMR-194, SMR-235)

tschaffter:
## Description

This PR introduces an API key–based authentication service to OpenChallenges. Initially, authentication is limited to members of the OpenChallenges team. However, the service is designed to allow pre-registered users to generate and manage their own API keys.

The authentication service is now integrated to secure a new endpoint in the organization service, which enables the deletion of organizations.

### Logging in and managing API keys

The sequence diagram below illustrates how a pre-registered user can log in using their credentials (username and password) to obtain an API key. Once authenticated, the user can use this API key to interact with the authentication service, including creating additional API keys with specified expiration dates. The user can also list and delete their API keys through the auth service.

```mermaid
sequenceDiagram
    autonumber

    participant User
    participant Auth as Auth Service

    Note over User,Auth: Step 1 - Login
    User->>Auth: POST /v1/auth/login\n{ username, password }
    Auth-->>User: { apiKey }

    Note over User,Auth: Step 2 - Create a new API key
    User->>Auth: POST /v1/auth/api-keys<br>Authorization: Bearer [apiKey]<br>Body: { name, expiresIn? }
    Auth-->>User: { id, name, key, expirationDate }

    Note over User,Auth: Step 3 - List API keys using original key
    User->>Auth: GET /v1/auth/api-keys<br>Authorization: Bearer [apiKey]
    Auth-->>User: [ List of API keys ]

    Note over User,Auth: Step 4 - List API keys using new key
    User->>Auth: GET /v1/auth/api-keys<br>Authorization: Bearer [newApiKey]
    Auth-->>User: [ List of API keys ]

    Note over User,Auth: Step 5 - Validate the new API key
    User->>Auth: POST /v1/auth/validate<br>{ apiKey: newApiKey }
    Auth-->>User: { valid: true, username, role, scopes }

    Note over User,Auth: Step 5b - Validate invalid key
    User->>Auth: POST /v1/auth/validate<br>{ apiKey: invalidKey }
    Auth-->>User: { valid: false }

    Note over User,Auth: Step 6 - Delete the created API key
    User->>Auth: DELETE /v1/auth/api-keys/{keyId}<br>Authorization: Bearer [apiKey]
    Auth-->>User: 204 No Content

    Note over User,Auth: Step 7 - Verify deleted key is rejected
    User->>Auth: GET /v1/auth/api-keys<br>Authorization: Bearer [deletedApiKey]
[...]

PR #3264, opened 2025-07-01
feat(openchallenges): add DELETE endpoint for deleting a challenge platform by ID (SMR-238)

vpchung:
## Description

This PR adds an endpoint to delete a challenge platform in the challenge service.

## Related Issue

[SMR-238](https://sagebionetworks.jira.com/browse/SMR-238)

## Changelog

- Add API description for deleting a challenge platform by ID.
- Implement the new endpoint in the challenge service.
- Secure the new endpoint with the auth service (user must have the role `admin`).
- Remove endpoint to get challenge platform by name.
- Add endpoint to get challenge platform by ID.
- Remove the logic from the challenge microservice that allows to return `Challenge` document in JSON-LD format.

## Notes

* Per our last discussion, we decided it's ideal to use the platform ID (number) as the unique identifier rather than the name (string). This aligns better with existing endpoints and is likely what users would expect.  
* If desired for historical reasons, I saved the endpoints that use the platform name under a new path: `/challengePlatforms/name/{challengePlatformName}`
   * the `operationId` for these endpoints are suffixed with `ByName` to distinguish them from those using the platform ID

### Start the OC stack

```bash
openchallenges-build-images
nx serve-detach openchallenges-apex
```

### Accessing Swagger UIs

- Auth service: http://localhost:8087/swagger-ui/index.html
- Challenge service: http://localhost:8085/swagger-ui/index.html

### Deleting a challenge platform

The endpoint returns a 401 response if the API key is invalid:

```json
{
  "error": "Unauthorized",
  "message": "Invalid API key",
  "status": 401
}
```

The endpoint returns 403 is the user is authenticated but does not have the role `admin`.

```json
{
  "title": "Access Denied",
  "status": 403,
  "detail": "You do not have permission to access this resource.",
  "type": null
}
```

> [!NOTE]
> The test user `researcher` has the role `user`.

The endpoint returns a 204 (No Content) response when the API key is valid.

> [!NOTE]
> For now, the user just need to be authenticated to be able to delete resources to simplify the implementation since the endpoint will only be used by internal team members behind our VPN. This behavior will be updated as soon we open up the API to the public.

[SMR-238]: https://sagebionetworks.jira.com/browse/SMR-238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/openchallenges/api-description/src/paths/challengePlatforms/@{challengePlatformName}.yaml
```diff
@@ -4,8 +4,8 @@ get:
   tags:
     - ChallengePlatform
   summary: Get a challenge platform
-  description: Returns the challenge platform specified
-  operationId: getChallengePlatform
+  description: Returns the challenge platform specified by its unique name
+  operationId: getChallengePlatformByName
```
tschaffter, line 8: @vpchung This is the unconventional endpoint that retrieve a resource based on its name and not its Id or login. I think that we are not using this endpoint in the web app. Could you please check? If we don't use it, then we should remove it.

tschaffter:
@vpchung The port labels are only applied after rebuilding the dev container. I find this behavior to be annoying but the good news is that you can forward the port manually. Sometimes VS Code does it on its own when it detects that an app starts listing to a port. Based on your screenshot, it's all good on your end.

The Swagger UI screenshot shows that the default credentials defined in the OpenAPI description are `admin` / `admin123`. Those are old credentials that have been changed to `admin` / `changeme`.

Have you pulled the changes I made to the PR before building the Docker images and testing the implementation?

vpchung:
> The Swagger UI screenshot shows that the default credentials defined in the OpenAPI description are admin / admin123. Those are old credentials that have been changed to admin / changeme.

No, the default values were admin / changeme, as you've said.  I had changed it to what you see in the screenshot, since those were the credentials used when reviewing #3237 - I thought I had to replicate that step.  I see now that if I had just left the default values, it would've worked 😆 

Screenshot 2025-07-09 at 7 30 28 PM

Thanks!

> The port labels are only applied after rebuilding the dev container.

But I _did_ rebuild the container after pulling all of the latest changes -___-

vpchung (commented):
I'm not able to approve the PR since I opened it, but the new feature is working as expected!

Screenshot 2025-07-09 at 7 33 20 PM

Here is my pseudo-approval: ✅

tschaffter (approved):

PR #3277, opened 2025-07-06
feat(openchallenges): scaffold Python app and lib for OC API client and Tools

tschaffter:
## Description

This pull request introduces a Python library and accompanying application for OpenChallenges.

As part of this effort, we also revisit how the Nx plugin [@nxlv/python](https://github.com/nxlv/nx-python) can be used to better organize Python projects within the monorepo.

## Related Issue

- [SMR-252](https://sagebionetworks.jira.com/browse/SMR-252)

## Changelog

- Update `@nxlv/python` to align on Nx 21.
- Create the library project `openchallenges-api-client-python` that will receive code generated by OpenAPI generator.
- Create the application project `openchallenges-tools` that will include a collection of tools to manage OC data.
- Document how to organize Python project using `@nxlv/python` and the shared virtual environments.
- Provide insights into how to develop an individual Python package without a solution already available in the monorepo.

## Update `@nxlv/python`

```bash
pnpm add -w -D @nxlv/python@21.0.3
```

## Creating a Python app

```bash
nx generate @nxlv/python:uv-project openchallenges-tools \
  --directory apps/openchallenges/tools \
  --projectType application
```

## Creating a Python library

```bash
nx generate @nxlv/python:uv-project openchallenges-api-client-python \
  --directory libs/openchallenges/api-client-python \
  --projectType library
```

> [!NOTE]
> It's not pretty to pretty to include the "python" keyword as part of the project, but for now I will stick to our convention. The generator allows to update the package and module names, so that should provide additional flexibility if we want to publish the package as `openchallenges` or `openchallenges_client`, for instance.

## Adding the above library as a dependency of the app

```bash
nx run openchallenges-tools:add --name openchallenges-api-client-python --local
```

## Linting a project with Ruff

```bash
nx lint openchallenges-tools
```

## Testing a project with Pytest

```bash
nx test openchallenges-tools
```

## Building a project

```bash
nx build openchallenges-tools
```

Those files contain all the project code and dependencies, and those files are used to publish to a PyPi repository or could be easily copied to a Docker image.

## Creating a task to run a Python script

Add the following task (also called target) definition to the `project.json` of the app. This task will be used to execute the `hello.py` files generated by Nx.

```json
    "run-hello": {
      "executor": "@nxlv/python:run-commands",
      "options": {
        "command": "uv run openchallenges_tools/hello.py",
        "cwd": "{projectRoot}"
      }
    }
```

The `@nxlv/python:run-commands` wraps the `nx:run-commands` default Nx executor and activates the virtual environment before running the command.

After updating the app’s `hello.py` to demonstrate how to import the `openchallenges-api-client-python` library, you can run the task using:

```console
$ nx run-hello openchallenges-tools

> nx run openchallenges-tools:run-hello

Hello openchallenges-tools
Hello openchallenges-api-client-python

——————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 NX   Successfully ran target run-hello for project openchallenges-tools (980ms)
```

> [!TIP]
> Running the task `run-hello` just after modifying the code of the Python library `openchallenges-api-client-python` will automatically apply the changes without the need to build the library first. 

## Opening the shared virtual environment in VS Code

1. Open the Command Palette.
2. Python > Select Interpreter
3. Select `.venv/bin/python`

VS Code IntelliSense will now highlight the imports in all the Python files that below to a project that use the shared virtual environment.

image

## Adding an external dependency to a project

Example:

```bash
nx run openchallenges-tools:add click==8.2.1
```

The `click` package is added to the `pyproject.toml` of the project `openchallenges-tools`. However, the workspace uv lock file is updated, indicating that the project use the shared virtual environment.

## Dependency graph

The plugin `@nxlv/python` infers the dependencies between the two Python projects.

Run `nx graph` to visualize the project dependencies:

image

> [!NOTE]
> Check if that's to case when the projects do not use a shared virtual environment.

## Specify global setting for the `@nxlv/python:uv-project` project generator

In `nx.json`:

```json
{
  ...
  "generators": {
    "@nxlv/python:uv-project": {
      "codeCoverageHtmlReport": true,
      "codeCoverageThreshold": 100,
      "useNxReleaseLegacyVersioning": false
    }
  }
  ...
}
```

## Use shared virtual environment

The project generator `@nxlv/python:uv-project` has a property named `individualPackage` that specifies if the project use the shared virtual environment created in the root workspace or not.

By default, the value of this property depends on whether the root workspace includes the file `pyproject.toml` (this file exists in this monorepo) as defined in the code of the plugin. Hence, the value of `individualPackage` is `false` by default in our monorepo.

https://github.com/lucasvieirasilva/nx-plugins/blob/9e5fd1ad316213aaa7a9dc5e111e5e05949af053/packages/nx-python/src/generators/uv-project/generator.ts#L64

Unfortunately, it's not possible to override `individualPackage` set to `true` using a CLI option. However, it seems possible to override this property using in a [project template](https://www.npmjs.com/package/@nxlv/python). One use case if when importing existing and/or legacy projects that can not use the dependencies available in the shared virtual environment.

An alternative is to use one of the existing Python project like `agora-data` that behave as individual project. In order to achieve this, I rely on `nx:run-commands` instead of using the plugin `@nxlv/python`. The gap between the two could be reduce, e.g. by aligning the name of their project tasks.

Unfortunately, the option `individualPackage` specified below does NOT exist: 

```bash
nx generate @nxlv/python:uv-project sandbox-py-app-standalone \
  --directory apps/sandbox/py-app-standalone \
  --projectType application \
  --individualPackage true \
  --dry-run
```

TODO: Try generating a project after removing / renaming the workspace-level `pyproject.toml` to simulate a non-shared virtual environment before looking at the supposedly individual project generated to see if / how it differs.

UPDATE: The plugin `@nxlv/python` consider either that all the projects use the shared virtual environment, or that none of them do. Creating individual projects is still possible using the Python projects that already exist in the monorepo (before this PR was merged).

## References

- https://www.npmjs.com/package/@nxlv/python
- https://github.com/lucasvieirasilva/nx-plugins/issues/258
  - Discussion about the pro and cons of using the shared virtual environment.
- [Poetry Python Nx Monorepo](https://medium.com/better-programming/poetry-python-nx-monorepo-5750d8627024) (now old but the plugin goals and concepts remain the same)

[SMR-252]: https://sagebionetworks.jira.com/browse/SMR-252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rrchai (approved):
LGTM!

PR #2965, opened 2025-01-14
feat(agora): migrate Agora e2e tests (AG-1609)

hallieswan:
## Description

Migrates Agora end-to-end tests to the monorepo.

## Related Issue

- [AG-1609](https://sagebionetworks.jira.com/browse/AG-1609)

## Changelog

- Adds Playwright to agora-app: 
   ```
   nx g @nx/playwright:configuration --project=agora-app --webServerCommand="nx run agora-apex:serve-detach" --webServerAddress="http://localhost:8000" --linter="eslint"
  ```
- Updates to latest Playwright version: `pnpm update @playwright/test --latest`
- Installs [Playwright browsers and system dependencies](https://playwright.dev/docs/browsers) as part of `workspace-install`
- Migrates existing Agora e2e tests to monorepo, but skips GCT tests until [AG-1618](https://sagebionetworks.jira.com/browse/AG-1618) is fixed
- Lints agora-app e2e tests per [Playwright best practices](https://playwright.dev/docs/best-practices#lint-your-tests)
   
## Setup

Before running tests, ensure that `DATA_VERSION` in apps/agora/data/.env is 71. Then rebuild all Agora images with `agora-build-images`.

## Preview

Run e2e tests when agora containers are not yet running -- Playwright will start the services before running the tests:

```
$ nx run agora-app:e2e

> nx run agora-app:e2e


 NX   Ensuring Playwright is installed.

use --skipInstall to skip installation.
[WebServer]  Container agora-api-docs  Creating
[WebServer]  Container agora-mongo  Creating
[WebServer]  Container agora-api-docs  Created
[WebServer]  Container agora-mongo  Created
[WebServer]  Container agora-data  Creating
[WebServer]  Container agora-api  Creating


[WebServer]  Container agora-app  Creating
[WebServer]  Container agora-app  Created

[WebServer]  Container agora-apex  Created
[WebServer]  Container agora-api-docs  Starting
[WebServer]  Container agora-mongo  Starting
[WebServer]  Container agora-api-docs  Started
[WebServer]  Container agora-mongo  Started
[WebServer]  Container agora-api  Starting
[WebServer]  Container agora-data  Starting
[WebServer]  Container agora-data  Started
[WebServer]  Container agora-api  Started
[WebServer]  Container agora-data  Waiting
[WebServer]  Container agora-api  Waiting
[WebServer]  Container agora-api  Healthy
[WebServer]  Container agora-data  Exited
[WebServer]  Container agora-app  Starting
[...]

PR #3285, opened 2025-07-09
feat(model-ad): update data version to 62, add mongodb-schema package (MG-289)

sagely1:
## Description

Add mongodb-schema package to generate schema files.
Update env.example & collections to match version 62.

## Related Issue

[MG-289](https://sagebionetworks.jira.com/jira/software/c/projects/MG/boards/243?assignee=633b138697148a8301fe2c72&selectedIssue=MG-289)

## Changelog

- The data version in env.example was updated to the latest version (62).
- The collections.csv file was updated with new collections from version 62.
- New package: mongodb-schema package has been added.
Example usage:
`mongodb-schema <user>:<password>@localhost:27017 model-ad.model_overview > schema.json`

[MG-289]: https://sagebionetworks.jira.com/browse/MG-289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -61,6 +61,7 @@
     "json5": "2.2.3",
     "lodash": "4.17.21",
     "marked": "15.0.11",
+    "mongodb-schema": "^12.6.2",
```
hallieswan, line 64: We should pin the version. See conversation about the convention in the sage-monorepo I believe the convention in the monorepo is to pin the version number -- see conversation [here](https://github.com/Sage-Bionetworks/sage-monorepo/pull/2965#discussion_r1917087886)..

hallieswan (changes requested):
Looks good to me! Just one small change

package.json
tschaffter, line 64: - The package must be moved to `devDependencies` if it's used  only as a CLI.
- The version number must be pinned.

From the PR description: run the native command provided by the packaeg without using `npx`, which may attempt to install the package again and potential cause trouble with the packages installed by our package manager, `pnpm`.

tschaffter (changes requested):

hallieswan (approved):

tschaffter (approved):

PR #3299, opened 2025-07-14
feat(sage-monorepo): add Atlassian remote MCP server to `mcp.json` (SMR-267)

tschaffter:
## Description

Add Atlassian remote MCP server to `mcp.json` to extract information and manage resources in Jira and Confluence.

## Related Issue

- [SMR-267](https://sagebionetworks.jira.com/browse/SMR-267)

## How to install an MCP server

There are two ways to install a (remote) MCP server:

1. By adding its configuration directly to `.vscode/mcp.json`.
2. Using the Extension menu of VS Code
    - Open the Extension menu
    - Open the section "MCP Servers - Installed"
    - Place the cursor on this menu section to reveal the "globe" button named "Browse MCP Servers"
      - You are redirected to a web page that list MCP server.
    - Click on the "Install" button of the MCP server you want to install, e.g. GitHub or Atlassian
    - Back to VS Code, the MCP server will be listed as shown below

image

The preferred method is to define the MCP server in `mcp.json` so that developers don't have to install them themselves and for audit reasons.

> [!INFO]
> MCP servers that are configured directly in `mcp.json` will all have the MCP icon with just a name. The MCP servers added via the Extension menu will appears with a custom icon, a name, description and publishing organization.

image

## Preview

image

> [!TIP]
> The MCP server may not need to be manually started. In at least one occurrence, Copilot was able to start the server on its own when asked a question that triggered it to use the Atlassian MCP server. This was done after the server had been manually started and stopped once.

image

> [!WARNING]
> The two methods described above to install an MCP server (edit `mcp.json` and via the Extension menu) lead to the instantiation of different MCP server, even if their configuration is the same. For example, I first added GitHub via the Extension menu before using its metadata to add a corresponding configuration to `mcp.json`. This led to the creation of two MCP server instances for GitHub (74 tool each) triggered the error message reported below about Copilot's context being limited to 128 tools. 

## Copilot limits the number of MCP tools

image

[SMR-267]: https://sagebionetworks.jira.com/browse/SMR-267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.vscode/mcp.json
+++ b/.vscode/mcp.json
@@ -1,11 +1,9 @@
 {
   "servers": {
-    // https://www.atlassian.com/platform/remote-mcp-server
     "atlassian": {
       "type": "http",
       "url": "https://mcp.atlassian.com/v1/sse"
     },
-    // https://github.com/github/github-mcp-server/
     "github": {
       "type": "http",
       "url": "https://api.githubcopilot.com/mcp/"
--- a/.vscode/mcp.json
+++ b/.vscode/mcp.json
@@ -1,5 +1,9 @@
 {
   "servers": {
+    "atlassian": {
+      "type": "sse",
+      "url": "https://mcp.atlassian.com/v1/sse"
+    },
     "github": {
       "type": "http",
       "url": "https://api.githubcopilot.com/mcp/"
--- a/.vscode/mcp.json
+++ b/.vscode/mcp.json
@@ -1,9 +1,11 @@
 {
   "servers": {
+    // https://www.atlassian.com/platform/remote-mcp-server
     "atlassian": {
-      "type": "sse",
+      "type": "http",
       "url": "https://mcp.atlassian.com/v1/sse"
     },
+    // https://github.com/github/github-mcp-server/
     "github": {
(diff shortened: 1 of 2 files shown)


tschaffter:
I'm getting this error:

> Sorry, your request failed. Please try again. Request id: 1e9fede6-3041-448e-8582-07f29a977a9c
> Reason: You may not include more than 128 tools in your request.

.vscode/mcp.json
```diff
@@ -1,5 +1,9 @@
 {
   "servers": {
+    "atlassian": {
+      "type": "http",
```
tschaffter, line 4: The URL looks like the type should be "sse", but the type is "http" with the same URL when installing this MCP server via the Extensions menu.

hallieswan (approved):
Very cool, this will be so useful! I was able to use the Atlassian MCP server to fetch information about tickets that I've worked on recently.

PR #3305, opened 2025-07-16
chore(sage-monorepo): activate the July 2025 dev container

tschaffter:
## Changelog

- Activate the July 2025 dev container.
- Update pnpm.
- Remove the default value of the env var `NX_CI_EXECUTION_ENV`.
- Update `dev-env.sh` to unset empty `NX_` env vars.
- Pass the value `NX_CI_EXECUTION_ENV` to the dev container when run by a GitHub workflow.

## Preview

Demonstrate that the new utility function does not remove env vars that are set.

```console
$ export NX_CI_EXECUTION_ENV=test
$ workspace-unset-empty-nx-env-vars
$ echo $NX_CI_EXECUTION_ENV
test
```

### DIFF
--- a/.github/actions/setup-dev-container/action.yml
+++ b/.github/actions/setup-dev-container/action.yml
@@ -53,6 +53,7 @@ runs:
     - name: Start the dev container
       env:
         DEVCONTAINER_USER: ${{ inputs.devcontainer-user }}
+        NX_CI_EXECUTION_ENV: 'ubuntu-22.04'
       shell: bash
       run: |
         mkdir -p \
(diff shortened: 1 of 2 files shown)


.github/actions/setup-dev-container/action.yml
```diff
@@ -53,6 +53,7 @@ runs:
     - name: Start the dev container
       env:
         DEVCONTAINER_USER: ${{ inputs.devcontainer-user }}
+        NX_CI_EXECUTION_ENV: 'ubuntu-22.04'
```
tschaffter, line 56: The value should be set dynamically in the future.

hallieswan (approved):

PR #3316, opened 2025-07-18
feat(openchallenges): add Python script to fetch Kaggle challenges and index them to OC OpenSearch (SMR-285)

tschaffter:
## Related Issue

- [SMR-285](https://sagebionetworks.jira.com/browse/SMR-285)

## Changelog

- Add `kaggle` Python package to `openchallenges-tools`.
- Remove `click` Python package from `openchallenges-tools`.

### Manage Python dependencies in projects that use the shared venv

Add a package:

```bash
nx run openchallenges-tools:add kaggle==1.7.4.5
```

Remove a package:

```bash
nx run openchallenges-tools:remove --name click
```


tschaffter:
> Are there specific credentials I should be passing?

Did you create the config files and set `KAGGLE_USERNAME` and `KAGGLE_KEY`? I think you may have because the script check if you are authenticated and if not, a more user-friendly message should be displayed.

tschaffter:
> Are we not supposed to fill out the KAGGLE_USERNAME and KAGGLE_KEY with our Kaggle credentials? 

@vpchung That's right, but the error you shared is about not having the permission to index the data in OpenSearch. Indexing occurs after the script has collected data with Kaggle so it seems your Kaggle credentials are working fine.

> ...[trunated]
> opensearchpy.exceptions.AuthorizationException: AuthorizationException(403, 'index_create_block_exception', 'blocked by: [FORBIDDEN/10/cluster create-index blocked (api)];')

tschaffter:
Merging now as announced on Slack:

> I'm going to merge several PRs even though they haven't been reviewed, since we're all short on time. Instead of doing individual reviews, we can schedule a walkthrough of the changes together next week to validate everything.

PR #3325, opened 2025-07-22
fix(model-ad): load model details page for models with special characters in model name (MG-323)

hallieswan:
## Description

This PR adds URL encoding so the model details page will load for model names with special characters. 

## Related Issue

- [MG-323](https://sagebionetworks.jira.com/browse/MG-323)

## Changelog

- Adds custom URL serializer to encode special characters in model names (parentheses and forward slashes in parentheses)
- Adds fallback `/models/**` route to ensure that all models match the model details route, so the custom url serializer can encode special characters
- Adds e2e tests to ensure that model details pages load for models with special characters
- Fixes the tracking expression to use a unique set of fields in the model details modified genes `@for` loop

## Preview

`http://localhost:8000/models/5xFAD (UCI)`

https://github.com/user-attachments/assets/535911bf-78d8-4ca7-beb6-639061ab02e5

`http://localhost:8000/models/5xFAD (IU/Jax/Pitt)`

https://github.com/user-attachments/assets/a8ffe0c2-e408-40a2-9188-238fe02bb442

### DIFF
--- a/libs/model-ad/model-details/src/lib/components/model-details-hero/model-details-hero.component.html
+++ b/libs/model-ad/model-details/src/lib/components/model-details-hero/model-details-hero.component.html
@@ -8,7 +8,7 @@
 
     <h4>{{ `Modified Gene${model().genetic_info.length > 1 ? 's' : '' }`}}</h4>
     <div class="modified-genes">
-      @for (gene of model().genetic_info; track gene.ensembl_gene_id) {
+      @for (gene of model().genetic_info; track gene.ensembl_gene_id + gene.allele) {
         <div class="modified-gene">
           <p>
             {{ gene.modified_gene }} |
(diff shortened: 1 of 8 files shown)


libs/model-ad/model-details/src/lib/components/model-details-hero/model-details-hero.component.html
```diff
@@ -8,7 +8,7 @@ <h1>{{ model().model }}</h1>
 
     <h4>{{ `Modified Gene${model().genetic_info.length > 1 ? 's' : '' }`}}</h4>
     <div class="modified-genes">
-      @for (gene of model().genetic_info; track gene.ensembl_gene_id) {
+      @for (gene of model().genetic_info; track gene.ensembl_gene_id + gene.allele) {
```
hallieswan, line 11: The modified genes may have the same `ensembl_gene_id`, so track both `ensembl_gene_id` and `allele` to create a unique identifier.

sagely1 (approved):
LGTM!

PR #3351, opened 2025-07-31
feat(bixarena): enable Docker Stack to Use Locally Modified FastChat (SMR-316)

rrchai:
## Related Issue

[SMR-316](https://sagebionetworks.jira.com/browse/SMR-316)

## Preview

```
cd apps/bixarena/app
git clone --branch v0.2.36 --depth 1 https://github.com/lm-sys/FastChat.git
mv FastChat/* .
rm -rf .git
```


rrchai:
I have following errors by running `nx prepare bixarena-app` using the original [pyproject.toml](https://github.com/lm-sys/FastChat/blob/v0.2.36/pyproject.toml) for release tag `v0.2.36` and followed by the changes applied to fix the errors in the pr:
- error: `× Failed to build flash-attn==2.7.0.post2` 
  - ✅ downgrade python version to `3.11` as `requires-python = ">=3.8"` will install the latest python `3.13`
- error: `ModuleNotFoundError: No module named 'plotly'`
  - ✅ add `plotly` in `[model_worker]
- error: `No module named 'openai'`
  - ✅ add `openai` in `[model_worker]
- error: `module 'openai' has no attribute 'OpenAI'`
  - ✅ remove version limit `openai<1` in `[llm_judge]` and specify `openai>1` in `[model_worker]` as the attribute 'OpenAI'` only appears in `openai=1` or newer
  - it might break something in `[llm_judge]`, but we're not using the `[llm_judge]` extra dependencies for app

apps/bixarena/app/project.json
```diff
@@ -18,6 +18,16 @@
...
+    "install": {
+      "executor": "@nxlv/python:run-commands",
+      "options": {
+        "command": ["uv pip install --upgrade pip", "uv pip install -e '.[model_worker,webui]'"],
+        "cwd": "{projectRoot}",
+        "silent": false,
+        "verbose": false,
+        "debug": false
+      }
+    },
```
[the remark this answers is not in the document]
rrchai: this change is outdated and not included in the latest pr :)

tschaffter:
@rrchai The PR does not describe what the controller is. Also following our discussion, does that mean that the controller is needed or will we remove it in the future?

apps/bixarena/app/src/assets/demo_narrow.gif
tschaffter, line 1: We have to be very conservative when it comes to add "large" binary files to the monorepo to ensure that `git` commands remain fast as the monorepo continue growing.

- All asset files that are larger than 100 kb should be removed.
- Consider removing the docs folder.

apps/bixarena/app/entrypoint.sh
tschaffter, line 1: Is there a reason this entrypoint file differ from the entrypoint file used in other Python projects in the monorepo?

apps/bixarena/app/src/assets/demo_narrow.gif
tschaffter: Consider adding these files to the project `.gitignore` so that you can keep the files in your local environment but also prevent them from being tracked again with git in the future, e.g. if we clone fastchat again.

apps/bixarena/app/entrypoint.sh
rrchai: The other entrypoint file like what in `agora-data`:

```sh
#!/usr/bin/env bash
set -e

if [ "$1" = 'python' ]; then
    cd ${APP_DIR}
    exec gosu "$APP_USERNAME" "$@"
fi

exec "$@"
```

This setup requires running a Python command after the `ENTRYPOINT` in the `Dockerfile`, otherwise the container exits immediately due to the empty arg in `exec "$@"`.

I initially tried specifying `python -m fastchat.serve.gradio_web_server_multi ...` as the `CMD` after the `ENTRYPOINT` script, but it seemed to fail to read environment variables - possibly due to quote escaping issues. Another way is to make a new python script `main.py` to call the `gradio_web_server_multi`.

To keep things simple, I ended up moving the `python -m fastchat.serve.gradio_web_server_multi ...` command directly into the entrypoint script as well.

tschaffter: > exec "$@".

@rrchai This is meant to allow us to easily perform other commands in the container, e.g. `docker exec -it {container} bash`, in particular to access the container as the `root` user.

Your current entrypoint file runs Python as the `root` user, which is not a good practice. Please use the approach we use in other projects. To solve the "hydration" of the environment variables, you could define the entire command in the entry point file: `exec gosu "$APP_USERNAME" python -m fastchat.serve.gradio_web_server_multi ...`

apps/bixarena/app/src/fastchat/serve/example_images/city.jpeg
tschaffter, line 1: Remove this image

tschaffter:
@rrchai PR titles should not be capitalized

tschaffter (approved):

PR #3386, opened 2025-08-14
feat(model-ad): add table links (MG-355)

sagely1:
## Description
The comparison tool tables have internal/external links.  If no url is present in the data, don't show the link.

## Related Issue

[MG-355](https://sagebionetworks.jira.com/browse/MG-355?assignee=633b138697148a8301fe2c72&selectedIssue=MG-251)

## Changelog
- Update logic to only show links when link_url is present in the data.
- Internal links in the data are not always prepended with a slash so add a leading slash as needed.
- External link behavior did not require any changes.

## Preview
Before:
image

After:
image

### DIFF
--- a/libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.html
+++ b/libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.html
@@ -1,9 +1,11 @@
-<div class="comparison-tool-table-link">
-  @if (isExternalLink()) {
-    <a target="_blank" href="{{ linkUrl() }}"
-      >{{ linkText() }} <i class="pi pi-external-link"></i
-    ></a>
-  } @else {
-    <a routerLink="{{ linkUrl() }}">{{ linkText() }} <i class="pi pi-external-link"></i></a>
-  }
-</div>
+@if (linkUrl()) {
+  <div class="comparison-tool-table-link">
+    @if (isExternalLink()) {
+      <a target="_blank" href="{{ linkUrl() }}"
+        >{{ linkText() }} <i class="pi pi-external-link"></i
+      ></a>
+    } @else {
+      <a routerLink="{{ internalLinkUrl }}">{{ linkText() }} <i class="pi pi-external-link"></i></a>
+    }
+  </div>
+}
--- a/libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.ts
+++ b/libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.ts
@@ -12,7 +12,20 @@ export class ComparisonToolTableLinkComponent {
   linkText = input<string>('');
   linkUrl = input<string>('');
 
-  isExternalLink() {
-    return isExternalLink(this.linkUrl());
+  get internalLinkUrl(): string {
+    const url = this.linkUrl().trim();
+
+    // Return URL as-is if it already starts with '/'
+    if (url.startsWith('/')) {
+      return url;
+    }
+
+    // Prepend '/' to make it a proper internal route
+    return `/${url}`;
+  }
+
+  isExternalLink(): boolean {
+    const url = this.linkUrl()?.trim();
+    return url ? isExternalLink(url) : false;
   }
 }
--- a/libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.spec.ts
+++ b/libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.spec.ts
@@ -23,5 +23,16 @@ describe('ComparisonToolTableLinkComponent', () => {
     await setup(linkText, linkUrl);
     const link = screen.getByRole('link', { name: /My Link/i });
     expect(link).toBeInTheDocument();
+    expect(link).toHaveAttribute('href', linkUrl);
+  });
+
+  it('should render internal link without leading slash character with correct text and url', async () => {
+    const linkText = 'My Link';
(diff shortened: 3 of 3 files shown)


libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.ts
```diff
@@ -12,7 +12,20 @@ export class ComparisonToolTableLinkComponent {
...
+    const url = this.linkUrl().trim();
+
+    // Return URL as-is if it already starts with '/'
+    if (url.startsWith('/')) {
+      return url;
+    }
+
+    // Prepend '/' to make it a proper internal route
+    return `/${url}`;
+  }
```
hallieswan, line 25: Could consider using the `computed` signal here instead, since we're using `input` signals above, e.g. something like this: 

```suggestion
  internalLinkUrl = computed(() => {
    const url = this.linkUrl().trim();

    // Return URL as-is if it already starts with '/'
    if (url.startsWith('/')) {
      return url;
    }

    // Prepend '/' to make it a proper internal route
    return `/${url}`;
  });
```

libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.ts
```diff
@@ -12,7 +12,20 @@ export class ComparisonToolTableLinkComponent {
...
+
+  isExternalLink(): boolean {
+    const url = this.linkUrl()?.trim();
+    return url ? isExternalLink(url) : false;
```
hallieswan, line 29: Would it make sense to move the `trim` into the shared `isExternalLink` function and just call the shared function directly? If not, could we rename the function in this component to improve readability?

libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.html
```diff
@@ -1,9 +1,11 @@
...
-  }
-</div>
+@if (linkUrl()) {
+  <div class="comparison-tool-table-link">
+    @if (isExternalLink()) {
+      <a target="_blank" href="{{ linkUrl() }}"
+        >{{ linkText() }} <i class="pi pi-external-link"></i
+      ></a>
+    } @else {
+      <a routerLink="{{ internalLinkUrl }}">{{ linkText() }} <i class="pi pi-external-link"></i></a>
```
hallieswan, line 8: The Angular router performs automatic URL encoding, so `?` and `=` are encoded and the query parameters aren't passed along as expected. For example, this Gene Correlation link `http://localhost:4200/comparison/expression?model=3xTg-AD` is encoded as `http://localhost:4200/comparison/expression%3Fmodel%3D5xFAD%20%28UCI%29`, which breaks the link. Were you planning to fix in a future PR?

hallieswan (commented):
A few small suggestions and one small issue (that you might already be planning to fix for a future PR)!

libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.ts
sagely1: Good idea.  Fixed in next commit.

libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.ts
sagely1: Good idea.  Fixed in next commit.

libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.html
sagely1: fixed in next commit!

libs/explorers/comparison-tools/src/lib/comparison-tool-table-link/comparison-tool-table-link.component.ts
```diff
@@ -12,20 +12,37 @@ export class ComparisonToolTableLinkComponent {
...
+    if (urlParts.length > 1) {
+      const paramsString = urlParts[1];
+      paramsString.split('&').forEach((param) => {
+        const paramParts = param.split('=');
+        // Ensure there is a key
+        if (paramParts[0]) {
+          const key = decodeURIComponent(paramParts[0]);
+          const value = paramParts.length > 1 ? decodeURIComponent(paramParts[1]) : true;
+          queryParams[key] = value;
+        }
```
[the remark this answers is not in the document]
sagely1: I did give that a try but it is not available because of SSR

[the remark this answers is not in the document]
hallieswan: Oh interesting! From what I've seen, `URLSearchParams` is [supported in Node.js](https://nodejs.org/api/url.html#class-urlsearchparams), so I would expect the function to work in SSR. Also, since `URLSearchParams` is a built-in API in both browsers and Node.js, we likely wouldn’t need to add any new dependencies to use it. We use it [throughout](https://github.com/search?q=repo%3ASage-Bionetworks%2Fsage-monorepo%20URLSearchParams&type=code) `explorers`, `agora` and `model-ad`.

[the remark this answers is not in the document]
sagely1: You were right.  I was getting linting errors for some reason when attempting to use this.  Rebuilding the dev container seem to have fixed this.

hallieswan (approved):
Looks great!

PR #3387, opened 2025-08-15
feat(sage-monorepo): format Python files with Ruff (SMR-394)

tschaffter:
## Description

Format Python files with Ruff instead of Black.

## Notes

A subsequent PR will format all Python files in the monorepo using Ruff. This step is intentionally omitted here so that this PR can serve as a focused reference for the changes required to migrate from Black to Ruff. Formatting with Ruff is not expected to introduce many changes, as it is advertised as a drop-in replacement for Black and we have not customized its configuration.

## Related Issue

- [SMR-394](https://sagebionetworks.jira.com/browse/SMR-394)

## Changelog

- Add Ruff VS Code extension to the dev container and remove Black's.
- Update VS Code settings to enable format on save with Ruff.
- Configure lint-staged to format staged Python files with Ruff.
- Update Ruff Python package, pin its version and set minimal configuration in the workspace-level `pyproject.toml`.
- Remove Black Python package and its config from the workspace-level `pyproject.toml`.

## Preview

image

✅ I confirm that Python files are automatically formatted with Ruff when saving them.

[SMR-394]: https://sagebionetworks.jira.com/browse/SMR-394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rrchai (approved):
LGTM

pyproject.toml
```diff
@@ -45,12 +67,11 @@ package-mode = false
 
 [dependency-groups]
 dev = [
-  "black==24.8.0",
   "sqlfluff==3.2.0",
-  "ruff>=0.8.2",
+  "ruff==0.12.9",
```
BryanFauble, line 71: Is there something preventing us from doing `ruff>=0.12.9` here?

BryanFauble (approved):
Just one comment around version ranges, otherwise this looks good to me.

pyproject.toml
tschaffter: We pin version numbers in the monorepo to improve reproducibility and to provide greater transparency regarding the versions installed. I also find it simpler to always pin version numbers rather than deciding case by case which strategy to use in different scenarios. This approach has worked well for our projects in TypeScript and Java.

In Python, however, there are additional constraints that can affect dependency resolution, so a strict pinning strategy may not be as effective or straightforward. We can address this challenge when we reach that stage.

I would also welcome a side-by-side comparison of the two strategies for Python, strict pinning vs. more flexible version ranges, to serve as a reference for making an informed decision.

andrewelamb: I think for dev dependencies (things like ruff, black, mypy etc.) this is fine. But when the project is a package(Schematic, SynapsePythonClient) that others use, we want to be more flexible and provide the widest range possible.

PR #3421, opened 2025-08-25
feat(synapse): add automated Synapse API update workflow

tschaffter:
## Description

This PR adds a new GitHub workflow that automatically monitors and updates the Synapse OpenAPI specification and regenerates dependent client libraries when changes are detected.

The workflow runs daily to check for changes in the Synapse OpenAPI specification and automatically updates the API description and regenerates dependent clients to ensure they stay synchronized with upstream changes.

## Related Issue

This addresses the need to keep Synapse API clients up-to-date with the latest API specification changes without manual intervention.

## Changelog

- Add `.github/workflows/update-synapse-api.yml` workflow
- Implements daily automated checking of Synapse OpenAPI specification
- Uses `nx get synapse-api-description` to download latest spec
- Uses git diff for efficient change detection
- Automatically regenerates all synapse-* client libraries when changes detected
- Creates automated PRs with comprehensive descriptions
- Includes manual trigger option with force update capability
- Provides detailed logging and error handling

## Preview

The workflow will:
1. Run daily at 9 AM UTC
2. Download the latest OpenAPI spec from https://rest-docs.synapse.org/rest/openapi/openapispecification.json
3. Compare with existing specification using git diff
4. If changes detected:
   - Update API description files
   - Regenerate Angular API client and other synapse-* projects
   - Create a PR with all changes for review
5. Include comprehensive logging and change summaries

This ensures that Synapse API clients automatically stay in sync with API changes without manual intervention.

### DIFF
--- a/.github/workflows/update-synapse-api.yml
+++ b/.github/workflows/update-synapse-api.yml
@@ -1,0 +1,200 @@
+name: Update Synapse API Description
+
+on:
+  schedule:
+    # Run daily at 9 AM UTC (4 AM EST, 1 AM PST)
+    - cron: '0 9 * * *'
+  workflow_dispatch:
+    # Allow manual trigger
+    inputs:
+      force_update:
+        description: 'Force update even if no changes detected'
+        required: false
+        default: false
+        type: boolean
+
+env:
+  NX_BRANCH: ${{ github.ref_name }}
+  NX_CLOUD_ENCRYPTION_KEY: ${{ secrets.NX_CLOUD_ENCRYPTION_KEY }}
+
+jobs:
+  update-synapse-api:
+    runs-on: ubuntu-22.04-4core-16GBRAM-150GBSSD
+    env:
+      NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN_READ_WRITE }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          # Fetch full history for proper comparison
+          fetch-depth: 0
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up the dev container
+        uses: ./.github/actions/setup-dev-container
+
+      - name: Download and check for API changes
+        id: check_changes
+        run: |
+          devcontainer exec --workspace-folder ../sage-monorepo bash -c ". ./dev-env.sh \
+            && echo '🔄 Downloading latest OpenAPI specification...' \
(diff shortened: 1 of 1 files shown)


.github/workflows/update-synapse-api.yml
```diff
@@ -0,0 +1,193 @@
...
+        if: steps.check_changes.outputs.has_changes == 'true' && steps.check_generated_changes.outputs.has_generated_changes == 'true'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Configure git
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+          # Create a new branch with timestamp
+          BRANCH_NAME="synapse/update-synapse-api-$(date +%Y%m%d-%H%M%S)"
```
hallieswan, line 103: Since we call `date` to create each timestamp, the timestamps will be slightly different (e.g. branch name will have a slightly different timestamp than the commit message). If we want the timestamps to match, then we could consider storing the date in a variable then formatting as needed. However, the differences will probably be negligible, so up to you!

.github/workflows/update-synapse-api.yml
```diff
@@ -0,0 +1,193 @@
...
+          COMMIT_MSG="chore(synapse): update Synapse API description and regenerate clients
+
+          This is an automated update triggered by changes detected in the Synapse OpenAPI specification.
+
+          Updated components:
+          - libs/synapse/api-description/openapi/openapi.json (source specification)
+          - libs/synapse/api-description/openapi/openapi.yaml (bundled specification)
+          - Regenerated synapse-* client libraries
+
+          Source: https://rest-docs.synapse.org/rest/openapi/openapispecification.json
```
hallieswan, line 119: We could consider storing this path in a variable, since we use it in several places in this step

hallieswan (approved):
Very cool! This looks great, just a couple suggestions!

.github/workflows/update-synapse-api.yml
tschaffter: I actually prefer to not include the date to prevent the PR to open duplicates. A constant head branch lets the workflow force-push new commits and keeps a single rolling PR open until merged.

.github/workflows/update-synapse-api.yml
tschaffter: See previous comment. I removed the date variables. Git already captures the timestamp of every events.

PR #3425, opened 2025-08-26
chore(sage-monorepo): make the pre-commit hook more snappy

tschaffter:
## Description

Keep hooks fast and let CI be the gate for tests.

## Changelog

- No longer run tests on affected projects during during pre-commit.
- Document recommendations on which processes should run during pre-commit and pre-push.


hallieswan (approved):
Looks good to me! This will definitely speed up local development

PR #3435, opened 2025-09-03
feat(model-ad): align style of "no data" boxplot placeholder with design, fix bug where legend is not displayed when last boxplot doesn't have data (MG-317)

hallieswan:
## Description

The Model-AD model details data was recently updated to include placeholders for ages that don't have data, but should still be displayed in the boxplots grid. Here, we align the style of the "no data" boxplot placeholder with the design. We also fix a bug where the legend is not displayed when the last boxplot doesn't have data by creating a new legend component that is displayed in a separate row below the boxplots.

## Related Issue

[MG-317](https://sagebionetworks.jira.com/browse/MG-317)

## Changelog

- Adds style options for "no data" boxplot placeholder -- `textOnly` to maintain style for Agora and `grayBackground` for the Model-AD design
- Adds new legend chart that only displays a scatterplot legend
- Moves the boxplots grid legend from the last plot in the grid to a separate row to fix bug where legend is not displayed when the last plot doesn't have data
- Increases the offset between Female and Male categories to increase visual separation between the points on the plots and align with the design
- Moves the boxplot point style definition to constants

## Preview

`https://dev.modeladexplorer.org/models/Abca7*V1599M/biomarkers` | `http://localhost:8000/models/Abca7*V1599M/biomarkers` 
:---: | :---:
dev_dataUnavailable_noLegend | MG-317_dataUnavailable_newLegend

### DIFF
--- a/libs/model-ad/model-details/src/lib/components/model-details-boxplots-grid/model-details-boxplots-grid.component.ts
+++ b/libs/model-ad/model-details/src/lib/components/model-details-boxplots-grid/model-details-boxplots-grid.component.ts
@@ -1,39 +1,23 @@
-import { BreakpointObserver } from '@angular/cdk/layout';
-import { Component, computed, inject, input } from '@angular/core';
-import { toSignal } from '@angular/core/rxjs-interop';
-import { BreakpointConfigService } from '@sagebionetworks/explorers/services';
+import { Component, computed, input } from '@angular/core';
 import { IndividualData, ModelData } from '@sagebionetworks/model-ad/api-client-angular';
+import { MODEL_DETAILS_BOXPLOT_POINT_STYLES } from '@sagebionetworks/model-ad/config';
+import { LegendDirective } from '@sagebionetworks/shared/charts-angular';
 import { ModelDetailsBoxplotComponent } from '../model-details-boxplot/model-details-boxplot.component';
 
 @Component({
   selector: 'model-ad-model-details-boxplots-grid',
-  imports: [ModelDetailsBoxplotComponent],
+  imports: [ModelDetailsBoxplotComponent, LegendDirective],
   templateUrl: './model-details-boxplots-grid.component.html',
   styleUrls: ['./model-details-boxplots-grid.component.scss'],
 })
 export class ModelDetailsBoxplotsGridComponent {
-  private readonly breakpointObserver = inject(BreakpointObserver);
-  private readonly breakpointConfigService = inject(BreakpointConfigService);
-
   modelDataList = input.required<ModelData[]>();
   genotypeOrder = input<string[] | undefined>();
   sexes = input.required<IndividualData.SexEnum[]>();
 
-  private readonly observeLargeScreen = toSignal(
-    this.breakpointObserver.observe(`(min-width: ${this.breakpointConfigService.largeBreakpoint})`),
-    { initialValue: { matches: false, breakpoints: {} } },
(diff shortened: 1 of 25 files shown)


libs/model-ad/model-details/src/lib/components/model-details-boxplots-grid/model-details-boxplots-grid.component.ts
```diff
@@ -1,39 +1,23 @@
...
-      return totalPlots - 1;
-    } else {
-      // 2-column layout: legend on leftmost of last row
-      return totalPlots % 2 === 0 ? totalPlots - 2 : totalPlots - 1;
-    }
+  pointStyles = computed(() => {
+    return MODEL_DETAILS_BOXPLOT_POINT_STYLES.filter((pointStyle) =>
+      this.sexes().includes(pointStyle.label as IndividualData.SexEnum),
+    );
   });
```
hallieswan, line 38: Since the legend is always shown in a separate row beneath the boxplots, we no longer need to track the screen size to determine which boxplot within the grid needs to display the legend

libs/shared/typescript/charts/src/lib/boxplot-chart/boxplot-chart.ts
```diff
@@ -95,7 +96,7 @@ export class BoxplotChart {
       points,
       xAxisCategories,
       pointCategories,
-      0.1,
+      0.3,
```
hallieswan, line 99: Increase default separation between categories to make it easier to view points from each category

libs/shared/typescript/charts/src/lib/legend-chart/legend-chart.ts
```diff
@@ -0,0 +1,52 @@
...
+        itemWidth: DEFAULT_POINT_SIZE,
+        selectedMode: false,
+      },
+      xAxis: { show: false },
+      yAxis: { show: false },
+      series: pointStyles.map((item) => ({
+        name: item.label,
+        type: 'scatter',
+        symbol: item.shape,
+        data: [[0, 0]], // single dummy point
```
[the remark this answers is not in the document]
hallieswan: Using `[0, 0]` directly is appropriate in this case because it clearly represents the origin point, and defining it as a constant may add unnecessary complexity without significantly improving readability or maintainability.

libs/shared/typescript/charts/src/lib/boxplot-chart/boxplot-chart.ts
```diff
@@ -26,10 +27,12 @@ const boxplotStyle = {
 
 const yAxisPadding = 0.2;
 const defaultPointShape = 'circle';
-const defaultPointSize = 18;
 const defaultPointColor = '#8b8ad1';
 const defaultPointOpacity = 0.8;
 
+const defaultPointCategoryOffset = 0.3;
```
hallieswan, line 33: Increase default separation between categories to make it easier to view points from each category

tschaffter:
@hallieswan I'm assuming that all the screenshot show the current implementation since there is no reference to Before and After/Now.

- The gray box instead of a block looks a bit like a broken image. Using a transparent background would address this issue.
- We could replace the following two sentences by the same one
  - Data unavailable (use in the first row of screenshots)
  - No data is currently available (used in the second row of screenshots)

sagely1 (approved):
LGTM!

PR #3448, opened 2025-09-10
feat(model-ad): scroll to panel content when tab specified in url, delay scroll when changing between tabs to give content time to load (MG-390)

hallieswan:
## Description

The model details page should scroll to the panel nav element when a tab is specified in the URL. If no tab is specified in the URL, then the page should not scroll.

## Related Issue

[MG-390](https://sagebionetworks.jira.com/browse/MG-390) 

## Changelog

- Scrolls to panel nav element when a tab is specified in the URL
- Delays scroll when changing between tabs to give content time to load so panel nav element is positioned correctly
- Does not scroll when URL includes a hash fragment; expect the scroll to be handled by a component within the panel content (e.g. scrolling to the Nfl section of the biomarkers plots), so we should not scroll to the panel nav element
- Moves `getHashFragment` to explorers HelperService

## Preview

Scrolls to panel nav element when a tab is specified in the URL: 

https://github.com/user-attachments/assets/7b81b8fb-b74a-4094-a745-a4a6f47450d8

Does not scroll to panel nav element when tab is not specified in URL and keeps panel nav element at top of screen when switching tabs: 

https://github.com/user-attachments/assets/c811d186-92da-4d61-aa3a-11836d30f7a1

### DIFF
--- a/libs/explorers/ui/src/lib/components/panel-navigation/panel-navigation.component.ts
+++ b/libs/explorers/ui/src/lib/components/panel-navigation/panel-navigation.component.ts
@@ -28,6 +28,7 @@ export class PanelNavigationComponent implements AfterViewInit, AfterViewChecked
   panels = input.required<Panel[]>();
   activePanel = input.required<string>();
   activeParent = input<string>('');
+  scrollToPanelNavElementOnInitialLoad = input<boolean>(false);
   panelChange = output<Panel>();
 
   faAngleRight = faAngleRight;
@@ -80,6 +81,10 @@ export class PanelNavigationComponent implements AfterViewInit, AfterViewChecked
     setTimeout(() => {
       this.onWindowResize();
     }, 100);
+
+    if (this.scrollToPanelNavElementOnInitialLoad()) {
+      this.scrollToPanelNavElement();
+    }
   }
 
   ngAfterViewChecked() {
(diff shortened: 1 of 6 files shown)


libs/explorers/ui/src/lib/components/panel-navigation/panel-navigation.component.ts
```diff
@@ -88,13 +93,23 @@ export class PanelNavigationComponent implements AfterViewInit, AfterViewChecked
 
   activatePanel(panel: Panel) {
     if (isPlatformBrowser(this.platformId)) {
+      this.panelChange.emit(panel);
+      this.scrollToPanelNavElement();
```
hallieswan, line 97: Notify the parent component that panel changed, then scroll so the parent has time to start loading the panel content and we scroll to the correct position

libs/model-ad/model-details/src/lib/model-details.component.ts
```diff
@@ -114,9 +116,11 @@ export class ModelDetailsComponent implements OnInit, AfterViewInit {
   }
 
   private setActivePanelAndParentFromUrl(params: ParamMap) {
+    const noHashFragment = this.helperService.getHashFragment() === '';
     if (params.get('subtab')) {
       this.activePanel = params.get('subtab') as string;
       this.activeParent = params.get('tab') as string;
+      this.scrollToPanelNavElementOnInitialLoad = noHashFragment; // selector will handle to scroll if hash fragment is present
```
hallieswan, line 123: For example, `https://dev.modeladexplorer.org/models/Abca7*V1599M/biomarkers#nfl` should scroll to the Nfl biomarkers plots, not the panel nav element.

PR #3469, opened 2025-09-16
chore: replace OpenAPI extension `x-audience` by `x-internal`

tschaffter:
## Related Issue

- [SMR-465](https://sagebionetworks.jira.com/browse/SMR-465)

## Changelog

- Replace OpenAPI vendor extension `x-audience: string[]` to `x-internal: boolean`.
- Remove `x-audience: [public]` to simplify OpenAPI files because it's the same as not including this annotation.
- Update Redocly config files to remove nodes annotated with `x-internal: true`.


tschaffter:
Cc: @hallieswan

PR #3479, opened 2025-09-17
fix(model-ad): conditionally display omics tab, fallback to default tab when invalid or disabled tab in URL (MG-282, MG-415)

hallieswan:
## Description

We would like to hide the omics tab from the model details page when there is no omics data. We would also like to gracefully handle the case where the URL contains a tab that does not exist or is disabled for the current model.

## Related Issue

- [MG-282](https://sagebionetworks.jira.com/browse/MG-282)
- [MG-415](https://sagebionetworks.jira.com/browse/MG-415)

## Changelog

- Adds conditional disabling of the Omics tab
- Implements fallback logic to redirect users from invalid/disabled tabs to the first available tab and update the URL
- Changes the default activePanel from 'omics' to empty string to ensure proper fallback behavior instead of assuming omics is always available
- Moves URL parsing from ngOnInit to inside the HTTP response handler so panel validation occurs after model data is loaded

## Preview

Start the app: `model-ad-build-images && model-ad-docker-start`
Connect to the model-ad-mongo db
Add a test model without omics data: 

```
// Get an existing model
let testModel = db.model_details.findOne({ name: "LOAD1" });

// Remove the _id field so MongoDB will generate a new one
delete testModel._id;

// Update the specific fields
testModel.name = "MG-282-test";
testModel.gene_expression = null;
testModel.disease_correlation = null;

// Insert the new model
db.model_details.insertOne(testModel);
```

Navigate to the test model without omics data: 

- Defaults to first tab with data: "Resources"
- Does not scroll
- URL is unchanged

https://github.com/user-attachments/assets/4f6a2961-0c76-478b-b21b-b7ded805b8fc

Navigate to a tab that is disabled for a model:

- Defaults to first tab with data: "Omics"
- Does not scroll
- URL corrected to remove the incorrect tab

https://github.com/user-attachments/assets/9d967da4-64c4-4230-8d45-8af0f15134d6

[MG-282]: https://sagebionetworks.jira.com/browse/MG-282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MG-415]: https://sagebionetworks.jira.com/browse/MG-415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/model-ad/model-details/src/lib/model-details.component.ts
+++ b/libs/model-ad/model-details/src/lib/model-details.component.ts
@@ -65,14 +65,15 @@ export class ModelDetailsComponent implements OnInit, AfterViewInit {
     },
   ];
 
-  activePanel = 'omics';
+  activePanel = '';
   activeParent = '';
 
+  maybeScrollToPanelNavElementOnInitialLoad = false;
   scrollToPanelNavElementOnInitialLoad = false;
 
   reset() {
     this.model = undefined;
-    this.activePanel = 'omics';
+    this.activePanel = '';
     this.activeParent = '';
     this.isLoading = true;
   }
@@ -84,7 +85,6 @@ export class ModelDetailsComponent implements OnInit, AfterViewInit {
       // only fetch data during client hydration
       if (this.platformService.isBrowser) {
         this.loadPanelData(params);
-        this.setActivePanelAndParentFromUrl(params);
       }
     });
   }
@@ -98,7 +98,11 @@ export class ModelDetailsComponent implements OnInit, AfterViewInit {
         .subscribe({
           next: (model: Model) => {
             this.model = model;
+            this.setActivePanelAndParentFromUrl(params);
             this.updatePanelDisabledState();
+            this.changePanelAndUrlIfInitialActivePanelIsInvalid();
+            this.scrollToPanelNavElementOnInitialLoad =
+              this.maybeScrollToPanelNavElementOnInitialLoad;
             this.isLoading = false;
           },
           error: (error) => {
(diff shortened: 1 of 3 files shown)


libs/model-ad/model-details/src/lib/model-details.component.ts
```diff
@@ -65,14 +65,15 @@ export class ModelDetailsComponent implements OnInit, AfterViewInit {
     },
   ];
 
-  activePanel = 'omics';
+  activePanel = '';
```
hallieswan, line 68: The omics tab will not always be available, so set to an empty string so the proper fallback logic is triggered

libs/model-ad/model-details/src/lib/model-details.component.ts
```diff
@@ -116,6 +120,12 @@ export class ModelDetailsComponent implements OnInit, AfterViewInit {
         p.disabled = true;
       } else if (p.name === 'pathology' && this.model?.pathology.length === 0) {
         p.disabled = true;
+      } else if (
+        p.name === 'omics' &&
+        this.model?.gene_expression === null &&
+        this.model?.disease_correlation === null
+      ) {
+        p.disabled = true;
```
hallieswan, line 128: Disable the omics tab if data is unavailable

libs/model-ad/model-details/src/lib/model-details.component.ts
```diff
@@ -127,7 +137,7 @@ export class ModelDetailsComponent implements OnInit, AfterViewInit {
     if (params.get('subtab')) {
       this.activePanel = params.get('subtab') as string;
       this.activeParent = params.get('tab') as string;
-      this.scrollToPanelNavElementOnInitialLoad = noHashFragment; // selector will handle to scroll if hash fragment is present
+      this.maybeScrollToPanelNavElementOnInitialLoad = noHashFragment; // selector will handle to scroll if hash fragment is present
```
hallieswan, line 140: Use `maybeScrollToPanelNavElementOnInitialLoad` because the selected panel may be invalid or disabled, in which case, the panel will be updated by `changePanelAndUrlIfInitialActivePanelIsInvalid` and we should not scroll.

PR #3482, opened 2025-09-18
feat(bixarena): add user page and support synapse login (SMR-474)

rrchai:
## Description

This PR implements Synapse OAuth authentication flow, enabling users to login with their Synapse accounts. 

Auth flow:
- User clicks login → Redirected to Synapse authorization server and login page
- User authorizes → Synapse redirects back with an authorization code (e.g. `?code=abc123`) in URL
- Exchanges the authorization code to auth server for access token and store in the session
- Clean up parameters from URL after callback
- Use access token to make Synapse API calls (e.g. `Get /userProfile`)

> [!Note]
> The app only stores access tokens in-memory per session and tokens are lost after browser refresh. No refresh token mechanism is implemented to maintain persistent login state, ~~so users must re-authenticate when refreshing the page~~. The cookie is store in the client-site using javascript with 2 hr life.

## Related Issue

[SMR-474](https://sagebionetworks.jira.com/browse/SMR-474)

## Changelog

- Implement user page with login/logout functionality and proper navigation flow
- Add authentication system to support login with a synapse account
- Add development authentication bypass using `SKIP_AUTH` environment variable

## Preview

> [!Note]
> A [OAuth Client](https://help.synapse.org/docs/Using-Synapse-as-an-OAuth-Server.2048327904.html) credential is required to set in the environment variable. I will share a temporary client for testing pr via lastPass and create a client for bixarena later.

https://www.loom.com/share/d2a12af293a643cbb0b2717f4adc5c97?sid=64ff6d1b-4818-48e3-b859-25e81b0e8a0e

https://www.loom.com/share/e69a3974fbe64074bbcf85c6a696ddfb?sid=c3ba4061-640a-4d5a-bb1e-cd6b526e5763

[SMR-474]: https://sagebionetworks.jira.com/browse/SMR-474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tschaffter (changes requested):
> No refresh token mechanism is implemented to maintain persistent login state, so users must re-authenticate when refreshing the page.

This is a strong limitation that needs to be addressed now. Prefer a short-lived `httpOnly`, `sameSite=strict` cookie session backed by a server session.

apps/bixarena/app/src/main.py
```diff
@@ -1,113 +1,198 @@
...
+        if not _validate_oauth_params(code, state):
+            return update_login_button(), *update_user_page(), gr.HTML("")
+
+        success, session_id = auth_service.handle_oauth_callback(code, state)
+        if success and session_id:
+            # Set cookie with appropriate security for environment (2 hours)
+            secure_flag = "secure" if str(request.url).startswith("https") else ""
+            cookie_script = f"""
+            <script>
+            document.cookie = "bixarena_session={session_id}; path=/; max-age=7200; samesite=strict; {secure_flag}";
```
rrchai, line 69: The Gradio does not natively support the `httponly` for storing cookie, as it’s just a client server.

Current implementation uses JavaScript to set cookies, which is vulnerable to XSS attacks even with `samesite=strict`. To enable with `httponly` attribute, we need to add a FastAPI layer that handles cookie setting server-side, like:

```python
secure_flag = True if request.url.scheme == "https" else False
response.set_cookie(
        key="bixarena_session",
        value=session_id,
        max_age=7200,  
        httponly=True,  
        secure=secure_flag,  
        samesite="strict"  
}
```

As our auth service will be using spring-boot, plus we will switch to other frontend eventually, e.g. Angular, we might not need FastAPI for auth and routing. 
We will pause the exploration using Gradio alone to achieve secured authentication with cookie management. cc @tschaffter

apps/bixarena/app/.env.example
```diff
@@ -1,2 +1,7 @@
 APP_PORT=8100
-OPENAI_API_KEY=
+APP_REDIRECT_URI=http://127.0.0.1:8100
```
tschaffter, line 2: If related to auth, consider renaming it to `AUTH_REDICTED_URI` or similar.

tschaffter (approved):
Just one comment other looks good.

PR #3486, opened 2025-09-19
feat(openchallenges): make the image service zero-config (SMR-478)

tschaffter:
## Description

> [!IMPORTANT]
> The image service returns URLs for images hosted by the Thumbor server, rather than the images themselves.

The goal of this PR is to enable local deployments of OpenChallenges without requiring credentials to access a private S3 bucket. As a trade-off, production images will not be available. Instead, placeholder images are displayed in the web app when running in this mode.

This PR updates the OC image service to support returning placeholder images from a third-party provider during development. When placeholders are enabled, the image service returns URLs that point directly to the external provider, bypassing Thumbor.

Example of image URL returned by the image service when `app.placeholder.enabled` is set to `true`.

```json
{
  "url": "https://images.placeholders.dev/250x250?fontWeight=400"
}
```

In a previous PR, the default Thumbor configuration was updated to use its HTTP loader instead of the S3 loader, allowing image URLs from the service to be processed by Thumbor. However, this dual-configuration (HTTP vs. S3 loader) is unnecessary if placeholders are served directly from the external provider. This change lets us revert to a single Thumbor configuration.

Note that the Thumbor container starts successfully even if its configuration file lacks S3 credentials. In such cases, attempts to fetch images from S3 simply result in 500 errors.

> [!NOTE]
> Thumbor supports only one image loader at at time. Thus, it can not use both HTTP and S3 loader at the same time, e.g. to support both object key and HTTP URLs.

The config file `application-dev.yml` includes new config properties to enable the placeholder mode. In that mode, the generated image URLs will be served directly by the placeholder service defined by `app.placeholder.url-template`.

## Related Issue

- [SMR-478](https://sagebionetworks.jira.com/browse/SMR-478)

## Changelog

- Revert back to a single, AWS-specific config file.
- Add new `AppProperties` class that contains the app-specific properties.
- Update the image service to support placeholder image URLs (does not use Thumbor).

### Home Page

image

### Organizations Page

image

### Team Page

image

## References

- https://placeholders.dev/

[SMR-478]: https://sagebionetworks.jira.com/browse/SMR-478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/openchallenges/api-description/src/components/schemas/ImageQuery.yaml
```diff
@@ -1,11 +1,18 @@
...
 type: object
-description: An image query.
+description: >
+  An image query that identifies an image either by an object storage key
+  or by a direct URL. Exactly one of `objectKey` or `imageUrl` must be provided.
+additionalProperties: false
 properties:
   objectKey:
     $ref: ImageKey.yaml
+  imageUrl:
```
tschaffter, line 10: This property can be used if Thumbor is started with its HTTP image loader.

PR #3488, opened 2025-09-19
feat(explorers): update site version in model-ad footer, create shared service to manage data version and site version (MG-373)

hallieswan:
## Description

Currently, the Model-AD footer displays a static placeholder value (1.0.0) for the Site Version. We should update the app to dynamically display the Site Version based on the app version config value and the git tag, as in Agora. We should also create a shared service that can be used in both Agora and Model-AD to get and format Data Versions and Site Versions.

Note: `TAG_NAME` is already [set in modeladexplorer-infra](https://github.com/Sage-Bionetworks-IT/modeladexplorer-infra/blob/dev/app.py#L171), so we should see the updated Site Version in the deployed app when this feature is merged without any updates to the infra code.

## Related Issue

[MG-373](https://sagebionetworks.jira.com/browse/MG-373)

## Changelog

- Display dynamic site version in Model-AD footer
- Create shared explorers service to manage Data Version and Site Version fetching/formatting 
- Remove unused variables from Model-AD config

## Preview

Note: the tag is set to "edge" in the local environment. There isn't an "edge" tag in the sage-monorepo, so the app falls back to just displaying "edge" (the app version). 

dev (current) | pr (updated) 
:---: | :---: 
MG-373_modelad_dev | MG-373_modelad_pr

[MG-373]: https://sagebionetworks.jira.com/browse/MG-373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/model-ad/app/.env.example
```diff
@@ -1,7 +1,6 @@
 API_DOCS_URL="http://localhost:8000/api-docs"
-APP_VERSION="1.0.0-alpha"
+APP_VERSION="edge"
```
hallieswan, line 2: Align on default values used in Agora, which will be overridden by the deployment code

apps/model-ad/app/.env.example
```diff
@@ -1,7 +1,6 @@
...
 CSR_API_URL="http://localhost:8000/api/v1"
-DATA_UPDATED_ON="2023-09-26"
-ENVIRONMENT="production"
```
hallieswan, line 5: Remove unused variables

apps/model-ad/app/.env.example
```diff
@@ -1,7 +1,6 @@
...
 GOOGLE_TAG_MANAGER_ID=""
-SSR_API_URL="http://model-ad-api:3333/v1"
\ No newline at end of file
+SSR_API_URL="http://model-ad-api:3333/v1"
+TAG_NAME="edge"
```
hallieswan, line 6: Add `TAG_NAME`, which is used in the site version

beatrizsaldana (approved):
I am not too familiar with angular, so my approval might not be enough to merge to main, but this is a very clean and clear PR, thank you! I asked a question during my review process that doesn't requires any changes, so I am approving :)

libs/explorers/services/src/lib/version.service.ts
```diff
@@ -0,0 +1,103 @@
...
+        .getDataVersion()
+        .pipe(takeUntilDestroyed(this.destroyRef))
+        .subscribe({
+          next: (data) => {
+            const formattedVersion = this.formatDataVersion(data);
+            onSuccess(formattedVersion);
+          },
+          error: (error) => {
+            console.error('Error loading data version:', error);
+            if (onError) {
```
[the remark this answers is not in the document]
hallieswan: Great question! The [config](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/apps/agora/app/src/config/config.json) includes a value that we can fallback to for the site version (`appVersion`), but there isn't an equivalent value for the data version. If the call to fetch the data version fails, then we can only log an error and notify the caller.

apps/model-ad/app/src/config/config.json
```diff
@@ -1,9 +1,8 @@
 {
   "apiDocsUrl": "http://localhost:8000/api-docs",
-  "appVersion": "1.0.0-beta",
+  "appVersion": "edge",
```
tschaffter, line 3: "edge" is a publisheably tag (e.g. for images published on GHCR). An alternative tag we use for Docker images is "local", which represents whatever code is run locally (e.g. in a feature branch being developed). This convention has never been discussed before so your feedback is welcome!

apps/model-ad/app/src/config/config.json
```diff
@@ -1,9 +1,8 @@
...
   "csrApiUrl": "http://localhost:8000/api/v1",
-  "dataUpdatedOn": "yyyy-mm-dd",
-  "environment": "development",
   "googleTagManagerId": "",
-  "ssrApiUrl": "http://model-ad-api:3333/v1"
+  "ssrApiUrl": "http://model-ad-api:3333/v1",
+  "tagName": "edge"
```
tschaffter, line 7: What is the purpose of this property? What resource is being tagged? Consider changing the property name to be more descriptive.

libs/model-ad/testing/src/lib/mocks/config-mocks.ts
```diff
@@ -1,14 +1,13 @@
-import { AppConfig, Environment } from '@sagebionetworks/model-ad/config';
+import { AppConfig } from '@sagebionetworks/model-ad/config';
 
 export const configMock: AppConfig = {
   apiDocsUrl: 'http://localhost:8000/api-docs',
-  appVersion: '1.0.0-beta',
+  appVersion: 'edge',
```
tschaffter, line 5: See previous comment about edge/local.

libs/model-ad/testing/src/lib/mocks/config-mocks.ts
```diff
@@ -1,14 +1,13 @@
...
   csrApiUrl: 'http://localhost:4200/v1',
-  dataUpdatedOn: 'yyyy-mm-dd',
-  environment: Environment.Development,
   googleTagManagerId: '',
   ssrApiUrl: 'http://model-ad-api:3333/v1',
   isPlatformServer: false,
   privacyPolicyUrl: 'http://localhost:4200/privacy-policy',
   termsOfUseUrl: 'http://localhost:4200/terms-of-use',
+  tagName: 'edge',
```
tschaffter, line 12: See previous comment about edge/local.

tschaffter (changes requested):

apps/model-ad/app/src/config/config.json
hallieswan: We used "edge" to align with the value used for the [modeladexplorer-infra dev site](https://github.com/Sage-Bionetworks-IT/modeladexplorer-infra/blob/dev/app.py#L46), but I agree that "local" would be more appropriate here. I'll update in the next commit.

tschaffter (approved):

PR #3496, opened 2025-09-23
refactor(sage-monorepo): normalize Python uv project and package names, and update root `pyproject.toml`

tschaffter:
## Description

The first contribution of this PR is to normalize the name of Python projects within the `uv` workspace. While the "python" keyword is needed in Nx project names to differentiate a project from a counterpart developed for another language, projects in the uv workspace are understood as being all Python projects. Hence, the "-python" suffix is removed from the uv project names.

For example, the uv project name:

- "bixarena-api-client-python" becomes "bixarena-api-client"
- "openchallenges-api-client-python" becomes "openchallenges-api-client"

The same approach is applied to their package names, which typically match the name of their source folders.

> [!NOTE]
> An exception is Python projects that adopt the source folder name `src`. This folder name is fine for projects that are not meant to be imported in other projects. We may revisit this exception in the future.

For example:

- `libs/bixarena/api-client-python/bixarena_api_client_python` becomes `libs/bixarena/api-client-python/bixarena_api_client`

The immediate benefit of this approach is simpler import names:

For example:

```python
# Before:
# from bixarena_api_client_python import ApiClient, Configuration, LeaderboardApi
# from bixarena_api_client_python.exceptions import ApiException
from bixarena_api_client import ApiClient, Configuration, LeaderboardApi
from bixarena_api_client.exceptions import ApiException
```

Finally, this PR revisits the content of the root `pyproject.toml` and clarify when a project should be referenced in three different sections. For example, only one project currently needs to be listed as a root dependency (`openchallenges-client`) because its CLI should be executable from the root with `uv run openchallenges ...`.

One benefit of this approach is that the command `workspace-install` will complete faster, whether run inside the local dev environment or in CI/CD workflows.

## Changelog

- Normalize all the Python project names.
- Remove content from the root `pyproject.toml` that is not needed and add documentation as comment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,13 +80,13 @@ description = ""
 readme = "README.md"
 package-mode = false
 
-  [[project.authors]]
-  name = "Thomas Schaffter"
-  email = "thomas.schaffter@gmail.com"
+[[project.authors]]
+name = "Thomas Schaffter"
+email = "thomas.schaffter@gmail.com"
 
-  [[project.authors]]
-  name = "Rong Chai"
-  email = "rchai@sagebionetworks.org"
+[[project.authors]]
+name = "Rong Chai"
+email = "rongrong.chai1@gmail.com"
 
 [dependency-groups]
 dev = [


rrchai (approved):
LGTM

PR #3504, opened 2025-09-26
refactor: remove "-angular" suffix from TypeScript aliases

tschaffter:
## Description

Remove "-angular" suffix from TypeScript aliases.

## Changelog

- Remove "-angular" suffix from TypeScript aliases.
- Remove the project `sandbox-angular-lib`.

### Before

```typescript
import { Challenge } from '@sagebionetworks/openchallenges/api-client-angular';
```

### After

```typescript
import { Challenge } from '@sagebionetworks/openchallenges/api-client';
```


hallieswan:
> @hallieswan How should we apply this convention to these two libraries? Removing the suffix from the second one would conflict with the name of the first one.
> 
> ```json
>       "@sagebionetworks/explorers/charts": ["libs/explorers/charts/src/index.ts"],
>       "@sagebionetworks/explorers/charts-angular": ["libs/explorers/charts-angular/src/index.ts"],
> ```

@tschaffter That's a good question! `charts` currently contains charts built purely in TS and `charts-angular` wraps those charts as Angular directives. We originally added the `-angular` suffix with the idea that we could create a similar wrapper library for React in the future, e.g. `charts-react`. Do you have any ideas for names that would capture that context and follow this convention?

PR #3506, opened 2025-09-26
fix(explorers): handle paginated response from GitHub tags service (MG-418)

hallieswan:
## Description

The [explorers github tags service](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/explorers/services/src/lib/github.service.ts) currently returns the first page of tags. However, the tag of interest may not be on the first page. We should iterate through each page until the tag of interest is found or the last page is reached.

## Related Issue

[MG-418](https://sagebionetworks.jira.com/browse/MG-418)

## Changelog

- Handles paginated response from GitHub tags service

## Preview

Set `APP_VERSION="4.1.0-rc2"` and `TAG_NAME="agora/v4.1.0-rc2"` in `apps/agora/app/.env`
Build and run the app: `agora-build-images && agora-docker-start`

current (prod) | updated (this PR) 
:---: | :---: 
MG-418_siteVersion_prod | MG-418_siteVersion_PR

[MG-418]: https://sagebionetworks.jira.com/browse/MG-418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/agora/gene-comparison-tool/jest.config.ts
+++ b/libs/agora/gene-comparison-tool/jest.config.ts
@@ -15,7 +15,9 @@ export default {
     ],
   },
   testEnvironment: 'jest-fixed-jsdom',
-  transformIgnorePatterns: ['node_modules/(?!(.pnpm/.*/node_modules/)?(.*\\.mjs$|until-async))'],
+  transformIgnorePatterns: [
+    'node_modules/(?!(.pnpm/.*/node_modules/)?(.*\\.mjs$|until-async|@octokit/.*|universal-user-agent|before-after-hook))',
+  ],
   moduleNameMapper: {
     d3: '<rootDir>/../../../node_modules/d3/dist/d3.min.js',
     '^d3-(.*)$': '<rootDir>/../../../node_modules/d3-$1/dist/d3-$1.min.js',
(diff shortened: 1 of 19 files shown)


libs/agora/gene-comparison-tool/jest.config.ts
```diff
@@ -15,7 +15,9 @@ export default {
     ],
   },
   testEnvironment: 'jest-fixed-jsdom',
-  transformIgnorePatterns: ['node_modules/(?!(.pnpm/.*/node_modules/)?(.*\\.mjs$|until-async))'],
+  transformIgnorePatterns: [
+    'node_modules/(?!(.pnpm/.*/node_modules/)?(.*\\.mjs$|until-async|@octokit/.*|universal-user-agent|before-after-hook))',
+  ],
```
hallieswan, line 20: This library imports `explorers-services`, which imports `@octokit/rest` (which uses ES modules), so we need to transform ES modules in tests.

tschaffter (approved):
Following our discussion on Slack, it's OK to merge the current PR to fix the limitation of the current implementation (potentially unable to find the proper SHA if not on the first page of tags). However, fetching GitHub tags should be removed whenever we have the chance and instead rely on passing the SHA as part of the app config (see my previous comment).

hallieswan:
> fetching GitHub tags should be removed whenever we have the chance and instead rely on passing the SHA as part of the app config

@tschaffter Agreed! I opened [AG-1882](https://sagebionetworks.jira.com/browse/AG-1882) to track the work.

PR #3523, opened 2025-09-29
fix(synapse): use alternative method to run the dev container to update Synapse API (SMR-435)

tschaffter:
## Description

This PR fix the broken workflow that updates the Synapse API description and its generated artifacts.

As part of the solution, this PR explored another way to run tasks and GitHub Actions inside the dev container provided by the monorepo.

The benefit of this approach is it's "simplicity", compared to when using the dev container CLI, its faster dev container setup runtime, and the ability to using GitHub actions that depends on tools installed inside the dev container (here `peter-evans/create-pull-request@v7` that creates commits and push to a remote, triggering git hooks that executes tools provided by the dev container).

## Related issues

- [SMR-435](https://sagebionetworks.jira.com/browse/SMR-435)

### Dev container setup runtime

The new approach takes about 3 minutes to setup the dev container and run `workspace-install`.

image

This is potentially faster than when starting it using the dev container CLI, though this evaluation is not robust. Below is the setup run time of our CI workflow run when merging to `main` and executed 12 hours ago.

image

[SMR-435]: https://sagebionetworks.jira.com/browse/SMR-435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/update-synapse-api.yml
+++ b/.github/workflows/update-synapse-api.yml
@@ -3,7 +3,7 @@ name: Update Synapse API description and generated artifacts
 on:
   push:
     branches:
-      - synapse/update-synapse-api-test # trigger workflow when changes are pushed to this maintenance branch
+      - synapse/update-synapse-api-workflow-test # trigger workflow when changes are pushed to this maintenance branch
   schedule:
     # Run daily at 9 AM UTC (4 AM EST, 1 AM PST)
     - cron: '0 9 * * *'
--- a/.github/workflows/update-synapse-api.yml
+++ b/.github/workflows/update-synapse-api.yml
@@ -18,6 +18,9 @@ concurrency:
   cancel-in-progress: false
 
 env:
+  # NOTE: The following secrets must be defined in the repository or organization settings.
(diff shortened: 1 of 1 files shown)


.github/workflows/update-synapse-api.yml
```diff
@@ -1,6 +1,9 @@
 name: Update Synapse API description and generated artifacts
 
 on:
+  push:
+    branches:
+      - synapse/update-synapse-api-workflow-dev # trigger workflow when changes are pushed to this maintenance branch
```
hallieswan, line 6: Does this branch need to be protected with [a ruleset](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets#about-rulesets-and-protected-branches)?

hallieswan (approved):
Looks good to me!

.github/workflows/update-synapse-api.yml
tschaffter: Good catch! I will create one.

PR #3540, opened 2025-10-01
feat(agora): add api description and route for enhanced Agora search (AG-1876)

hallieswan:
## Description

We would like to enhance the Agora search to improve results ordering and return a response aligned with the new shared search input, which offers features that are not available in the current Agora search input (e.g. search query highlighting in search results, keyboard selection of search results).

## Related Issue

- [AG-1876](https://sagebionetworks.jira.com/browse/AG-1876)
- [AG-443](https://sagebionetworks.jira.com/browse/AG-443)

## Changelog

- Adds enhanced Agora search API description and route, which improves search result ordering, returns a tailored response object, and adds ability to search entire alias (rather than only by prefixes)
- Moves `escapeRegexChars` from `model-ad-api` to `shared-typescript-util` so the function can be used by both Model-AD and Agora
- Adds index for `alias` to `geneinfo` collection to improve search performance. Note: the same change will need to be made in [`agora-data-manager`](https://github.com/Sage-Bionetworks/agora-data-manager/blob/develop/create-indexes.js#L8-L15) so the index is created in the deployed collection.

## Preview

#### Exact Alias Match - 'pl'
 
- exact alias match ("PL"): ordered by hgnc_symbol (CSH1 before CSH2)
- partial hgnc_symbol match: ordered by match position, then hgnc_symbol

AG-1876_ordering

#### Partial Matches - '-tp'

- partial hgnc_symbol match: ordered by match position
- partial alias match: ordered by match position (PI-TP-beta (pos3) before NSL-TP (pos4)), then hgnc_symbol (PC-TP and PI-TPalpha have same match position, but PCTP is before PITPNA)

AG-1876_specialChars_ordering

#### Break ties by ensembl_gene_id 

search for 'gabarapl3'
- hgnc_symbol ties broken by ensembl_gene_id sort

AG-1876_gabarapl3

search for 'dpde1'
- alias ties broken by ensembl_gene_id sort

AG-1876_dpde1

#### Search with full ID

Result is returned: 

AG-1876_fullENSG

#### Search with partial ID 

Error is returned: 

AG-1876_error_ensg

### Invalid Search 

AG-1876_error_length

[AG-1876]: https://sagebionetworks.jira.com/browse/AG-1876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AG-443]: https://sagebionetworks.jira.com/browse/AG-443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/agora/api/src/api.ts
```diff
@@ -61,6 +62,7 @@ mongoose.connection.once('open', async () => {
   router.get('/data-version', dataVersionRoute);
   router.get('/distribution', distributionRoute);
   router.get('/genes/search', searchGeneRoute);
+  router.get('/genes/search/enhanced', searchGeneEnhancedRoute);
```
hallieswan, line 65: Maintain current search route until new route is fully integrated, then consider removing the original route.

apps/agora/data/src/data/collections-indexes.json
```diff
@@ -4,6 +4,7 @@
     "indexes": [
       { "ensembl_gene_id": 1, "hgnc_symbol": 1 },
       { "hgnc_symbol": 1 },
+      { "alias": 1 },
```
hallieswan, line 7: Add index for searched field to improve performance -- the same change will need to be made in agora-data-manager

hallieswan: I considered adding a text search index, but since the text search index expects whole word tokens, it's ineffective for our use case (substring search in single-word values).

hallieswan:
The enhanced search route also decreases response size and time:

query | current search | enhanced search
:---: | :---: | :---: 
`00` | image | image
`yes` | image | image
`ra` | image | image

PR #3555, opened 2025-10-04
fix: create Python-specific Nx task inputs to better identify affected Python projects (SMR-522)

tschaffter:
## Related Issue

- [SMR-522](https://sagebionetworks.jira.com/browse/SMR-522)

## Changelog

- Add the named input `defaultPython` modeled upon `default`.
- Add the named input `productionPython` modeled upon `production`.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -212,7 +212,13 @@
       "!{projectRoot}/.storybook/**/*",
       "!{projectRoot}/tsconfig.storybook.json",
       "!{projectRoot}/src/test/**/*"
-    ]
+    ],
+    "defaultPython": [
+      "{projectRoot}/**/*",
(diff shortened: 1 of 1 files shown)


nx.json
```diff
@@ -212,7 +225,13 @@
...
       "!{projectRoot}/tsconfig.storybook.json",
       "!{projectRoot}/src/test/**/*"
-    ]
+    ],
+    "defaultPython": [
+      "{projectRoot}/**/*",
+      "{workspaceRoot}/pyproject.toml",
+      "{workspaceRoot}/uv.lock"
+    ],
+    "productionPython": ["defaultPython"]
```
tschaffter, line 234: We can refine this value in the future to exclude test files and others using the "production" value designed for TypeScript projects.

PR #3529, opened 2025-09-30
feat(bixarena): enable to fetch real model data from database (SMR-475)

rrchai:
## Related Issue

[SMR-475](https://sagebionetworks.jira.com/browse/SMR-475)

PR #3556, opened 2025-10-06
feat(bixarena): fetch active example prompts from database to battle page (SMR-523)

rrchai:
## Description

Add dynamic fetching of active example prompts from database via BixArena API on the battle page.

## Related Issue

[SMR-523](https://sagebionetworks.jira.com/browse/SMR-523)

- depends on the new openAPI generator template from #3529 

## Changelog

- Integrate the BixArena API to dynamically load and display active example prompts on the battle page
- Reduce codes by replacing manual prompt text truncation with CSS 3-line clamp 

[SMR-523]: https://sagebionetworks.jira.com/browse/SMR-523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/bixarena/app/bixarena_app/page/example_prompt_ui.py
+++ b/apps/bixarena/app/bixarena_app/page/example_prompt_ui.py
@@ -14,6 +14,7 @@ from bixarena_api_client import ApiClient, Configuration, ExamplePromptApi
 from bixarena_api_client.models.example_prompt_search_query import (
     ExamplePromptSearchQuery,
 )
+from bixarena_api_client.models.example_prompt_sort import ExamplePromptSort
 
 logger = logging.getLogger(__name__)
 
@@ -39,7 +40,7 @@ class ExamplePromptUI:
                 api_instance = ExamplePromptApi(api_client)
                 search_query = ExamplePromptSearchQuery(
                     page_size=num_prompts,  # use alias defined by generator
-                    sort="random",
+                    sort=ExamplePromptSort.RANDOM,
                 )
                 resp = api_instance.list_example_prompts(
                     example_prompt_search_query=search_query


apps/bixarena/app/bixarena_app/page/bixarena_battle.py
```diff
@@ -39,22 +42,48 @@
...
+            # Fetch example prompts
+            response = api_instance.list_example_prompts(
+                example_prompt_search_query=search_query
+            )
+
+            # Extract questions and sample randomly
+            display_prompts = random.sample(
+                [prompt.question for prompt in response.example_prompts],
+                min(num_prompts, response.total_elements),
+            )
```
tschaffter, line 68: The behavior of this feature is not fully define here because the page size is not specified. Here, the behavior of the feature would change if the default page size changes in the future. The feature would even break if this number become smaller than `num_prompts`.

Always try to identify all the variables that may affect the behavior of a feature, assume that most of them may change in the future, and add guardrails to ensure that these changes don't go unnoticed.

For example, if the default page size was reduced to `num_prompts`, the app wouldn't not beak but the feature would as the same three prompts would always be returned (deterministic prompts).

The more variables you can control and fix, the more robust the feature will be.

tschaffter, line 68: Instead of fetch more example prompts than needed, which take more time to travel, we could request the exact number of prompts needed and request them to be randomly selected by the backend. We can use the `sort` property of the page to request the results to be sorted randomly, which the backend should use to fetch items randomly, and not only sort randomly the first page of results.

This is the approach I took for OC to display the three random challenges of the day on the home page. The OC implementation relies on OpenSearch to do the random selection, which is not available as part of BixArena. Instead, we could achieve this by defining an SQL query in the example prompt JPA repository in the Java backend. Since Spring Data JPA likely don't provide a "magic" syntax to select items randomly, we could add to the repository a method that is define by an SQL statement like done here:

https://github.com/Sage-Bionetworks/sage-monorepo/blob/51ea18038177e9e6cdd7aff12930173b8c6d60e9/apps/openchallenges/challenge-service/src/main/java/org/sagebionetworks/openchallenges/challenge/service/model/repository/ChallengeInputDataTypeRepository.java#L11-L13

tschaffter (changes requested):

apps/bixarena/api/src/main/java/org/sagebionetworks/bixarena/api/model/dto/ExamplePromptSortDto.java
```diff
@@ -28,7 +28,9 @@ public enum ExamplePromptSortDto {
   
   SOURCE("source"),
   
-  ACTIVE("active");
+  ACTIVE("active"),
```
tschaffter, line 31: Remove

apps/bixarena/api/src/main/java/org/sagebionetworks/bixarena/api/model/repository/ExamplePromptRepository.java
```diff
@@ -1,11 +1,25 @@
...
+import org.springframework.data.repository.query.Param;
 import org.springframework.stereotype.Repository;
 
 @Repository
 public interface ExamplePromptRepository
-  extends JpaRepository<ExamplePromptEntity, UUID>, JpaSpecificationExecutor<ExamplePromptEntity> {}
+  extends JpaRepository<ExamplePromptEntity, UUID>, JpaSpecificationExecutor<ExamplePromptEntity> {
+  @Query(
+    value = "SELECT * FROM example_prompt ep " +
+    "WHERE (:active IS NULL OR ep.active = :active) " +
```
tschaffter, line 17: Only fetch random results among the active ones.

PR #3572, opened 2025-10-14
feat(bixarena): authenticate with Synapse using BixArena backend (SMR-539)

tschaffter:
## Description

Currently, a Synapse OAuth client is integrated into the BixArena web app (Gradio) to enable users to login with their Synapse account.

This PR move the integration of the Synapse OAuth client to the backend to enable us to capture information about the users and, and soon authorize the web app to access protected endpoints on behalf of the user. This PR focus on the implementation of the OIDC authentication flow with Synapse. This flow is demonstrated by the Login and Logout buttons displayed in the web app (see Preview section).

## Related Issue

- [SMR-539](https://sagebionetworks.jira.com/browse/SMR-539)

## Changelog

- Add endpoint to `bixarena-api` to handle the Synapse OIDC login flow.
- Implement persistent login state that survives page reloads.
- Add logout button.
- Remove OAuth client and related configuration from bixarena-app.
- Simplify authentication module in bixarena-app.

### Build and start the stack

> [!IMPORTANT]
> The config file (`.env`) of `bixarena-app` and `bixarena-api` have been updated. In particular, the Synapse OAuth client ID and secret are now specified in the `bixarena-api`'s `.env` file and can be removed from the `.env` file of `bixarena-app`.

```bash
bixarena-build-images
nx serve-detach bixarena-apex
```

### Demo

https://github.com/user-attachments/assets/23960c31-b1ff-4e68-aca6-4b92a4107473

[SMR-539]: https://sagebionetworks.jira.com/browse/SMR-539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/bixarena/api/src/test/java/org/sagebionetworks/bixarena/api/BixArenaApiApplicationTests.java
+++ b/apps/bixarena/api/src/test/java/org/sagebionetworks/bixarena/api/BixArenaApiApplicationTests.java
@@ -7,7 +7,19 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.context.ActiveProfiles;
 
-@SpringBootTest
+@SpringBootTest(properties = {
+  "app.welcome-message=Welcome to BixArena API (test)",
+  "app.ui-base-url=http://localhost:8100",
+  "app.auth.authorize-url=https://example.org/authorize",
+  "app.auth.token-url=https://example.org/token",
+  "app.auth.jwks-url=https://example.org/jwks",
+  "app.auth.client-id=test-client",
+  "app.auth.client-secret=test-secret",
+  "app.auth.redirect-uri=http://localhost:8112/v1/auth/oidc/callback",
+  "app.auth.internal-issuer=http://localhost",
+  "app.auth.audience=test-audience",
+  "app.auth.token-ttl-seconds=600"
+})
 @ActiveProfiles("test")
 class BixArenaApiApplicationTests {
 


apps/bixarena/api/src/main/java/org/sagebionetworks/bixarena/api/model/dto/Echo200ResponseDto.java
tschaffter: The name of this class looks weird because it's generated from an inline OpenAPI component schema. A more developer-friendly name will be specified in the future by moving this component schema to its own file. The same comments applies to classes generated in Java and Python.

apps/bixarena/api/src/main/java/org/sagebionetworks/bixarena/api/model/entity/ExternalAccountEntity.java
```diff
@@ -0,0 +1,73 @@
...
+    @UniqueConstraint(columnNames = { "provider", "external_id" }),
+  }
+)
+@Getter
+@Setter
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@ToString(exclude = { "user" })
+public class ExternalAccountEntity {
```
tschaffter: Adapted from OC, where external accounts can be Google or Synapse accounts. Currently, only Synapse account are supported as external account to login in BixArena.

apps/bixarena/api/src/main/java/org/sagebionetworks/bixarena/api/model/entity/UserEntity.java
```diff
@@ -0,0 +1,96 @@
...
+  @UpdateTimestamp
+  @Column(name = "updated_at", nullable = false)
+  private OffsetDateTime updatedAt;
+
+  public enum Role {
+    admin,
+    user,
+    readonly,
+    service,
+  }
```
tschaffter: The service role will be used when one service sends a request to another (e.g., API → AI service). A read-only role may not be necessary. Role definitions will be implemented in a follow-up PR.

libs/bixarena/api-client-python/bixarena_api_client/api/auth_api.py
```diff
@@ -0,0 +1,1215 @@
...
+    MintInternalToken200Response,
+)
+from bixarena_api_client.models.oidc_callback200_response import OidcCallback200Response
+
+from bixarena_api_client.api_client import ApiClient, RequestSerialized
+from bixarena_api_client.api_response import ApiResponse
+from bixarena_api_client.rest import RESTResponseType
+
+
+class AuthApi:
```
tschaffter: We could try to use the Auth API provided by this client in the future. For now, the web app is implementing the redirection to start the OIDC auth flow and store the session cookie.

libs/bixarena/api-client-python/bixarena_api_client/models/admin_stats200_response.py
tschaffter: Same comment as before about auto-generated schema names.

PR #3574, opened 2025-10-15
infra: create CDK library `platform-infra-cdk-common` to centralize reusable components

tschaffter:
## Description

Create the CDK library `platform-infra-cdk-common` to centralize reusable infrastructure components. For demonstration purposes, this library defines the constant `VALID_ENVIRONMENTS`, which is then imported by the `bixarena-infra-cdk` and `openchallenges-infra-cdk` applications.

```bash
nx graph
```

image

## Related Issue

- [IT-4658](https://sagebionetworks.jira.com/browse/IT-4658)

## Changelog

- Create CDK library `platform-infra-cdk-common` to centralize reusable components.
- Import the CDK library into the `bixarena-infra-cdk` and `openchallenges-infra-cdk` applications.

### Test the library

```bash
nx test platform-infra-cdk-common
```

### Synth the CDK apps

```bash
nx synth bixarena-infra-cdk
nx synth openchallenges-infra-cdk
```

[IT-4658]: https://sagebionetworks.jira.com/browse/IT-4658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/bixarena/infra/cdk/bixarena_infra_cdk/app.py
+++ b/apps/bixarena/infra/cdk/bixarena_infra_cdk/app.py
@@ -3,15 +3,13 @@
 from os import environ
 
 import aws_cdk as cdk
-from platform_infra_cdk_common.hello import hello
+from platform_infra_cdk_common.constants import VALID_ENVIRONMENTS
 
 from .ecs_stack import EcsStack
 from .load_balancer_stack import LoadBalancerStack
 from .network_stack import NetworkStack
 from .service_props import ServiceProps
 from .service_stack import LoadBalancedServiceStack
-
-VALID_ENVIRONMENTS = ["dev", "stage", "prod"]
 
 
 def build_app() -> cdk.App:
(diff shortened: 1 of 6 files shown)


apps/bixarena/infra/cdk/pyproject.toml
```diff
@@ -25,6 +25,7 @@ readme = 'README.md'
 dependencies = [
   "aws-cdk-lib>=2.219.0",
   "boto3>=1.40.51",
+  "platform-infra-cdk-common",
```
zaro0508: what is the workflow for this dependency?  will it get published somewhere (like on pypi) for this reference to work?  or does it only get referenced locally?

zaro0508 (changes requested):

apps/bixarena/infra/cdk/pyproject.toml
tschaffter: > what is the workflow for this dependency?

This dependency is defined at the end of the app's `pyproject.toml`, which specified that the library is part of the uv workspace defined in the root `pyproject.toml`.

> will it get published somewhere (like on pypi) for this reference to work? or does it only get referenced locally?

The library can be consumed directly in another library and apps as discussed this morning, without requiring to publish it. Regarding your question about referencing another version of a local library, that's not the recommended approached used in the monorepo. It may technically be possible, but this may imply that the project that needs another version of the lib may need to be excluded from the shared uv workspace to become an independent Python project that still lives in the monorepo but manages its own dependencies.

zaro0508 (approved):

PR #3596, opened 2025-10-23
feat(model-ad): add column visibility control (MG-427)

sagely1:
## Description
Add column visibility control.  Each CT (if dropdowns exist) should preserve preferences and navigating between CTs should preserve preferences.

## Related Issue

[MG-427](https://sagebionetworks.jira.com/browse/MG-427)

## Changelog

- Add column visibility control
- Add new icon that shows a badge when not all columns are displayed (shows all by default)
- Preserve state when navigating any CT and dropdown combination
- Update ComparisonToolService to use ComparisonToolColumn objects

## Preview
https://www.loom.com/share/8af814ed81ca4a858852f1fe4f77b941

[MG-427]: https://sagebionetworks.jira.com/browse/MG-427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/services/src/lib/comparison-tool.service.ts
+++ b/libs/explorers/services/src/lib/comparison-tool.service.ts
@@ -1,4 +1,4 @@
-import { computed, signal, Signal } from '@angular/core';
+import { computed, Injectable, signal, Signal } from '@angular/core';
 import {
   ComparisonToolConfig,
   ComparisonToolColumn,
@@ -15,6 +15,8 @@ import { isEqual } from 'lodash';
  * - Update user selections through {@link setDropdownSelection}; consumers can read the latest value via {@link dropdownSelection}.
  * - Manage result totals via {@link totalResultsCount} and {@link pinnedResultsCount}.
  */
+
+@Injectable()
 export class ComparisonToolService {
   private readonly configsSignal = signal<ComparisonToolConfig[]>([]);
   private readonly dropdownSelectionSignal = signal<string[]>([]);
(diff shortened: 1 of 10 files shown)


libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -80,11 +102,24 @@ export class ComparisonToolService {
...
+      page: config.page,
+      dropdowns: config.dropdowns,
+      columns: config.columns
+        .filter((column) => column.name !== undefined)
+        .map((column) => ({
+          name: column.name as string,
+          selected: true,
+        })),
+    }));
+    this.columnsForAllPagesSignal.set(columnsData);
```
hallieswan: What do you think about making `columnsForAllPages` a `computed` signal instead? Then it would be automatically updated whenever `configsSignal` or `dropdownSelectionSignal` were updated, so then we wouldn't have to manually set `columnsForAllPagesSignal` in multiple places in the code.

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -97,6 +132,38 @@ export class ComparisonToolService {
...
+            .filter((column) => column.name !== undefined)
+            .map((column) => ({
+              name: column.name as string,
+              selected: true,
+            })),
+        };
+
+        return [...columnsData, newPageData];
+      });
+    }
```
hallieswan: This logic could also be moved into the computed `columnsForAllPages`

hallieswan (approved):
Looks good!

PR #3601, opened 2025-10-24
feat(bixarena): configure the API gateway to forward requests to downstream services (SMR-553)

tschaffter:
## Description

Configure the BixArena API Gateway to forward requests to downstream services and make routing/security behavior driven by route configuration generated from OpenAPI specs. This centralizes route metadata, enables deterministic generation of `routes.yml`, and allows per-route anonymous/public endpoints to be declared and enforced by the gateway's security configuration.

## Architecture

bixarena-architecture

## Related Issue

Resolves [SMR-553](https://sagebionetworks.jira.com/browse/SMR-553)

## Changelog

- Add route configuration generation tooling and Gradle task to derive `routes.yml` from OpenAPI files.
- Introduce runtime route registry and configuration binding to drive gateway behavior.
- Make gateway security and route forwarding behavior depend on route metadata (audience, scopes, anonymous access).
- Update gateway configuration and OpenAPI annotations to support the above.

## Notes

> [!WARNING]
> This PR modified the `.env` file of `bixarena-app` to route requests to the API and auth services via the API gateway. ~For now, the app continues to route auth-related requests to the authorization server as rerouting the requests with a session cookie requires extra works.~ See `.env.example`.

> [!IMPORTANT]
> The traffic to the auth service is now redirected via the API gateway.

## Testing

Build the Docker images:

```bash
bixarena-build-images
```

Update the `.env` file of `bixarena-app`.

Start the stack with Docker Compose:

```bash
nx serve-detach bixarena-apex
```

Access the app on `http://127.0.0.1:8100` and check that the leaderboard page includes two models. You can also check the logs of the `bixarena-app` container to ensure that the leaderboard data and model data are properly loaded.

```bash
docker logs -f bixarena-app
```

### Features

The API Gateway now provides:
- A one-step route generation flow (Gradle task) to produce canonical `routes.yml` from OpenAPI files.
- A route registry loaded at startup that the gateway uses to determine which paths are public and how to forward requests.
- Security configuration that permits routes marked as anonymous while requiring authentication for other endpoints.
- Easier maintenance of route-level audience and scope data, derived from service OpenAPI definitions.

[SMR-553]: https://sagebionetworks.jira.com/browse/SMR-553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/bixarena/app/.env.example
+++ b/apps/bixarena/app/.env.example
@@ -1,6 +1,6 @@
-API_BASE_URL=http://bixarena-api:8113/api/v1
+API_BASE_URL=http://bixarena-api-gateway:8113/api/v1
 APP_PORT=8100
 ENVIRONMENT=development
-AUTH_BASE_URL_SSR=http://bixarena-auth-service:8115
+AUTH_BASE_URL_SSR=http://bixarena-api-gateway:8113
 AUTH_BASE_URL_CSR=http://127.0.0.1:8113
 OPENAI_API_KEY=changeme


apps/bixarena/api-gateway/src/main/java/org/sagebionetworks/bixarena/api/gateway/configuration/SecurityConfiguration.java
```diff
@@ -25,33 +20,29 @@
...
-      .build();
+      // .oauth2ResourceServer(ServerHttpSecurity.OAuth2ResourceServerSpec::jwt) // enable if you use JWT
+      .authorizeExchange(exchanges -> {
+        // Always allow actuator probes
+        exchanges.pathMatchers("/actuator/health", "/actuator/metrics").permitAll();
+
+        // Guard: only register matchers when there's at least one
+        if (anonymousPaths.length > 0) {
+          exchanges.pathMatchers(anonymousPaths).permitAll();
+        }
```
tschaffter: `pathMatchers` throws an error if it receives an empty array (i.e., if we have zero anonymously accessible paths). This code guards against defining `pathMatchers` if the input array is empty.

apps/bixarena/api-gateway/src/main/java/org/sagebionetworks/bixarena/api/gateway/util/OpenApiRouteConfigGenerator.java
tschaffter: This tool and the system to generate route config and load it in the app has been imported from the OC API gateway. In the future, this feature should be added to a shared Java library.

libs/bixarena/api-description/openapi/api-public.openapi.yaml
```diff
@@ -48,6 +48,7 @@ paths:
           $ref: '#/components/responses/Forbidden'
   /battles:
     get:
+      x-anonymous-access: true
```
tschaffter: This vendor extension has been applied to almost all the endpoints of the API service so that the API gateway will allow the rerouting to occur until we have finalized the integration of the auth service with the API gateway. Once done, these anonymous vendor extensions will be removed from the endpoints that require a JWT, that is, the endpoints that require the user to be authenticated.

PR #3613, opened 2025-10-28
feat(bixarena): display the actual public metrics on the home page of BixArena (SMR-560)

tschaffter:
## Description

Display real, computed public metrics on the BixArena home page so visitors see live platform activity instead of placeholder values. This replaces hard-coded/mock counts with database-backed values and wires those counts into the Gradio home page UI, improving transparency and developer confidence in public-facing metrics.

## Related Issue

Resolves [SMR-560](https://sagebionetworks.jira.com/browse/SMR-560)

## Changelog

- Query real statistics from the database (user count, total battles, distinct models evaluated)
- Add a read-only JPA entity and repository to support user count queries
- Return computed values from the public stats API endpoint instead of placeholders
- Wire the Gradio home page to fetch and render public stats asynchronously on page load
- Note: No DB migrations included; changes rely on existing `auth` and `api` schemas and read-only access

## Preview

Before logging in on a fresh deployment (0 users):

image

After completing the authentication flow and returning to the home page (1 users):

image

[SMR-560]: https://sagebionetworks.jira.com/browse/SMR-560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/bixarena/app/bixarena_app/main.py
+++ b/apps/bixarena/app/bixarena_app/main.py
@@ -13,6 +13,7 @@ from bixarena_app.page.bixarena_header import (
 )
 from bixarena_app.page.bixarena_home import (
     build_home_page,
+    load_public_stats_on_page_load,
     load_user_battles_on_page_load,
 )
 from bixarena_app.page.bixarena_leaderboard import build_leaderboard_page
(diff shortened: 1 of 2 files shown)


tschaffter:
> LGTM - the stats fetched as expected.

@rrchai What is the probability to run 4 battles with the same two models each time? ;-) Do you think the number of models evaluated is incorrect?

rrchai:
> @rrchai What is the probability to run 4 battles with the same two models each time? ;-) Do you think the number of models evaluated is incorrect?

The first 3 battles failed because of the issue in #3615, so I was not able to submit evaluation successfully but the last one. This turned out that 1 total battle -> 2 evaluated models, which is expected. Good observation!

PR #3616, opened 2025-10-29
chore(bixarena): normalize the height of the example prompts box and regenerate example prompts on page reloads (SMR-558)

tschaffter:
## Description

Normalize the example prompts UI and tighten the prompt input sizing to improve layout and reduce noisy warnings. This change also makes runtime logging level configurable via an environment variable and connects the example prompts component to the page lifecycle so prompts refresh when the battle page loads or when history is cleared. These updates improve user-facing layout and developer experience.

## Related Issue

Fixes [SMR-558](https://sagebionetworks.jira.com/browse/SMR-558)

## Changelog

- Make logging configurable via `LOG_LEVEL` and update logging initialization
- Integrate `ExamplePromptUI` into the battle page and add prompt refresh lifecycle (on load, navigation, and clear)
- Constrain and center the prompt input box and normalize example prompt card height with CSS
- Suppress noisy deprecation warnings and adjust Gradio Chatbot usage to preserve behavior
- Update UI builder return signatures and navigation wiring to support prompt refresh

### Current design

image

### Proposed design

Adjusted the example prompt boxes to have a uniform height and reduced the length of the prompt input box.

image

### Features

- Example prompt cards render at a consistent, vertically centered height
- Prompts refresh automatically when the battle page loads or when history is cleared
- Prompt input is narrower and centered to improve readability
- Runtime log level can be configured using the `LOG_LEVEL` environment variable (defaults to INFO)
- Noisy deprecation warnings are suppressed during Chatbot creation

Completion notes: all changes are limited to the BixArena application (`apps/bixarena/app/...`) and affect the Python/Gradio UI and runtime configuration only. There are no database schema or API-breaking changes outside the app; internal callers were updated to accommodate the new return signatures where required.

[SMR-558]: https://sagebionetworks.jira.com/browse/SMR-558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/bixarena/app/bixarena_app/main.py
+++ b/apps/bixarena/app/bixarena_app/main.py
@@ -246,10 +246,8 @@ def build_app(moderate=False):
                 user_battles_box,
             ) = build_home_page()
 
-        with gr.Column(visible=False) as battle_page_col:
-            battle_page_blocks, example_prompt_ui, prompt_outputs = build_battle_page(
-                moderate
-            )
+        with gr.Column(visible=False) as battle_page:
+            _, example_prompt_ui, prompt_outputs = build_battle_page(moderate)
 
         with gr.Column(visible=False) as leaderboard_page:
             build_leaderboard_page()
@@ -277,7 +275,7 @@ def build_app(moderate=False):
             + "</span>"
         )
 
-        pages = [home_page, battle_page_col, leaderboard_page, user_page]
+        pages = [home_page, battle_page, leaderboard_page, user_page]
         navigator = PageNavigator(pages)
 
         # Navigation - also refresh example prompts when navigating to battle page


apps/bixarena/app/bixarena_app/main.py
```diff
@@ -275,13 +277,19 @@ def build_app(moderate=False):
             + "</span>"
         )
 
-        pages = [home_page, battle_page, leaderboard_page, user_page]
+        pages = [home_page, battle_page_col, leaderboard_page, user_page]
```
rrchai: is this renaming necessary?

PR #3623, opened 2025-10-29
chore(bixarena): show the battle button in the header only when the user is logged in

tschaffter:
## Description

Show the battle button in the header only when the user is logged in.


rrchai (approved):
LGTM

PR #3630, opened 2025-10-30
feat(model-ad): adding a loading message to the loading animation (MG-432)

sagely1:
## Description
Update the loading animation to show a loading message.  Previously it was a count.

## Related Issue
[MG-432](https://sagebionetworks.jira.com/browse/MG-432)

Fixes #(issue)

## Changelog
- Update loading container to handle a loading message (String instead of a Number)

## Preview
image

Note: The loading message is initially "Loading results for you".  When the api call finishes and retrieves the configs (the row_counts are included in this call), then the message will update to (for example): "Loading over 2000 results for you".  Model overview is the only CT with data and it currently only has 16 rows so the message shown in the screenshot is only displayed momentarily.

[MG-432]: https://sagebionetworks.jira.com/browse/MG-432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/comparison-tools/src/lib/base-comparison-tool/base-comparison-tool.component.spec.ts
+++ b/libs/explorers/comparison-tools/src/lib/base-comparison-tool/base-comparison-tool.component.spec.ts
@@ -35,9 +35,9 @@ describe('Base Comparison Tool Component', () => {
     expect(component).toBeTruthy();
   });
 
-  it('should have default values', async () => {
+  it('should have loading results count while loading', async () => {
     const { component } = await setup();
     expect(component.isLoading()).toBe(true);
-    expect(component.resultsCount()).toBe(0);
+    expect(component.loadingResultsCount()).toBe(mockComparisonToolConfigs[0].row_count);
   });
 });


libs/explorers/comparison-tools/src/lib/base-comparison-tool/base-comparison-tool.component.ts
```diff
@@ -26,7 +26,7 @@ export class BaseComparisonToolComponent {
   isLoading = input(true);
 
   currentConfig = this.comparisonToolService.currentConfig;
-  resultsCount = this.comparisonToolService.totalResultsCount;
+  resultsCount = this.comparisonToolService.loadingResultsCount;
```
hallieswan: What do you think about changing the variable name in the component to match the service for clarity? 

```suggestion
  loadingResultsCount = this.comparisonToolService.loadingResultsCount;
```

libs/explorers/util/src/lib/loading-container/loading-container.component.html
```diff
@@ -1,4 +1,4 @@
 <div class="loading-container">
   <explorers-loading-icon />
-  <p>Loading {{ count() | number }} results for you</p>
+  <p>Loading {{ count() }} results for you</p>
```
hallieswan: What do you think about hiding this message if `count()` is an empty string or undefined?

libs/explorers/util/src/lib/loading-container/loading-container.component.ts
```diff
@@ -10,5 +10,5 @@ import { LoadingIconComponent } from '../loading-icon/loading-icon.component';
   encapsulation: ViewEncapsulation.None,
 })
 export class LoadingContainerComponent {
-  count = input(0);
+  count = input<string>();
```
hallieswan: An input [without a default value](https://angular.dev/guide/components/inputs) will be set to undefined, so `count`'s type is `string | undefined`. Should we handle the case where `count` is undefined in the template? Alternately, should we initialize this `input` with an empty string, so the type is `string`?

hallieswan (changes requested):

libs/explorers/comparison-tools/src/lib/base-comparison-tool/base-comparison-tool.component.ts
sagely1: Sure.  In the future, it might make sense to consider it renaming it to loadingMessage.

PR #3634, opened 2025-10-31
feat(bixarena): compute user rank based on number of completed battles (SMR-581)

tschaffter:
## Description

Add user ranking to BixArena based on the number of completed battles and expose it in API and UI. The backend computes ranks with a database query (standard competition ranking), caches per-user rank results in Valkey for short-term performance, and invalidates the cache when a battle is completed so ranks remain accurate for clients.

## Related Issue

Resolves [SMR-581](https://sagebionetworks.jira.com/browse/SMR-581)

## Changelog

- Add user ranking based on completed battles and return it in the user stats API
- Compute ranks with a database query and reduce repeated work via a short-lived Redis cache
- Invalidate cached ranks when a user completes a battle to keep results fresh
- Update OpenAPI schemas and generated client code to expose the new `rank` field
- Show the user's rank in the BixArena frontend UI
- Show the number of battle completed on the home page instead of the number of battles created.

## Preview

image

### Features

- API responses include `rank` in `UserStats` for authenticated users
- Backend computes rank using a DB query and serves cached results for repeated reads
- Redis-based per-user rank cache reduces DB load and uses a 5-minute TTL
- Cache entries for a user are evicted when that user completes a battle
- Frontend shows a new "Your Rank" box on the BixArena home page for logged-in users
- Python API client and generated docs now surface the `rank` field

[SMR-581]: https://sagebionetworks.jira.com/browse/SMR-581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/bixarena/app/bixarena_app/page/bixarena_home.py
+++ b/apps/bixarena/app/bixarena_app/page/bixarena_home.py
@@ -73,7 +73,7 @@ def fetch_user_stats(request: gr.Request):
             api = UserApi(client)
             user_stats = api.get_user_stats()
             logger.info(
-                f"Fetched user stats: {user_stats.total_battles} battles, rank #{user_stats.rank}"
+                f"Fetched user stats: {user_stats.completed_battles} completed battles, rank #{user_stats.rank}"
             )
             return user_stats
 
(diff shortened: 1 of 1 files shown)


rrchai (approved):
LGTM

PR #3642, opened 2025-10-31
feat(bixarena): add a set of example prompts for release (SMR-578, SMR-579)

rrchai:
## Description

Add a set of example prompts for release and make the entire surface of the example prompts boxes clickable

Used custom HTML buttons instead of Gradio buttons because Gradio Button doesn't have child div which makes it impossible to have both full surface clickable container and CSS line-clamp in the same time.

## Related Issue

[SMR-578](https://sagebionetworks.jira.com/browse/SMR-578)
[SMR-579](https://sagebionetworks.jira.com/browse/SMR-579)

## Changelog

- Add a set of example prompts for release (31)
- Add clickable example prompt cards with horizontal navigation (left/right arrows)
- Remove the `refresh_prompts()` calls from the navigation button handlers 

## Preview

Screen Shot 2025-11-01 at 12 55 06 AM

[SMR-578]: https://sagebionetworks.jira.com/browse/SMR-578?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SMR-579]: https://sagebionetworks.jira.com/browse/SMR-579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/bixarena/app/bixarena_app/page/battle_page_css.py
+++ b/apps/bixarena/app/bixarena_app/page/battle_page_css.py
@@ -91,8 +91,8 @@ EXAMPLE_PROMPTS_CSS = """
     pointer-events: none;
 }
 
-/* Mobile responsive layout */
-@media (max-width: 640px) {
+/* Responsive layout */
+@media (max-width: 1024px) {
     /* Stack outer container vertically and hide arrows */
     #prompt-card-section > .row {
         flex-direction: column;
@@ -139,7 +139,7 @@ INPUT_PROMPT_CSS = """
 /* Container for the input textbox - limit width and center */
 #input_box.prompt_input {
     background: var(--background-fill-primary);
-    max-width: 700px;
+    max-width: 900px;
     margin: 0 auto;
     width: 100%;
 }
@@ -156,7 +156,7 @@ INPUT_PROMPT_CSS = """
 
 /* Also limit the parent row container */
 .row:has(#input_box.prompt_input) {
-    max-width: 700px;
+    max-width: 900px;
     margin: 0 auto;
 }
 """


rrchai:
It looks like the upstream introduces an unexpected behavior that clicking "Battle" button will create the example prompt cards during a battle:

chrome-capture-2025-10-31

I will see if I can fix it in this PR.

apps/bixarena/api/src/main/resources/db/migration/V1.0.2__insert_example_prompts.sql
tschaffter: Some prompts are so long that they don't fit into their box. The example prompt should be entirely visible in the box and preferably of reasonable length so that the user can read them quickly.

Suggestion:
1. Identify the max length of the text that could fit in the box + margin of error because of how the text is wrapped at the word level.
2. Identify PubMedQA prompts that meet this max length requirements and pic 20-30 of them.
3. Set max length constraints to the OpenAPI spec and SQL table definition.
4. (Optional) The boxes could be made a tiny bit wider if deemed necessary. The text should ideally not spread over more than 3-4 lines.

tschaffter (changes requested):
Clicking anywhere on the boxes to start a battle works nicely! I have one request regarding the length of the example prompts.

rrchai:
> 1. Identify the max length of the text that could fit in the box + margin of error because of how the text is wrapped at the word level.
> 2. Identify PubMedQA prompts that meet this max length requirements and pic 20-30 of them.
> 3. Set max length constraints to the OpenAPI spec and SQL table definition.
> 4. (Optional) The boxes could be made a tiny bit wider if deemed necessary. The text should ideally not spread over more than 3-4 lines.

I’ve widened the boxes and kept the text clamped at three lines. From testing, each box fits around 170 -180 characters. Looking at the distribution, most of examples we have stay under 160 characters. I tried to increase the line clamp to 4 lines to cover 200 characters, but that would create noticeable empty space for shorter one-line examples. Then I take a safer ~10% error margin, so it rounds up to 160 characters as the maximum length.

prompt_length_distribution

I’ve updated the SQL table to include only the records that meet the defined maximum length constraint. I also added additional examples from PubMedQA, resulting in a total of 20 examples: 16 from PubMedQA and 4 from Bixarena (custom examples contributed by the team).

tschaffter (approved):
The example prompt cards are now fully clickable, and I’ve generated a wide range of random example prompts — all of which fit nicely within the card layout. Thanks!

PR #3644, opened 2025-10-31
fix(model-ad): fix for CT columns not showing, CT columns width resizing bug (MG-470, MG-488)

sagely1:
## Description
This fixes CT columns not showing in the UI due to a but in the ComparisonToolService.
This fixes a CT column resizing bug due to column sizing happening only on initial load.

Fixes #(issue)
[MG-470](https://sagebionetworks.jira.com/browse/MG-470)
[MG-488](https://sagebionetworks.jira.com/browse/MG-488)

## Changelog
- Fixes logic to show user preferences for column visibility.  Note that column visibility preferences apply to all combinations of dropdowns as each CT now has its own ComparisonToolService.  [MG-468](https://sagebionetworks.jira.com/browse/MG-468) will need to be readdressed in the future to allow a "column visibility dropdown depth setting" as there are two CTs with a depth of 2 and we need a mechanism to apply different behavior to both.
- FIxes a recently discovered bug with CT column resizing as the latest code from main only calculates the width of columns on page load and not when toggling columns.  Also, navigating to another page and back can result in columns being invisible due to this error if the count of columns is not the same as the initial count.

## Preview

[MG-468]: https://sagebionetworks.jira.com/browse/MG-468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MG-470]: https://sagebionetworks.jira.com/browse/MG-470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MG-488]: https://sagebionetworks.jira.com/browse/MG-488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/services/src/lib/comparison-tool.service.ts
+++ b/libs/explorers/services/src/lib/comparison-tool.service.ts
@@ -3,6 +3,7 @@ import {
   ComparisonToolColumn,
   ComparisonToolColumns,
   ComparisonToolConfig,
+  ComparisonToolConfigColumn,
   ComparisonToolViewConfig,
 } from '@sagebionetworks/explorers/models';
 import { isEqual } from 'lodash';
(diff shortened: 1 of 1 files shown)


hallieswan (approved):
Looks good!

PR #3656, opened 2025-11-05
fix(bixarena): move the model_error table after battle and round tables (SMR-592)

rrchai:
## Description

Reordered table creation in the SQL migration file to ensure `model_error` table is created after the tables it references (`battle` and `battle_round`).

## Related Issue

```console
Caused by: org.postgresql.util.PSQLException: ERROR: relation "api.battle" does not exist
```

## Changelog

- Reorder table creation in migration to ensure `model_error` table is created after `battle` and `round`


tschaffter:
@rrchai Does this PR addresses https://sagebionetworks.jira.com/browse/SMR-592?

Update: It looks like this fixes SMR-592

tschaffter (approved):

PR #3672, opened 2025-11-10
fix(bixarena): improve example prompt card responsive design to show all texts (SMR-601)

rrchai:
## Description

The issue: The default Gradio container has a `max-width: 1280px` which prevented the `page-content` from reaching its previously set `max-width: 1400px`.

Improves the responsive design of example prompt cards on the battle page. 

## Related Issue

[SMR-601](https://sagebionetworks.jira.com/browse/SMR-601)

## Changelog

- Override Gradio's default 1280px container max-width to 1600px to allow page-content to reach intended 1400px width
- Loosen the line limit of line clamp for a smaller screen
- Hide example prompts section entirely on mobile (< 1024px) for cleaner UX
- Fix an issue where the top border of example prompt card was cropped on hover
- Rename CSS `footer-no-padding` to `html-no-padding`

## Preview

From 1800px wide screen to a mobile screen <1024px:
chrome-capture-2025-11-10

- ≥ 1280px: allow max 4 lines, consistent card heights, flexible equal widths
- 1024px - 1280px: dynamical height, consistent card heights, flexible equal widths
- < 1024px:  dynamical height,, stack over each other

[SMR-601]: https://sagebionetworks.jira.com/browse/SMR-601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/bixarena/app/bixarena_app/page/battle_page_css.py
+++ b/apps/bixarena/app/bixarena_app/page/battle_page_css.py
@@ -40,7 +40,7 @@ EXAMPLE_PROMPTS_CSS = """
     transition: all 0.2s ease;
     width: 100%;
     height: auto;
-    min-height: 87px;
+    min-height: 89px;
     display: flex;
     align-items: flex-start;
     cursor: pointer;


tschaffter (changes requested):
The text in the box wrap nicely and are not cropped up to a point when reducing the width of the page, but then the examples disappear. This means that they won't be visible on smartphones. At some point, the boxes may be better displayed one on top of another but should never disappear.

Image

Image

PR #3687, opened 2025-11-14
feat(model-ad): add caching for pinned items in CT (MG-471)

sagely1:
## Description
This PR adds caching to pinned items in the CT.  Composite key changes will be addressed in [MG-525](https://sagebionetworks.jira.com/browse/MG-525)

## Related Issue
[MG-471](https://sagebionetworks.jira.com/browse/MG-471)

## Changelog
- Add caching so changes in dropdown selection within each CT are preserved
- Changes when switching between CTs are also preserved

## Preview
https://www.loom.com/share/79edc170d45b4cd295b2d7e24e5494f1

[MG-471]: https://sagebionetworks.jira.com/browse/MG-471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MG-525]: https://sagebionetworks.jira.com/browse/MG-525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/services/src/lib/comparison-tool.service.ts
+++ b/libs/explorers/services/src/lib/comparison-tool.service.ts
@@ -156,10 +156,12 @@ export class ComparisonToolService<T> {
 
     this.configsSignal.set(configs ?? []);
     this.totalResultsCount.set(0);
+    if (this.hasInitializedConfig) {
+      this.resetPinnedItems();
+    }
     this.multiSortMetaSignal.set(this.DEFAULT_MULTI_SORT_META);
     this.setUnpinnedData([]);
     this.setPinnedData([]);
-    this.resetPinnedItems();
 
     const normalizedSelection = this.normalizeSelection(selection ?? [], configs);
     this.dropdownSelectionSignal.set(normalizedSelection);


libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -160,13 +162,13 @@ export class ComparisonToolService<T> {
...
     this.columnsForDropdownsSignal.set(columnsMap);
+    this.pinnedItemsForDropdownsSignal.set(new Map());
   }
 
   setDropdownSelection(selection: string[]) {
     const configs = this.configsSignal();
     if (!configs.length) {
-      const normalizedSelection = selection ?? [];
-      this.updateDropdownSelectionIfChanged(normalizedSelection);
+      this.dropdownSelectionSignal.set(selection ?? []);
```
hallieswan: Similar question here as well!

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -137,19 +138,20 @@ export class ComparisonToolService<T> {
...
     this.totalResultsCount.set(0);
-    this.resetPinnedItems();
     this.multiSortMetaSignal.set(this.DEFAULT_MULTI_SORT_META);
     this.setUnpinnedData([]);
     this.setPinnedData([]);
+    this.resetPinnedItems();
 
     if (!configs?.length) {
-      this.updateDropdownSelectionIfChanged([]);
+      this.dropdownSelectionSignal.set([]);
```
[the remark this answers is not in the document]
hallieswan: Per our discussion, we should move the guard from [each CT](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/model-ad/disease-correlation-comparison-tool/src/lib/disease-correlation-comparison-tool.component.ts#L118-L121) to `initialize`, so it's clear the `initialize` is only meant to be called once. So in this case, we can stick with the updated code (`this.dropdownSelectionSignal.set([]);`), since it should only be called once.

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -148,23 +149,20 @@ export class ComparisonToolService<T> {
...
+    // if it is already initialized, do not re-initialize
+    if (this.configs().length > 0) {
+      return;
+    }
+
     this.configsSignal.set(configs ?? []);
     this.totalResultsCount.set(0);
-    if (this.hasInitializedConfig) {
-      this.resetPinnedItems();
-    }
```
hallieswan: I think this might be an artifact from merging in #3687, but we need to keep this check to prevent `initialize` from clearing the pinned items parsed from the URL parameters

hallieswan: However, the e2e test that check that items from the URL are pinned passed, when I expected it to fail, so I'm investigating now!

hallieswan: Adding some context about why the e2e test passed, when it should fail: 
- The CT service clears pinned items only after URL syncing finishes. The URL sync waits on a debounced route params stream.
- In most e2e runs, the config fetch resolves before that stream emits, so `initialize` runs while `isInitialized` is still `false` and the effect that clears pins never fires.
- Manual repro slow the config response long enough for the query params to land first, flipping `isInitialized` and triggering the bug consistently.
We should fix the e2e test so that it reliably catches this bug and update the code to remove the race condition. I'll tackle this in a follow-up PR. Tracking in [MG-532](https://sagebionetworks.jira.com/browse/MG-532).

PR #3703, opened 2025-11-19
feat(model-ad): add pagination support in API and UI (MG-557, MG-558, MG-559)

sagely1:
## Description
Add pagination support in the Java API for Model Overview & Disease Correlation CTs.  Also add support for pagination in the UI.

## Related Issue

[MG-557](https://sagebionetworks.jira.com/browse/MG-557)
[MG-558](https://sagebionetworks.jira.com/browse/MG-558)
[MG-559](https://sagebionetworks.jira.com/browse/MG-559)

## Changelog
- Add pagination support - 10 rows by default
- Add pagination support for Model Overview CT
- Add pagination support for Disease Correlation CT
- Add support in the PrimeNG tables to support pagination (as the api now returns a max of 10 by default)
- Add Caddyfile changes to use api-next

## Preview
The screenshot below shows 2 pinned items in the pinned table and 14 items in the unpinned table.  Displayed Results shows 16 as expected.
image

[MG-557]: https://sagebionetworks.jira.com/browse/MG-557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MG-558]: https://sagebionetworks.jira.com/browse/MG-558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MG-559]: https://sagebionetworks.jira.com/browse/MG-559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/model-ad/api-description/src/paths/comparison-tools/model-overview.yaml
```diff
@@ -2,20 +2,20 @@ get:
...
     - Model Overview
   summary: Get model overview for comparison tools
-  description: Returns a list of model overview objects for use in comparison tools.
+  description: Returns a paginated list of model overview objects for use in comparison tools.
   operationId: getModelOverviews
   parameters:
     - $ref: ../query/itemQuery.yaml
     - $ref: ../query/itemFilterTypeQuery.yaml
+    - $ref: ../query/pageNumberQuery.yaml
+    - $ref: ../query/pageSizeQuery.yaml
```
tschaffter: The approach taken in OC and BixArena is to specify a single "*Query" object that centralizes all the query parameters for an endpoint, including page size and number.See this example: https://github.com/Sage-Bionetworks/sage-monorepo/blob/d9902a6ac12e20fe14b44dcec138f2391f00d06b/libs/bixarena/api-description/src/paths/example-prompts.yaml#L9

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/service/DiseaseCorrelationService.java
```diff
@@ -40,26 +47,41 @@ public List<DiseaseCorrelationDto> loadDiseaseCorrelations(
...
+      .number(page.getNumber())
+      .size(page.getSize())
+      .totalElements(page.getTotalElements())
+      .totalPages(page.getTotalPages())
+      .hasNext(page.hasNext())
+      .hasPrevious(page.hasPrevious());
+
+    return new DiseaseCorrelationsPageDto()
+      .diseaseCorrelations(diseaseCorrelations)
+      .page(pageMetadata);
```
tschaffter: 1) DiseaseCorrelationsPageDto extends PageMetadataDto, so you could directly build it instead of creating an instance of PageMetadataDto. See this example: https://github.com/Sage-Bionetworks/sage-monorepo/blob/d9902a6ac12e20fe14b44dcec138f2391f00d06b/apps/bixarena/api/src/main/java/org/sagebionetworks/bixarena/api/service/ExamplePromptService.java#L50-L58

2) Prefer using the `.builder()` pattern for consistency across the products instead of using the `new` operator. See the above example.

libs/explorers/comparison-tool/src/lib/comparison-tool-table/base-table/base-table.component.ts
```diff
@@ -39,6 +39,9 @@ export class BaseTableComponent {
   shouldPaginate = input<boolean>(true);
   shouldShowNoDataMessage = input<boolean>(true);
   columnWidth = input<string>('auto');
+  totalRecords = input<number | undefined>(undefined);
```
hallieswan: We already inject the CT service in this component -- can we get this value directly from the CT service instead of using prop drilling? 


```suggestion
  totalRecords = this.comparisonToolService.totalResultsCount;
```

libs/model-ad/model-overview-comparison-tool/src/lib/model-overview-comparison-tool.component.ts
```diff
@@ -106,39 +112,55 @@ export class ModelOverviewComparisonToolComponent implements OnInit {
...
-          this.isLoading.set(false);
-        },
       });
   }
 
   getPinnedData(pinnedItems: string[]) {
+    const query: ModelOverviewSearchQuery = {
+      items: pinnedItems,
+      itemFilterType: ItemFilterTypeQuery.Include,
+    };
```
hallieswan: Should we rely on the route's default `pageSize` and `pageNumber` here or should we use explicit values for pinned data? e.g. pageNumber = 0, pageSize = this.comparisonToolService.maxPinnedItems

libs/model-ad/api-description/src/paths/query/pageNumberQuery.yaml
```diff
@@ -0,0 +1,8 @@
+name: pageNumber
+in: query
+required: false
+description: The page number.
```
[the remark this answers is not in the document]
sagely1: I'll add the 0-based index note to ModelOverviewSearchQuery.yaml

libs/explorers/comparison-tool/src/lib/comparison-tool-table/base-table/base-table.component.ts
```diff
@@ -51,4 +54,8 @@ export class BaseTableComponent {
   sortCallback(event: SortEvent) {
     this.comparisonToolService.setSort(event);
   }
+
+  onLazyLoad(event: TableLazyLoadEvent) {
+    this.lazyLoad.emit(event);
+  }
```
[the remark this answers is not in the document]
sagely1: Good idea.  I'll implement a callback instead.

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -448,17 +472,16 @@ export class ComparisonToolService<T> {
       this.lastSerializedState = JSON.stringify(this.serializeState(this.pinnedItems()));
 
       this.syncToUrlInProgress.set(false);
-      this.isInitialized.set(true);
+      this.isInitializedSignal.set(true);
```
hallieswan: We need to move `this.isInitialzedSignal.set(true)` to the end of the `initialize` method

libs/model-ad/model-overview-comparison-tool/src/lib/model-overview-comparison-tool.component.ts
hallieswan: The service does paginate the pinned results, but just uses the default values for the pageNumber (0) and pageSize (100).  Since the default maxPinnedItems (50) is less than the default pageSize, we always return all the pinned items, because they are all in the first page. However, if the maxPinnedItems is larger than the default pageSize, then we wouldn't return the full set of pinned items. Since the default pageSize is larger than the default maxPinnedItems, I agree it isn't an issue right now and we can defer any changes for now.

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -34,7 +37,7 @@ export class ComparisonToolService<T> {
     filterResultsButtonTooltip: 'Filter results',
     showSignificanceControls: true,
     viewDetailsTooltip: 'View detailed results',
-    viewDetailsClick: (id: string, label: string) => {
+    viewDetailsClick: () => {
```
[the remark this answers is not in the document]
sagely1: I am having the same issue except I must've forgot to save it when I reverted it.

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -334,6 +343,24 @@ export class ComparisonToolService<T> {
...
+  setPageNumber(pageNumber: number) {
+    this.pageNumberSignal.set(pageNumber);
+  }
+
+  setPageSize(pageSize: number) {
+    this.pageSizeSignal.set(pageSize);
+  }
+
+  handleLazyLoad(event: TableLazyLoadEvent) {
+    const defaultRowsPerPage = this.viewConfigSignal().rowsPerPage || 10;
```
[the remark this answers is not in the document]
sagely1: Good catch.  Now that I've added the viewConfig variable, there's no need for a fallback.

PR #3711, opened 2025-11-21
feat(bixarena): show the latest visible leaderboard (SMR-610)

rrchai:
## Description

Refines leaderboard loading so it fetches only on navigation, hides content when empty; adds support for latest visible snapshot, refetch after reload, richer entry metadata (organization/url), and sorting by `rank`/`btScore`.

## Related Issue

[SMR-610](https://sagebionetworks.jira.com/browse/SMR-610)

## Changelog

- Fetch latest visible leaderboard snapshot and refetch after reload
- Add `modelOrganization` and `modelUrl` fields to leaderboard entries
- Allow sorting by `rank` with `btScore` tie-break
- Refresh leaderboard only when visiting the page and toggle placeholder/table based on data
- Add custom tooltip with icon to briefly describe Score column and remove the tooltips on rest of columns

### Create a Leaderboard for Testing

Review the **Preview** in #3702 to generate a dummy snapshot and update the visibility of snapshot for testing.

### Leaderboard Page

Screen Shot 2025-11-21 at 7 30 35 PM

### Tooltip on "Score" Column

Screen Shot 2025-11-21 at 7 30 48 PM

### Hyperlink to the Model Page of OpenRouter
Screen Shot 2025-11-21 at 7 31 00 PM

### Search on "Model" and "Organization"
Screen Shot 2025-11-21 at 7 31 20 PM

[SMR-610]: https://sagebionetworks.jira.com/browse/SMR-610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/bixarena/api/src/main/java/org/sagebionetworks/bixarena/api/service/LeaderboardService.java
+++ b/apps/bixarena/api/src/main/java/org/sagebionetworks/bixarena/api/service/LeaderboardService.java
@@ -155,7 +155,11 @@ public class LeaderboardService {
 
     if ("rank".equals(sortField)) {
       // Create primary rank sort, and add secondary btScore sort
-      return Sort.by(direction, entityField).and(Sort.by(Sort.Direction.DESC, "btScore"));
+      // Since rank is inversely proportional to btScore, use opposite direction
+      Sort.Direction btScoreDirection = direction == Sort.Direction.ASC
+        ? Sort.Direction.DESC
+        : Sort.Direction.ASC;
+      return Sort.by(direction, entityField).and(Sort.by(btScoreDirection, "btScore"));
     }
 
     return Sort.by(direction, entityField);
--- a/apps/bixarena/app/bixarena_app/page/bixarena_leaderboard.py
+++ b/apps/bixarena/app/bixarena_app/page/bixarena_leaderboard.py
@@ -205,7 +205,7 @@ def build_leaderboard_page():
                     font-size: var(--text-xl);
                     color: var(--body-text-color-subdued);
                     margin-bottom: 40px !important;
-                ">Community-driven evaluation of biomedical AI models</p>
+                ">Community-driven evaluation of AI models on biomedical topics</p>
                 """
             )
 


libs/bixarena/api-description/src/components/schemas/LeaderboardEntry.yaml
```diff
@@ -13,6 +13,17 @@ properties:
...
+    type: string
+    maxLength: 200
+    description: Organization that created the model
+    example: 'OpenAI'
+    nullable: true
+  modelUrl:
+    type: string
+    maxLength: 300
+    description: External link to model information
+    example: 'https://openai.com/gpt-4'
```
rrchai: add `modelOrganization` and `ModelUrl` properties for leaderboard display

libs/bixarena/api-description/src/components/schemas/LeaderboardSort.yaml
```diff
@@ -2,9 +2,6 @@ type: string
 description: The field to sort leaderboard entries by.
 enum:
   - rank
-  - bt_score
-  - vote_count
   - created_at
-  - model_name
 default: rank
-example: bt_score
+example: rank
```
rrchai: By default, the leaderboard will be sorted by rank (+ secondary sort  `btScore` processed in backend); also remove unused sorting.

apps/bixarena/api/src/main/java/org/sagebionetworks/bixarena/api/service/LeaderboardService.java
```diff
@@ -147,13 +149,15 @@ private Sort createSort(LeaderboardSearchQueryDto query) {
...
-        case "bt_score" -> "btScore";
-        case "vote_count" -> "voteCount";
         case "created_at" -> "createdAt";
-        case "model_name" -> "model.name";
         default -> "rank";
       };
 
+    if ("rank".equals(sortField)) {
+      // Create primary rank sort, and add secondary btScore sort
+      return Sort.by(direction, entityField).and(Sort.by(Sort.Direction.DESC, "btScore"));
```
tschaffter: Since the rank is inversely proportional to btScore, we should likely use the specified `direction` for sorting btScore too so that we perfectly reverse the order when switching from one direction to the other one.

- If sort direction for rank is ASC, use DESC to sort btScore
- If sort direction for rank is DESC, use ASC to sort btScore

tschaffter:
Let's also replace the headline below the Leaderboard page title to avoid using the term "biomedical AI models". Instead, a more descriptive headline can refer to "AI models evaluated on biomedical topics".

PR #3803, opened 2026-01-20
fix(model-ad): prevent empty $addFields error when sorting by non-string fields (MG-709)

rrchai:
## Description

Fixes a bug where sorting by non-string fields caused DocumentDB erro: "Invalid $addFields :: caused by :: specification must have at least one field". The `buildLowercaseSortFields()` method was unconditionally adding an empty `$addFields` operation to the MongoDB aggregation pipeline when no string fields needed case-insensitive sorting. This fix ensures the $addFields operation is only added when there are actually string fields to transform.

## Related Issue

[MG-709](https://sagebionetworks.jira.com/browse/MG-709)

## Changelog

- Fix DocumentDB error by conditionally adding `$addFields` operation only when string fields need case-insensitive sorting
- Improve debug logging to display full query object instead of only categories for better troubleshooting

[MG-709]: https://sagebionetworks.jira.com/browse/MG-709?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hallieswan (approved):
Looks good to me!

PR #3804, opened 2026-01-21
fix(explorers): prevent duplicate API calls in comparison tool (MG-676)

hallieswan:
## Description

This PR fixes a bug where the comparison tool components were making duplicate API calls for `getPinnedData` and `getUnpinnedData` during initialization, category changes, and sort changes. The root cause was a circular update loop between component effects and URL synchronization, combined with multiple initialization paths for filters and redundant `handleLazyLoad` events from PrimeNG. The fix consolidates URL sync tracking, establishes a single source of truth for filter initialization, and guards against duplicate pagination events.

## Related Issue

Fixes [MG-676](https://sagebionetworks.jira.com/browse/MG-676)

## Changelog

- Fix duplicate API calls by consolidating URL sync state tracking into a single `lastSyncedState` object that detects self-triggered URL changes
- Add guard in `handleLazyLoad` to skip processing when pagination values haven't actually changed
- Move filter initialization from the filter panel component to the comparison tool service, establishing a single source of truth
- Remove redundant `filterConfigs` input prop drilling from comparison tool template and filter panel component
- Optimize `pinnedDataEffect` in comparison tool pages to track only the signals that affect pinned data fetching
- Extract shared `findConfigForSelection` and `convertFiltersFromConfig` helper methods to reduce code duplication

## Preview

2 calls on initial load (pinned, unpinned)
2 calls on pinning (pinned, unpinned)
2 calls on sort (pinned, unpinned)
1 call on filterbox search (unpinned)
2 calls on categories change (pinned, unpinned) 
1 call on filter selection (unpinned)
1 call on pagination (unpinned)

https://github.com/user-attachments/assets/68497d1e-75ea-48d3-a825-b69a182706d4

[MG-676]: https://sagebionetworks.jira.com/browse/MG-676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/services/src/lib/comparison-tool.service.ts
+++ b/libs/explorers/services/src/lib/comparison-tool.service.ts
@@ -97,7 +97,7 @@ export class ComparisonToolService<T> {
   private readonly isFilterPanelOpenSignal = signal(false);
 
   // URL Sync State
-  private lastSyncedState: ComparisonToolUrlParams | null = null;
+  private lastSyncedUrlParamsState: ComparisonToolUrlParams | null = null;
   private initialSelection: string[] | undefined;
 
   // Public Readonly Signals
@@ -115,8 +115,10 @@ export class ComparisonToolService<T> {
   // Computed Query Accessors
   readonly query = this.querySignal.asReadonly();
   readonly dropdownSelection = computed(() => this.querySignal().categories);
-  readonly pinnedItems = computed(() => new Set(this.querySignal().pinnedItems));
-  readonly pinnedItemsArray = computed(() => this.querySignal().pinnedItems);
+  // set used to quickly check if an item is pinned
+  readonly pinnedItemsSet = computed(() => new Set(this.querySignal().pinnedItems));
+  // array exposed separately from query() to allow effects to depend only on pinnedItems changes
+  readonly pinnedItems = computed(() => this.querySignal().pinnedItems);
   readonly pageNumber = computed(() => this.querySignal().pageNumber);
   readonly pageSize = computed(() => this.querySignal().pageSize);
   readonly multiSortMeta = computed(() => this.querySignal().multiSortMeta);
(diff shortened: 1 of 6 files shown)


libs/explorers/comparison-tool/src/lib/comparison-tool-filter-panel/comparison-tool-filter-panel.component.ts
```diff
@@ -27,32 +23,13 @@ import { FilterPanelMainMenuItemButtonComponent } from './filter-panel-main-menu
...
-            label: value,
-            selected: false,
-          })),
-        };
-        return filter;
-      });
-      this.comparisonToolFilterService.setFilters(filters);
-    });
-  }
-
```
hallieswan: Moved filter initialization to CT service to centralize initial query update

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -94,8 +96,7 @@ export class ComparisonToolService<T> {
   } | null>(null);
 
   // URL Sync State
-  private lastSerializedState: string | null = null;
-  private lastSyncedPins: string[] = [];
+  private lastSyncedState: ComparisonToolUrlParams | null = null;
```
hallieswan: Consolidate these variables into `lastSyncedState`, so that we can consolidate URL sync state tracking with an object that can also detect self-triggered URL changes from changes to pinned items, categories, or sort

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -498,6 +525,14 @@ export class ComparisonToolService<T> {
       event,
       defaultRowsPerPage,
     );
+
+    // Skip if pagination hasn't actually changed
+    const currentPageNumber = this.pageNumber();
+    const currentPageSize = this.pageSize();
+    if (pageNumber === currentPageNumber && pageSize === currentPageSize) {
+      return;
+    }
```
hallieswan: Guard against calling `updateQuery` when the pagination values haven't actually changed

sagely1 (approved):
LGTM!

PR #3815, opened 2026-01-28
fix(model-ad): fix breakpoint behavior in shared header (MG-730)

sagely1:
## Description
This fixes the header breakpoint being too low, causing UX issues with the search field in the header for Model-AD

## Related Issue
[MG-730](https://sagebionetworks.jira.com/browse/MG-730)

## Changelog
- Simplified header breakpoint logic to be only driven by code
- Set breakpoint at 1300px as a default (can be overridden)
- Updates Agora and Model-AD e2e tests reliant on previous breakpoint
- Fix flaky e2e tests

## Preview
Browser width of 1299px shows hamburger menu as expected
image

[MG-730]: https://sagebionetworks.jira.com/browse/MG-730?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/model-ad/app/e2e/model-details.spec.ts
+++ b/apps/model-ad/app/e2e/model-details.spec.ts
@@ -121,10 +121,13 @@ test.describe('model details', () => {
   test('does not scroll to panel content on initial load when tab not specified in url', async ({
     page,
   }) => {
+    await page.setViewportSize({ width: 1280, height: 720 });
     const model = '3xTg-AD';
     await page.goto(`/models/${model}`);
-
-    // Only check that the page is at the top (not scrolled)
+    await expect(page.getByRole('heading', { level: 1, name: model })).toBeInViewport();
+    await expect(
+      page.getByRole('heading', { level: 2, name: 'Available Data' }),
+    ).not.toBeInViewport();
     await expectPageAtTop(page);
   });
 


apps/agora/app/playwright.config.ts
```diff
@@ -37,40 +39,4 @@ export default defineConfig({
...
-    // Uncomment for branded browsers
-    /* {
-      name: 'Microsoft Edge',
-      use: { ...devices['Desktop Edge'], channel: 'msedge' },
-    },
-    {
-      name: 'Google Chrome',
-      use: { ...devices['Desktop Chrome'], channel: 'chrome' },
-    } */
-  ],
```
sagely1: this conflicts with the viewport setting above

apps/agora/app/playwright.config.ts
```diff
@@ -37,40 +39,4 @@ export default defineConfig({
     cwd: workspaceRoot,
     timeout: 60000 * 3, // give time for agora-data to populate agora-mongo
   },
-  projects: [
-    {
-      name: 'chromium',
-      use: { ...devices['Desktop Chrome'] },
-    },
```
hallieswan: These projects are useful if we need to run an e2e test against a particular browser, e.g. for a Firefox-specific bug. What do you think about creating a constant for the viewport, then setting the viewport at the project level? e.g. something like: 

```
DESKTOP_VIEWPORT = { width: 1600, height: 900 };
...
projects: [
    {
      name: 'chromium',
      use: { 
        ...devices['Desktop Chrome'] 
        viewport: DESKTOP_VIEWPORT,
     },
    },
```

sagely1: sounds good!

PR #3823, opened 2026-01-30
fix(agora): prevent layout shift from inactive tabs during scroll (AG-1948)

hallieswan:
## Description

The "View all brain regions" link scrolled to the correct header in the gene details RNA evidence tab, but then immediately scrolled a bit further so the header was off the screen. Investigation revealed that the extra scroll was due to inactive tabs that were hydrating and rendering concurrently during SSR, with the Summary tab's barchart component expanding asynchronously, which pushed active tab content downward. This PR eliminates the phantom scroll issue by ensuring only the active tab exists in the DOM, preventing background rendering from affecting visible content.

## Related Issue

[AG-1948](https://sagebionetworks.jira.com/browse/AG-1948)

## Changelog

- Ensures that "View all brain regions" link scrolls to correct position by replacing CSS-based tab visibility with Angular structural directives to render only active tab
- Add minimum height to tab container to prevent layout collapse during tab transitions

## Preview

https://github.com/user-attachments/assets/e62033f7-1b71-4e1e-8507-8caecaeda251

[AG-1948]: https://sagebionetworks.jira.com/browse/AG-1948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/model-ad/app/e2e/model-details.spec.ts
+++ b/apps/model-ad/app/e2e/model-details.spec.ts
@@ -1,4 +1,5 @@
 import { expect, Page, test } from '@playwright/test';
+import { waitForScrollToStop } from '@sagebionetworks/explorers/testing/e2e';
 import { baseURL } from '../playwright.config';
 import { searchAndGetSearchListItems } from './helpers/search';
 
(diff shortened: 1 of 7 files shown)


apps/model-ad/app/e2e/model-details.spec.ts
```diff
@@ -14,30 +15,6 @@ async function expectPageNotAtTop(page: Page) {
...
-        } else {
-          stableCount = 0;
-          previousOffset = currentOffset;
-          return false;
-        }
-      },
-      { timeout, intervals: [200] },
-    )
-    .toBe(true);
-}
```
hallieswan: Moved to explorers, so can be shared with agora

libs/agora/genes/src/lib/components/gene-details/gene-details.component.scss
```diff
@@ -152,23 +152,12 @@
...
-  opacity: 0;
-  visibility: hidden;
-  z-index: 100;
-
-  &.active {
-    opacity: 1;
-    visibility: visible;
-    position: relative;
-    top: auto;
-    z-index: 200;
```
hallieswan: Styles no longer needed, since only the active panel exists in the DOM

libs/agora/genes/src/lib/components/gene-details/gene-details.component.ts
```diff
@@ -109,7 +109,7 @@ export class GeneDetailsComponent implements OnInit, AfterViewInit, AfterViewChe
   activeParent = '';
   navSlideIndex = 0;
 
-  @HostListener('window:scroll', ['$event'])
+  @HostListener('window:scroll')
```
hallieswan: Fix `@HostListener` TypeScript errors by removing unused event parameters

PR #3831, opened 2026-02-03
infra(bixarena): deploy the API gateway before the web app

tschaffter:
## Description

This PR establishes an explicit deployment dependency between the BioArena web app and API gateway in the CDK infrastructure across all environments (dev, stage, prod). Previously, the web stack could deploy independently of the API gateway, leading to potential runtime errors when the web app attempted to make API calls before the gateway was ready. By adding this dependency, we ensure the API gateway is fully deployed and healthy before the web app container starts, improving deployment reliability and preventing API connectivity issues during stack updates.

## Changelog

- Add explicit CDK stack dependency so web stack waits for API gateway deployment
- Reorder stack creation to deploy API gateway before web app across all environments
- Move `use_https` variable initialization earlier to support backend service configuration

## Preview

The BioArena infrastructure now has:

- Guaranteed deployment ordering ensuring API gateway is ready before web app starts
- Improved reliability during stack updates and initial deployments
- Consistent deployment behavior across dev, stage, and production environments

### DIFF
--- a/apps/bixarena/infra/cdk/bixarena_infra_cdk/dev/app.py
+++ b/apps/bixarena/infra/cdk/bixarena_infra_cdk/dev/app.py
@@ -245,28 +245,8 @@ def main() -> None:
             "container at runtime. Never exposed in CloudFormation or logs."
         )
 
-    # Create web stack (depends on ECS cluster and ALB)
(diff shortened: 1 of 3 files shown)


rrchai (approved):
Tested deployment that Web stack is dependent on API-Gateway stack.

Image

Image

PR #3832, opened 2026-02-03
refactor(bixarena): dynamically update the quest content

rrchai:
## Description

This PR refactors the Community Quest section to dynamically update its content on every page load instead of only at app startup. Previously, quest progress, contributors list, and carousel images were rendered once when the application started, requiring a full app restart to reflect new data. Now all three components refresh automatically when users reload the page, ensuring they always see the latest quest information without restarting the server. 

## Changelog

- Rename `load_quest_progress_on_page_load` to `load_quest_content_on_page_load` to reflect expanded scope
- Extract carousel and contributors HTML builders into reusable helper functions for dynamic updates
- Update page load handler to refresh progress, contributors, and carousel HTML on every page load
- Add early return optimization to skip HTML generation when quest doesn't exist
- Change carousel ID from dynamic timestamp-based to static to preserve JavaScript rotation state
- Update return signatures to expose progress, contributors, and carousel containers for dynamic updates

### DIFF
--- a/apps/bixarena/app/bixarena_app/page/bixarena_home.py
+++ b/apps/bixarena/app/bixarena_app/page/bixarena_home.py
@@ -271,6 +271,10 @@ def load_quest_content_on_page_load() -> tuple[dict, dict, dict]:
         # Fetch contributors (single API call for all data)
         contributors_data = fetch_quest_contributors(QUEST_CONFIG["quest_id"])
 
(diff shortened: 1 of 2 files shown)


tschaffter:
While still in Draft mode, I tried the PR and it successfully enable the refresh of the community quest. Below is a screenshot taken from my local stack after completing a battle and waiting 1-2 minutes for the contributors cache to refresh.

image

PR #3842, opened 2026-02-06
feat(model-ad): improve api error handling (MG-699)

sagely1:
## Description

This PR improves error handling across the Model-AD application by adding a persistent error overlay to all pages (except the error page) that allows users to reload the page. When API errors occur, users now see a centered warning banner over a dimmed overlay, prompting them to reload the page. This provides a clearer, more intentional error recovery experience compared to the previous implementation of error toasts. The HTTP error interceptor has been refactored to use smart retry logic (only retrying transient errors) before showing the error overlay.

## Related Issue

[MG-699](https://sagebionetworks.jira.com/browse/MG-699)
[AG-1787](https://sagebionetworks.jira.com/browse/AG-1787)

## Changelog
- Add new ErrorOverlayService that displays a persistent overlay with warning banner when errors occur, requiring users to click "Reload Page" to recover
- Refactor NotificationService into ToastNotificationService for non-blocking toast notifications (warnings, success, info) and ErrorOverlayService for blocking error states
- Update HTTP error interceptor with smart retry logic that only retries transient errors (network issues, 5xx) and skips retrying client errors (4xx)
- Update Model-AD comparison tool components to throw with user-friendly messages that trigger the error overlay
- Remove deprecated Agora http-error-interceptor.ts in favor of shared explorers interceptor
- Replace "Go Home" button with "Reload" button on error page component
- This PR may fix Agora the ErrorEvent error in [AG-1787](https://sagebionetworks.jira.com/browse/AG-1787)

## Preview
Error overlay
image

image

Different resolutions shown to illustrate how the error message displays when the api goes down
https://www.loom.com/share/d6a88af9a7a74782a1df147828c0d6a8

## Testing
To test, stop the api server for Agora/Model-AD to see the error behavior.  Note: For Agora, we redirect on errors but Model-AD will show the overlay if the error occurs after a successful navigation and api call to a CT route.  In either case, if the user navigates to a different route and the api is down, we redirect to the error page.

[MG-699]: https://sagebionetworks.jira.com/browse/MG-699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AG-1787]: https://sagebionetworks.jira.com/browse/AG-1787?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/model-ad/app/src/app/app.component.html
+++ b/apps/model-ad/app/src/app/app.component.html
@@ -4,7 +4,7 @@
 <explorers-error-overlay />
 <div id="container">
   <explorers-header
-    headerLogoPath="/model-ad-assets/images/header-logo.svg"
+    headerLogoPath="model-ad-assets/images/header-logo.svg"
     [headerLinks]="headerLinks"
     [footerLinks]="footerLinks"
   >
@@ -14,7 +14,7 @@
     <router-outlet></router-outlet>
   </div>
   <explorers-footer
-    footerLogoPath="/model-ad-assets/images/footer-logo.svg"
+    footerLogoPath="model-ad-assets/images/footer-logo.svg"
     [footerLinks]="footerLinks"
     [siteVersion]="siteVersion"
     [dataVersion]="dataVersion"
--- a/apps/agora/app/src/app/app.component.html
+++ b/apps/agora/app/src/app/app.component.html
@@ -4,7 +4,7 @@
 <explorers-error-overlay />
 <div id="container">
   <explorers-header
-    headerLogoPath="/agora-assets/images/header-logo.svg"
+    headerLogoPath="agora-assets/images/header-logo.svg"
     [headerLinks]="headerLinks"
     [footerLinks]="footerLinks"
   >
@@ -14,7 +14,7 @@
     <router-outlet></router-outlet>
   </div>
   <explorers-footer
-    footerLogoPath="/agora-assets/images/footer-logo.svg"
+    footerLogoPath="agora-assets/images/footer-logo.svg"
     [footerLinks]="footerLinks"
     [siteVersion]="siteVersion"
     [dataVersion]="dataVersion"
--- a/apps/agora/app/src/app/app.routes.ts
+++ b/apps/agora/app/src/app/app.routes.ts
@@ -1,5 +1,9 @@
 import { Route } from '@angular/router';
-import { ROUTE_PATHS, SUPPORT_EMAIL } from '@sagebionetworks/agora/config';
+import {
+  DEFAULT_HERO_BACKGROUND_IMAGE_PATH,
+  ROUTE_PATHS,
+  SUPPORT_EMAIL,
+} from '@sagebionetworks/agora/config';
 
 export const routes: Route[] = [
   {
@@ -108,7 +112,7 @@ export const routes: Route[] = [
       title: 'Agora | Terms of Service',
       description:
         'Agora is powered by Synapse, a platform for supporting scientific collaborations centered around shared biomedical data sets. Our goal is to make biomedical research more transparent, more reproducible, and more accessible to a broader audience of scientists.',
-      heroBackgroundImagePath: 'agora-assets/images/hero-background.svg',
+      heroBackgroundImagePath: DEFAULT_HERO_BACKGROUND_IMAGE_PATH,
     },
   },
   {
--- a/apps/model-ad/app/src/app/content/visualization-overview.content.ts
+++ b/apps/model-ad/app/src/app/content/visualization-overview.content.ts
@@ -6,24 +6,24 @@ export const VISUALIZATION_OVERVIEW_PANES: VisualizationOverviewPane[] = [
     content: `<p>Comparison Tools allow you to discover, compare and share results and related information. While each Comparison Tool presents different kinds of information, they all support similar exploration features. This tutorial demonstrates how you can use these features to find, download, and share results.</p>
       <p>You can always revisit this tutorial by clicking the Visualization Overview link at the bottom of the page.</p>
       <p>If the tool you're using has a legend, you can view it by clicking on the Legend link at the bottom of the page.</p>
-      <div class="image-container"><img src="/model-ad-assets/images/visualization-overview-1.png" /></div>`,
+      <div class="image-container"><img src="model-ad-assets/images/visualization-overview-1.png" /></div>`,
   },
   {
     heading: 'Search & Filter',
     content: `<p>Quickly find results of interest using the tool’s embedded search box. Enter a value to get a list of matching results, or perform a bulk search using a list of comma-separated full values.</p>
       <p>You can also use the tool’s supported filters to find results of interest. The filter region near the top of the page shows which filters are currently applied. You can clear specific filters, or all of the filters at once.</p>
-      <div class="video-container"><video autoplay muted loop playsinline preload="none" controls><source src="/model-ad-assets/videos/visualization-overview-2.mp4" type="video/mp4"></video></div>`,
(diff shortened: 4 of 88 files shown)


libs/explorers/util/src/lib/http-error-interceptor/http-error.interceptor.ts
```diff
@@ -5,67 +5,85 @@ import {
...
+  }
+
+  // Server-side error
+  switch (error.status) {
+    case 0:
+      return 'Unable to connect to the server. Please check your connection.';
+    case 400:
+      return 'Invalid request. Please check your input and try again.';
+    case 401:
+      return 'You are not authorized. Please log in and try again.';
```
hallieswan: The explorers don't have auth yet -- should we update this message?

apps/model-ad/app/src/app/app.component.html
```diff
@@ -1,6 +1,16 @@
...
 }
+@if (hasActiveError()) {
+  <div class="error-overlay"></div>
+  <div class="warning-banner">
+    <div class="warning-banner-content">
+      <i class="pi pi-exclamation-triangle warning-icon"></i>
+      <span>Something went wrong. Please reload the page.</span>
+      <button class="reload-button" (click)="reloadPage()">Reload Page</button>
+    </div>
+  </div>
```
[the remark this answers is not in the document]
hallieswan: Actually, by removing Agora's http-error-interceptor, we won't automatically display a toast for any HTTP error -- instead, only components with explicit error handlers will show errors (or redirect to the error page, like the GCT). What do you think about adding the error overlay to Agora as well, so we can display an error on any HTTP error?

apps/model-ad/app/src/app/app.component.html
```diff
@@ -1,6 +1,16 @@
 @if (useGoogleTagManager) {
   <sage-google-tag-manager />
 }
+@if (hasActiveError()) {
+  <div class="error-overlay"></div>
+  <div class="warning-banner">
+    <div class="warning-banner-content">
+      <i class="pi pi-exclamation-triangle warning-icon"></i>
+      <span>Something went wrong. Please reload the page.</span>
```
hallieswan: Can we display the specific message from the error overlay service rather than the generic message here? Right now, the user-facing messages specified in the CT route calls aren't displayed

libs/explorers/util/src/lib/http-error-interceptor/http-error.interceptor.ts
```diff
@@ -5,67 +5,85 @@ import {
...
+    }),
+    catchError((error: HttpErrorResponse) => {
+      const errorMessage = buildErrorMessage(error);
 
-      // TODO implement Rollbar service
-      // https://sagebionetworks.jira.com/browse/MG-230
-      // Send the error message to Rollbar
-      // rollbar.error(error);
+      // Log error for debugging
+      logger.error(`HTTP Error: ${errorMessage}`, error);
```
hallieswan: Should we write to rollbar here? If not as part of this PR, should we preserve the TODO to implement the Rollbar service tracked in https://sagebionetworks.jira.com/browse/MG-230?

hallieswan: Or alternately, should the logger service itself write to Rollbar?

libs/agora/services/src/lib/http-error-interceptor.ts
```diff
@@ -1,70 +0,0 @@
...
-        detail: errorDetail,
-        life: 3000,
-      });
-
-      setTimeout(() => {
-        messageService.clear();
-      }, 3000);
-
-      // Send the error message to Rollbar
-      rollbar.error(error);
```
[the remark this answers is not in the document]
JessterB: I don't find rollbar to be particularly useful and I doubt anyone else ever looks at it...

libs/explorers/services/src/lib/toast-notification.service.ts
```diff
@@ -1,14 +1,13 @@
...
 
+const TOAST_DURATION_MS = 5000;
+
 @Injectable({ providedIn: 'root' })
-export class NotificationService {
+export class ToastNotificationService {
   private readonly messageService = inject(MessageService);
 
   showError(message: string) {
-    this.messageService.clear();
```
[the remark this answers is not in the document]
hallieswan: Got it! There is still the issue of earlier toasts' `setTimeout` callbacks inadvertently clearing later toasts -- would you mind opening a ticket to capture this issue, so we can track as tech debt?

libs/explorers/util/src/lib/http-error-interceptor/http-error.interceptor.ts
```diff
@@ -5,31 +5,33 @@ import {
...
+import { ErrorOverlayService, LoggerService } from '@sagebionetworks/explorers/services';
 
 /**
  * HTTP interceptor that handles errors from HTTP requests.
  *
  * This interceptor:
  * - Retries failed requests once for transient errors (network issues, 5xx)
  * - Does NOT retry client errors (4xx) as they won't succeed
- * - Logs errors for debugging
+ * - Shows error overlay for connectivity errors (network failures, 5xx server errors)
```
hallieswan: Are other types of HTTP errors surfaced if they occur, e.g. a 404? If not, should we just default to showing the overlay on any error so the user is informed if a request doesn't complete?

libs/explorers/services/src/lib/version.service.ts
```diff
@@ -24,56 +24,33 @@ export interface DataVersionService {
...
-          },
-          error: (error) => {
-            console.error('Error loading data version:', error);
-            if (onError) {
-              onError(error);
-            }
-          },
-        });
+  getDataVersion$(dataVersionService: DataVersionService): Observable<string> {
+    if (!this.platformService.isBrowser) {
```
hallieswan: Just a nit, but I think we've been trying to use the positive check, rather than the inversion, for these types statements: 


```suggestion
    if (this.platformService.isServer) {
```

libs/explorers/services/src/lib/version.service.ts
```diff
@@ -24,56 +24,33 @@ export interface DataVersionService {
...
-        console.error('Error loading commit SHA:', error);
-        const fallbackVersion = this.formatSiteVersion('', config);
-        if (fallbackVersion) {
-          onSuccess(fallbackVersion);
-        } else {
-          console.error('Error loading appVersion');
-          if (onError) {
-            onError(error);
-          }
-        }
```
hallieswan: If the GitHub API call failed, the old code would gracefully degrade to showing just the `appVersion`. The new code throws an error, leaving the users without version info. Would it be better to keep the old code here? Then a failed GitHub api call wouldn't be a showstopper where the error overlay was surfaced.

libs/explorers/ui/src/lib/components/error-overlay/error-overlay.component.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  it('should call reloadPage when button is clicked', async () => {
+    const mockService = createMockErrorOverlayService(true, 'Error');
+    await render(ErrorOverlayComponent, {
+      providers: [{ provide: ErrorOverlayService, useValue: mockService }],
+    });
+
+    const reloadButton = screen.getByRole('button', { name: 'Reload Page' });
+    fireEvent.click(reloadButton);
```
hallieswan: I think it's generally [preferred](https://testing-library.com/docs/user-event/intro/) to use `userEvent` over `fireEvent` to more closely simulate user interaction -- could these tests be updated to use something like `await user.click(button)` instead?

PR #3847, opened 2026-02-10
IT-4711: Add CDK script to deploy zone

xschildw:
## IT-4711

Add a CDK script to bootstrap a hosted zone for dev.bioarena.io.
This script is run once to create the hosted zone, then the nameservers
are manually added to the parent zone (bioarena.io) in the SageIT account.
Going forward, the zone can be used to create DNS records for developer's
stacks.

## Note
Since this is a one-time setup, I kept it out of the existing nx infra.
I placed the code in a subdirectory 'bootstrap' in the 'infra' for the 'bixarena' app since it's related to that app.


rrchai:
Closing for now as we don't use dev much and don't need the hosted zone at this point. We will visit and request again if needed - Thanks for putting this together @xschildw.

PR #3969, opened 2026-04-07
feat(model-ad): cache non-default page size on CTs (MG-782, MG-837)

sagely1:
## Description

Persists the user's rows-per-page selection on comparison tool paginators using localStorage, so the preference survives page refreshes and works across tabs. As part of this change, the AppCookieService (backed by ngx-cookie-service) is replaced with a new AppStorageService backed entirely by localStorage, since all stored preferences are purely client-side and were never read during SSR. This also removes the ngx-cookie-service dependency from the project.

## Related Issue

[MG-782](https://sagebionetworks.jira.com/browse/MG-782)
[MG-837](https://sagebionetworks.jira.com/browse/MG-837)

## Changelog

- Add AppStorageService to replace AppCookieService, migrating all user preferences from cookies to localStorage with SSR guards and error handling
- Add LocalStorageService with SSR guard and error handling that can be reused across the SMR
- Persist the selected rows-per-page value (comparison_tool_page_size) in localStorage, validated on read and write against the allowed options [10, 25, 50]
- Define pageSizeOptions as a single source of truth in AppStorageService, replacing the hardcoded [10, 25, 50] in the footer template
- Remove ngx-cookie-service dependency from package.json
- Clean up invalid or legacy stored values on read
- Fix eslintrc and tsconfig inconsistencies 

## Preview

Checkbox and tooltip reverted as seen below:
image

Video shows tooltip reverted and refreshes preserving the user's preference on the rows-per-page setting.
https://www.loom.com/share/7b9b1434e6d946b8815e20d22c75f6b6

[MG-782]: https://sagebionetworks.jira.com/browse/MG-782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SMR-731]: https://sagebionetworks.jira.com/browse/SMR-731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MG-837]: https://sagebionetworks.jira.com/browse/MG-837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/explorers/services/src/lib/app-storage.service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    try {
+      const stored = localStorage.getItem(this.PAGE_SIZE_KEY);
+      if (stored === null) {
+        return this.DEFAULT_PAGE_SIZE;
+      }
+      const parsed = Number(stored);
+      if (this.VALID_PAGE_SIZES.includes(parsed)) {
+        return parsed;
+      }
```
sagely1: check for page sizes that are valid

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -60,7 +60,7 @@ export class ComparisonToolService<T> {
       sizeChartUpperLabel: '',
       sizeChartText: '',
     },
-    rowsPerPage: 10,
+    rowsPerPage: this.appStorageService.getPageSize(),
```
hallieswan: Could consider moving `this.appStorageService.getPageSize()` into a constant that can be used to initialize `DEFAULT_VIEW_CONFIG` and `querySignal`, e.g. something like `INITIAL_PAGE_SIZE`

rrchai (approved):
LGTM, just two nitpick suggestions.

hallieswan (changes requested):
Looking good! Just a couple of suggestions. 

Also, just wanted to double check the expected behavior for caching the page size -- when I set the page size in a particular CT (e.g. Model Overview), the page size is "remembered" when returning to that CT. However, the default page size is still used for other CTs (e.g. Disease Correlation). The ticket states that the cached page size should be used for all CTs -- did we decide against supporting the global cache? If not, we might need to add a `pageSize` signal to the `AppStorageService`, so each CT service can get the latest page size

libs/explorers/services/src/lib/app-storage.service.ts
```diff
@@ -0,0 +1,83 @@
...
+  providedIn: 'root',
+})
+export class AppStorageService {
+  private readonly platformService = inject(PlatformService);
+
+  private readonly HIDE_VISUALIZATION_OVERVIEW_KEY = 'hide_visualization_overview';
+  private readonly PAGE_SIZE_KEY = 'comparison_tool_page_size';
+  private readonly VALID_PAGE_SIZES: number[] = [10, 25, 50];
+  readonly pageSizeOptions: number[] = this.VALID_PAGE_SIZES;
+  private readonly DEFAULT_PAGE_SIZE = this.VALID_PAGE_SIZES[0];
```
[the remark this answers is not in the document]
sagely1: Sure, I can make that part of this PR.

libs/explorers/services/src/lib/comparison-tool.service.ts
sagely1: Since we are getting the user's preference for page size (if it exists), I think it is clearer that it is retrieving from the service vs. setting a default which the service also handles.

libs/web-shared/angular/logger/src/lib/logger.ts
```diff
@@ -0,0 +1,12 @@
+import { InjectionToken } from '@angular/core';
+
+export interface Logger {
+  error(message: string, error?: unknown): void;
+}
+
+export const LOGGER = new InjectionToken<Logger>('Logger', {
+  providedIn: 'root',
+  factory: () => ({
+    error: (message: string, error?: unknown) => console.error(message, error),
```
sagely1: For now, this is the only method a Logger needs

libs/web-shared/angular/storage/src/lib/local-storage.service.ts
```diff
@@ -0,0 +1,44 @@
+import { isPlatformBrowser } from '@angular/common';
```
sagely1, line 1: LocalStorageService moved to a centralized location in SMR

libs/web-shared/angular/logger/src/lib/logger.ts
```diff
@@ -0,0 +1,16 @@
+import { InjectionToken } from '@angular/core';
```
sagely1, line 1: Logger also moved to a centralized location in the SMR

libs/web-shared/angular/logger/jest.config.ts
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable */
+export default {
+  displayName: 'web-shared-angular-logger',
+  preset: '../../../../jest.preset.js',
+  setupFilesAfterSetup: ['<rootDir>/src/test-setup.ts'],
```
rrchai:
```suggestion
  setupFilesAfterEnv: ['<rootDir>/src/test-setup.ts'],
```

libs/web-shared/angular/storage/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
```
rrchai, line 1: For consistency, the structure of this file should match the others'

rrchai (changes requested):
Just two small consistency suggestions.

rrchai (approved):
lgtm

libs/web-shared/angular/storage/.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+        "plugin:jest/recommended"
+      ],
+      "rules": {
+        "@angular-eslint/directive-selector": [
+          "error",
+          {
+            "type": "attribute",
+            "prefix": "web-shared",
+            "style": "camelCase"
+          }
```
hallieswan, line 19: Should this be camelCase? 


```suggestion
          {
            "type": "attribute",
            "prefix": "webShared",
            "style": "camelCase"
          }
```

hallieswan: It looks like this fix should also be made in the existing `web-shared-gtm` lib

libs/web-shared/angular/logger/.eslintrc.json
```diff
@@ -0,0 +1,38 @@
...
+        "plugin:jest/recommended"
+      ],
+      "rules": {
+        "@angular-eslint/directive-selector": [
+          "error",
+          {
+            "type": "attribute",
+            "prefix": "web-shared",
+            "style": "camelCase"
+          }
```
hallieswan, line 19:
```suggestion
          {
            "type": "attribute",
            "prefix": "webShared",
            "style": "camelCase"
          }
```

libs/web-shared/angular/storage/src/lib/local-storage.service.ts
```diff
@@ -0,0 +1,44 @@
...
+  private readonly logger = inject(LOGGER);
+
+  getItem(key: string): string | null {
+    if (!this.isBrowser) {
+      return null;
+    }
+    try {
+      return localStorage.getItem(key);
+    } catch (e) {
+      this.logger.error('Failed to get item from localStorage', e);
```
hallieswan: Should we use `logger.warn` for all of these catch blocks, since there are known cases where these throw and it's not really an error? e.g. user disabled `localStorage`, using legacy safari version that throws on `setItem` in private browsing mode, local storage quota is exceeded, etc...

libs/explorers/services/src/lib/app-storage.constants.ts
```diff
@@ -0,0 +1,4 @@
+export const HIDE_VISUALIZATION_OVERVIEW_KEY = 'hide_visualization_overview';
+export const PAGE_SIZE_KEY = 'comparison_tool_page_size';
+export const VALID_PAGE_SIZES: number[] = [10, 25, 50];
+export const DEFAULT_PAGE_SIZE = VALID_PAGE_SIZES[0];
```
hallieswan, line 4: Should these be `readonly`, particularly `VALID_PAGE_SIZES`?

libs/explorers/comparison-tool/src/lib/help-links/visualization-overview-panel/visualization-overview-panel.component.spec.ts
```diff
@@ -16,9 +16,11 @@ import { MessageService } from 'primeng/api';
 import { VisualizationOverviewPanelComponent } from './visualization-overview-panel.component';
 
 describe('VisualizationOverviewPanelComponent', () => {
-  const createMockCookieService = (isHidden = false) => ({
+  const createMockStorageService = (isHidden = false) => ({
     isVisualizationOverviewHidden: jest.fn().mockReturnValue(isHidden),
     setVisualizationOverviewHidden: jest.fn(),
+    getPageSize: jest.fn().mockReturnValue(10),
+    setPageSize: jest.fn(),
   });
```
hallieswan, line 25: `ComparisonToolService` reads `appStorageService.pageSizeOptions` as a property at construction time, so the mock must include it or the field resolves to undefined at runtime -- 

```suggestion
  const createMockStorageService = (isHidden = false) => ({
    isVisualizationOverviewHidden: jest.fn().mockReturnValue(isHidden),
    setVisualizationOverviewHidden: jest.fn(),
    getPageSize: jest.fn().mockReturnValue(10),
    setPageSize: jest.fn(),
    pageSizeOptions: [10, 25, 50],
});
```

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -595,10 +601,14 @@ export class ComparisonToolService<T> {
...
     }));
+
+    if (query.pageSize !== undefined) {
+      this.appStorageService.setPageSize(query.pageSize);
+    }
   }
 
   handleLazyLoad(event: TableLazyLoadEvent) {
-    const defaultRowsPerPage = this.viewConfigSignal().rowsPerPage;
+    const defaultRowsPerPage = this.pageSize();
```
hallieswan, line 611: I missed this last time -- do we still use `viewConfig.rowsPerPage` anywhere? It was previously used as the default page size per CT, but I'm not sure it's actually read now. Should we remove it, or update the code to use it as the initial page size when there's no localStorage value? The latter sounds complicated, and since all CTs use 10 anyway, I'd lean toward removing it until we have a real use case for per-CT defaults. What do you think?

hallieswan (commented):
Looking great - the page number preference works well across all CTs now! Just a couple small questions/suggestions

libs/explorers/services/src/lib/comparison-tool.service.ts
sagely1: That's a good idea.  This is dead code now.

libs/explorers/comparison-tool/src/lib/help-links/visualization-overview-panel/visualization-overview-panel.component.spec.ts
sagely1: No longer reading from the service going forward.

libs/web-shared/angular/logger/.eslintrc.json
sagely1: Good catch.  I'll update everything in libs/web-shared.  I've also found some pre-existing issues in other libraries.

libs/web-shared/angular/storage/.eslintrc.json
sagely1: Fixed in next commit

libs/web-shared/angular/storage/src/lib/local-storage.service.ts
hallieswan, line 42: Should these be `logger.warn` instead of `logger.error` too?

hallieswan (approved):
Looks great! Just one small question

libs/web-shared/angular/storage/src/lib/local-storage.service.ts
sagely1: If removeItem fails, we could end up with a key causing unintended behavior so it should be error in this case.

PR #3978, opened 2026-04-10
chore(devcontainer): install Claude Code natively and add devcontainer rule (SMR-738)

hallieswan:
## Description

Publish devcontainer with Claude Code installed natively. Add Claude rule to support devcontainer development.

> [!NOTE] 
> A follow up PR will activate the devcontainer published by this PR and remove `@anthropic-ai/claude-code` from `devDependencies` in `package.json`

## Related Issue

- [SMR-738](https://sagebionetworks.jira.com/browse/SMR-738)
- [SMR-695](https://sagebionetworks.jira.com/browse/SMR-695)

## Changelog

- Install Claude Code natively in devcontainer image
- Restore default provenance attestations in devcontainer workflow to prevent SBOM attestation error, see this [comment](https://sagebionetworks.jira.com/browse/SMR-695?focusedCommentId=316104) for more context
- Add Claude rule to support devcontainer development

[SMR-738]: https://sagebionetworks.jira.com/browse/SMR-738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SMR-695]: https://sagebionetworks.jira.com/browse/SMR-695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/.devcontainer/Dockerfile
+++ b/.github/.devcontainer/Dockerfile
@@ -177,7 +177,7 @@ RUN pipx ensurepath \
   && pipx install uv=="${uvVersion}"
 
 # Install Claude Code.
-RUN curl -fsSL https://claude.ai/install.sh | bash
+RUN curl -fsSL https://claude.ai/install.sh | bash -s stable
 
 # Prepare the user's environment.
 RUN usermod --shell /bin/bash $user \
--- a/.github/.devcontainer/Dockerfile
+++ b/.github/.devcontainer/Dockerfile
@@ -176,6 +176,9 @@ RUN corepack install --global pnpm@"${pnpmVersion}"
 RUN pipx ensurepath \
   && pipx install uv=="${uvVersion}"
 
+# Install Claude Code.
+RUN curl -fsSL https://claude.ai/install.sh | bash
+
 # Prepare the user's environment.
 RUN usermod --shell /bin/bash $user \
   && printf "%s\n" \
--- a/.claude/rules/devcontainer.md
+++ b/.claude/rules/devcontainer.md
@@ -1,0 +1,29 @@
+---
+paths:
+  - .github/.devcontainer/**
+  - .devcontainer/**
+---
+
+# Dev Container
+
+Full documentation: [`docs/develop/maintenance/dev-container.md`](docs/develop/maintenance/dev-container.md)
+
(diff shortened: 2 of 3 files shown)


sagely1 (approved):
LGTM!

.github/.devcontainer/Dockerfile
```diff
@@ -176,6 +176,9 @@ RUN corepack install --global pnpm@"${pnpmVersion}"
 RUN pipx ensurepath \
   && pipx install uv=="${uvVersion}"
 
+# Install Claude Code.
+RUN curl -fsSL https://claude.ai/install.sh | bash
```
rrchai: Is it possible to pin a version via installer?

rrchai (approved):
LGTM - just one question

.github/.devcontainer/Dockerfile
hallieswan: We can install [a specific version](https://code.claude.com/docs/en/setup#install-a-specific-version), but I'm wondering whether it would be worth [allowing stable auto-updates](https://code.claude.com/docs/en/setup#configure-release-channel)? Then we get critical security patches without having to update manually and ensure that the harness matches the model's API / capabilities. For example, the version of Claude Code we had previously pinned had issues using the plugins defined in `settings.json` -- at first, I thought our configuration was wrong, but it just turned out that we needed to upgrade the version. What do you think?

rrchai: hmm. That's true. Using latest sounds good to me.

hallieswan: > Using latest sounds good to me.

We're using `latest` right now -- should we use `stable` instead? I think critical security patches are pushed to the `stable` channel as needed, so we wouldn't miss out on those updates: 

> "latest", the default: receive new features as soon as they’re released
> "stable": use a version that is typically about one week old, skipping releases with major regressions

rrchai (approved):

PR #3985, opened 2026-04-14
feat(explorers): apply auto width sizing to the comparison tool table (MG-828)

sagely1:
## Description

This update replaces the uniform column width calculation in the explorers comparison tool table with content-based auto-sizing. Each column is now individually measured using DOM-based width detection across all three tables (header, pinned, unpinned), ensuring columns are sized to their widest cell content (including the header) unless the content is smaller than a pre-defined min / max value. Headers account for the sort icon and multi-sort badge width so toggling sort columns does not cause layout shift.

Note: There is a possibility these changes may not be desired during testing and design review in which case we can rollback these changes.

## Related Issue

[MG-828](https://sagebionetworks.jira.com/browse/MG-828)

## Changelog

- Replace uniform column width with per-column content-based width calculation using in-place DOM measurement
- Account for PrimeNG sort icon and multi-sort badge in header width by injecting dummy badge elements during measurement
- Enforce minimum (80px) and maximum (300px) column width
- Allow horizontal scrolling when total column widths exceed the container
- Wait for font loading before measuring to prevent icon font race conditions
- Remove unused HelperService injection and dead CSS

## Preview
https://www.loom.com/share/ba85a62ab1b943569348cd15a7b10355

https://www.loom.com/share/02a9a173284847c1b33ad228a4cf80ba

[MG-828]: https://sagebionetworks.jira.com/browse/MG-828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
```diff
@@ -58,45 +57,39 @@ export class ComparisonToolTableComponent implements AfterViewInit {
...
-        getComputedStyle(document.documentElement).getPropertyValue(
-          '--comparison-tool-primary-column-width',
-        ),
-      );
-
-      // Recalculate column widths whenever selected columns change
+      // Recalculate column widths whenever selected columns, pinned data, or unpinned data change
       effect((onCleanup) => {
-        // Access selectedColumns to track changes
+        // Access the signals to track changes
```
sagely1: Note that any change to these signals can result in new rows/columns requiring a recalculation of column widths.

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
```diff
@@ -58,45 +57,39 @@ export class ComparisonToolTableComponent implements AfterViewInit {
...
     if (this.platformService.isBrowser) {
-      const tableElementWidth = this.tableElement()?.nativeElement?.offsetWidth || 0;
-      this.columnWidth = this.calculateNonprimaryColumnWidth(
-        this.selectedColumns().length - 1,
-        tableElementWidth,
-      );
+      // icon fonts may not be loaded yet on initial render,
+      // so wait for fonts to be ready before calculating column widths
+      // to prevent incorrect measurements
+      document.fonts.ready.then(() => {
```
sagely1, line 96: Without this, widths were being calculated inconsistently

libs/explorers/comparison-tool/src/lib/comparison-tool-table/base-table/base-table.component.ts
```diff
@@ -34,7 +34,7 @@ export class BaseTableComponent {
 
   data = input.required<Record<string, any>[]>();
   shouldShowNoDataMessage = input<boolean>(true);
-  columnWidth = input<string>('auto');
+  columnWidth = input<Record<string, string>>({});
```
hallieswan: I think updating to use a plural variable name in `comparison-tool-table` is helpful! What do you think we could standardize on the plural variable for the `base-table` and `comparison-tool-columns` too? 


```suggestion
  columnWidths = input<Record<string, string>>({});
```

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
```diff
@@ -131,8 +124,119 @@ export class ComparisonToolTableComponent implements AfterViewInit {
     this.comparisonToolService.resetPinnedItems();
   }
 
-  calculateNonprimaryColumnWidth(nCols: number, tableWidth: number) {
-    const count = Math.max(nCols, 5);
-    return Math.ceil((tableWidth - this.primaryColumnWidth) / count) + 'px';
+  calculateColumnWidths(): Record<string, string> {
```
hallieswan: What do you think about keeping `nonPrimary` in the variable name? It doesn't look like primary column widths are set here


```suggestion
  calculateNonPrimaryColumnWidths(): Record<string, string> {
```

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
```diff
@@ -131,8 +124,119 @@ export class ComparisonToolTableComponent implements AfterViewInit {
...
+    const savedHeaderWidths = new Map<HTMLElement, string>();
+    const saved: {
+      el: HTMLElement;
+      style: string;
+      descendants: { el: HTMLElement; ws: string }[];
+    }[] = [];
+
+    for (const column of nonPrimaryColumns) {
+      const cells = Array.from(
+        container.querySelectorAll(`[data-column-key="${column.data_key}"]`),
```
hallieswan: We reuse `data-column-key` in a couple places -- would it be possible to move it into a constant?

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
hallieswan: This method is also pretty complicated. Would it be possible to refactor into smaller, testable functions? And add tests for these functions? If they're difficult to test due to JSDOM constraints, could we add e2e tests?

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
hallieswan: Can we specify which signals are necessary in this comment?

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.constants.ts
```diff
@@ -0,0 +1,2 @@
+export const MIN_COLUMN_WIDTH_PX = 80;
+export const MAX_COLUMN_WIDTH_PX = 300;
```
hallieswan, line 9: Does this intentionally match the primary column width or is it incidental? Might be worth a comment if it's intentional

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
hallieswan: What do you think about escaping the attribute selector?


```suggestion
        container.querySelectorAll(`[data-column-key="${CSS.escape(column.data_key)}"]`),
```

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
```diff
@@ -131,8 +124,119 @@ export class ComparisonToolTableComponent implements AfterViewInit {
...
+            const dummy = document.createElement('p-badge');
+            dummy.className =
+              'p-sortable-column-badge p-badge p-badge-circle p-badge-sm p-component';
+            dummy.textContent = '0';
+            sortIcon.appendChild(dummy);
+            injectedBadges.push({ el: dummy, parent: sortIcon as HTMLElement });
+          }
+        }
+      }
+    }
```
hallieswan: Instead of injecting the badge ourselves, would it be possible to adding a hidden real badge directly in the `comparison-tool-columns` template? Then the measurement would naturally include the badge without such tight coupling to PrimeNG classes (which could change in a release and silently break the column sizing). The hidden badge could be conditionally displayed if the column is not sorted

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
```diff
@@ -131,8 +124,119 @@ export class ComparisonToolTableComponent implements AfterViewInit {
...
+      cellsByColumn.set(column.data_key, cells);
+
+      for (const el of cells) {
+        const descEntries: { el: HTMLElement; ws: string }[] = [];
+        el.querySelectorAll('*').forEach((d) => {
+          const desc = d as HTMLElement;
+          descEntries.push({ el: desc, ws: desc.style.whiteSpace });
+          desc.style.whiteSpace = 'nowrap';
+          // Remove width:100% on .column-header so it sizes to content during measurement
+          if (desc.classList.contains('column-header')) {
```
hallieswan: Similarly could this be a constant? Or at least moved to the top of the file, so the classname dependency is clearer?

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
```diff
@@ -58,45 +57,39 @@ export class ComparisonToolTableComponent implements AfterViewInit {
   pinnedData = this.comparisonToolService.pinnedData;
   unpinnedData = this.comparisonToolService.unpinnedData;
 
-  columnWidth = 'auto';
-  primaryColumnWidth = 300;
+  columnWidths: Record<string, string> = {};
```
hallieswan: What do you think about making this a signal? Assigning inside `document.fonts.ready.then()` works with Zone.js but would silently break in zoneless Angular, if we migrate to zoneless Angular in the future

hallieswan (changes requested):
Looking good! A couple of suggestions. 

I also wanted to double check that a couple of functional changes are intended? 
- columns no longer resize when window resized
- all initial columns are no longer guaranteed to fit on screen (down to a min width) -- e.g. on my laptop, all [target nominations columns](https://agora-dev.adknowledgeportal.org/comparison/targets) used to fit on my screen, but now the pharos class column is offscreen

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
sagely1: Signals being changed are spelled out in a comment on line 64.  Are those the signals you were referring to?

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.constants.ts
sagely1: It was just an arbitrary size since auto-sizing the columns means comma-delimited cells would be extremely wide.  It's possible we can tweak this in the future if needed in the future if we keep this auto-sizing feature.

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
sagely1: I had the same thought but unfortunately PrimeNG does not render the badge element at all (does not exist in DOM) when the multi-sort is inactive.

hallieswan: Oh, I was thinking that we would conditionally add a hidden [badge](https://primeng.org/badge) in the template when the column is not sorted ourselves, since PrimeNG doesn't insert the badge for us. What do you think?

hallieswan: Based on discussion, we will use a pixel constant here instead to avoid adding dummy elements to our code and avoids silent failures on PrimeNG updates.

sagely1:
> Looking good! A couple of suggestions.
> 
> I also wanted to double check that a couple of functional changes are intended?
> 
> * columns no longer resize when window resized

Correct, columns are now just calculated on their auto-sized width and thus the window size is not a factor.

> * all initial columns are no longer guaranteed to fit on screen (down to a min width) -- e.g. on my laptop, all [target nominations columns](https://agora-dev.adknowledgeportal.org/comparison/targets) used to fit on my screen, but now the pharos class column is offscreen

Because we have variable numbers of columns depending on the CT, I don't believe we've guaranteed a fit for any resolution.  In my case, I have a smaller screen and that page (in dev) does not fit for me given the number of columns.

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
hallieswan: Oh, yes! I see that, but I think it would be a bit clearer if this comment explicitly included the tracked signals: 


```suggestion
        // Access selectedColumns, pinnedData, and unpinnedData to track changes
```

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
hallieswan: Can we keep the `nonPrimary` in the function name? It doesn't look like primary column widths are set here.

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.constants.ts
```diff
@@ -0,0 +1,32 @@
...
+// narrow visualizations such as heat maps.
+export const MIN_COLUMN_WIDTH_PX = 80;
+export const MAX_COLUMN_WIDTH_PX = 300;
+
+// CSS class applied to column header elements in the comparison tool table.
+// Used during width measurement to temporarily override styles.
+export const COLUMN_HEADER_CLASS = 'column-header';
+
+// CSS class applied to the text container inside column headers.
+export const COLUMN_HEADER_TEXT_CLASS = 'column-header-text';
```
hallieswan: Can we use these classes in the templates as well? e.g. `libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-columns/comparison-tool-columns.component.html`

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.helpers.ts
```diff
@@ -0,0 +1,139 @@
...
+};
+
+export function getCellsByColumn(
+  container: HTMLElement,
+  columns: ComparisonToolColumn[],
+): Map<string, HTMLElement[]> {
+  const cellsByColumn = new Map<string, HTMLElement[]>();
+  for (const column of columns) {
+    const cells = Array.from(
+      container.querySelectorAll(`[data-column-key="${CSS.escape(column.data_key)}"]`),
```
hallieswan, line 28: Is it possible to store this value in a constant as well? And use the constant in template files? e.g. `libs/explorers/comparison-tool/src/lib/comparison-tool-table/base-table/base-table.component.html` and `libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-columns/comparison-tool-columns.component.html`

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.constants.ts
```diff
@@ -0,0 +1,32 @@
...
+
+// Left margin in pixels applied to column header content.
+export const COLUMN_HEADER_MARGIN_LEFT_PX = 10;
+
+// Right margin in pixels applied to column header content.
+export const COLUMN_HEADER_MARGIN_RIGHT_PX = 22;
+
+// Border width in pixels on each side of a column header cell.
+export const COLUMN_HEADER_BORDER_WIDTH_PX = 1;
```
hallieswan: Should we create similar constants in [the styles files](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-columns/comparison-tool-columns.component.scss#L26-L36) and specify that the values should be kept in sync between these files?

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.helpers.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  for (const column of columns) {
+    const cells = cellsByColumn.get(column.data_key) || [];
+    for (const el of cells) {
+      const descEntries: { el: HTMLElement; ws: string }[] = [];
+      el.querySelectorAll('*').forEach((d) => {
+        const desc = d as HTMLElement;
+        descEntries.push({ el: desc, ws: desc.style.whiteSpace });
+        desc.style.whiteSpace = 'nowrap';
+        // Remove width:100% on .column-header so it sizes to content during measurement
```
hallieswan:
```suggestion
        // Remove width:100% on COLUMN_HEADER_CLASS so it sizes to content during measurement
```

hallieswan (commented):
Thanks for splitting the sizing function into separate functions -- it's way clearer! Just a couple small suggestions

rrchai (approved):
tested UI - looks good.

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.helpers.ts
sagely1: [attr.data-column-key]="..." requires the attribute name to be a literal in the template so readability will suffer.

hallieswan (approved):
Looks good to me!

PR #3988, opened 2026-04-15
refactor(explorers): restructure logger lib to GTM-style token+providers pattern

rrchai:
## Description

Currently, `LocalStorageService` cannot be used in apps (like bixarena) using esbuild  `isolatedModules: true` with a build error, because the logger lib's `index.ts` re-exported Logger as `export { Logger }`, and an isolated compiler cannot tell whether Logger is a type or a runtime value, causing a hard build failure.

This PR refactors the logger lib to mirror the GTM lib's pattern - splitting the single `logger.ts` into `logger.token.ts` and `logger.providers.ts`. Declaring Logger as export type directly in the source file makes it unambiguous to any compiler. It also refactors to use new `provideLogger()` in agora and model-ad, so all apps wire the logger consistently.

## Related Issue

N/A

## Changelog

- Restructure logger lib into `logger.token.ts` and `logger.providers.ts` following the GTM lib pattern
- Replace `export { Logger, LOGGER }` with `export *` to fix isolatedModules: true build error
- Add `provideLogger(implementation?)` function for consistent app-level logger wiring
- Update agora and model-ad to use `provideLogger(LoggerService)` instead of raw provider objects

## Preview

1. Start the stack
```
nx serve-detach model-ad-apex
```
2. Force setItem to throw an error "quota exceeded"
```
localStorage.setItem = () => { throw new Error('quota exceeded') }
```

3. Trigger a localStorage write by changing a page size. In the browser console you should see:

Screenshot 2026-04-15 at 2 38 45 PM

### DIFF
--- a/libs/web-shared/angular/logger/src/lib/logger.providers.ts
+++ b/libs/web-shared/angular/logger/src/lib/logger.providers.ts
@@ -1,16 +1,6 @@
 import { Provider, Type } from '@angular/core';
 import { Logger, LOGGER } from './logger.token';
 
-export function provideLogger(implementation?: Type<Logger>): Provider {
-  if (implementation) {
-    return { provide: LOGGER, useExisting: implementation };
-  }
-  return {
-    provide: LOGGER,
-    useValue: {
-      log: (msg: string, data?: Record<string, unknown>) => console.log(msg, data),
-      warn: (msg: string, data?: Record<string, unknown>) => console.warn(msg, data),
-      error: (msg: string, err?: unknown) => console.error(msg, err),
-    },
-  };
+export function provideLogger(implementation: Type<Logger>): Provider {
+  return { provide: LOGGER, useExisting: implementation };
 }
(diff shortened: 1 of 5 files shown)


libs/web-shared/angular/logger/src/lib/logger.providers.ts
```diff
@@ -0,0 +1,16 @@
...
+    return { provide: LOGGER, useExisting: implementation };
+  }
+  return {
+    provide: LOGGER,
+    useValue: {
+      log: (msg: string, data?: Record<string, unknown>) => console.log(msg, data),
+      warn: (msg: string, data?: Record<string, unknown>) => console.warn(msg, data),
+      error: (msg: string, err?: unknown) => console.error(msg, err),
+    },
+  };
```
sagely1: This seems redundant.  Can we just keep line 6 and make implementation required on line 4?  I think then we can DRY up the implementation.

Like this:
`
export function provideLogger(implementation: Type<Logger>): Provider {
  return { provide: LOGGER, useExisting: implementation };
}
`

sagely1 (changes requested):
Looks great, I just have one change to suggest.

libs/web-shared/angular/logger/src/lib/logger.providers.ts
rrchai: good catch! fix in next commit

sagely1 (approved):
LGTM!

hallieswan (approved):
Looks good to me!

PR #3992, opened 2026-04-16
ci(sage-monorepo): audit and harden GitHub Actions workflows (SMR-747)

hallieswan:
## Description

Improvements from Claude Opus 4.7 audit of GitHub Actions workflows. Pin remaining third-party actions by SHA instead of version tags. 

## Related Issue

- [SMR-747](https://sagebionetworks.jira.com/browse/SMR-747)

## Changelog

- Gate `check-explorer-affected` behind the same `{explorer}-pr` environment approval already used by `run-explorer-e2e-tests`, so fork PRs cannot execute PR-controlled code before a maintainer approves. *NOTE: this will require an approver to click through approving the job on every external PR, even when the explorer turns out not to be affected.*
- Drop `secrets: inherit` from the e2e-explorers caller since `SYNAPSE_AUTH_TOKEN` is an environment secret that resolves via the job's `environment:` binding
- Move `NX_CLOUD_ENCRYPTION_KEY` from workflow-level env to the `push` job so the `pr` job (which runs fork-controlled code) cannot access it
- Add explicit least-privilege `permissions:` blocks to all workflows and scope `contents: write` to only the jobs that deploy or create PRs
- Replace direct `${{ secrets.GITHUB_TOKEN }}` interpolation in shell commands with step-level env vars and `--remote-env` to keep the token out of inner shell argv
- Scope `continue-on-error` in `scan-images` to the Trivy scan steps only so infrastructure failures (login, SARIF upload) are no longer silently swallowed
- Pin all third-party actions to SHA, bump stale `setup-python` (v4 -> v6) and `cache` (v3 -> v5), and add `.github/dependabot.yml` for weekly grouped action-version monitoring

[SMR-747]: https://sagebionetworks.jira.com/browse/SMR-747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,10 +80,16 @@ jobs:
         run: echo "SHORT_SHA=${GITHUB_SHA::7}" >> "$GITHUB_ENV"
 
       - name: Build and publish the images of the affected projects
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_ACTOR: ${{ github.actor }}
         run: |
-          devcontainer exec --workspace-folder ../sage-monorepo bash -c ". ./dev-env.sh \
-            && echo ${{ secrets.GITHUB_TOKEN }} | docker login ghcr.io -u ${{ github.actor }} --password-stdin \
-            && nx affected --target=build-image --configuration=edge"
+          devcontainer exec --workspace-folder ../sage-monorepo \
+            --remote-env GH_TOKEN="$GH_TOKEN" \
+            --remote-env GH_ACTOR="$GH_ACTOR" \
+            bash -c '. ./dev-env.sh \
+            && echo "$GH_TOKEN" | docker login ghcr.io -u "$GH_ACTOR" --password-stdin \
+            && nx affected --target=build-image --configuration=edge'
 
       - name: Upload Sentry sourcemaps for the affected projects
         run: |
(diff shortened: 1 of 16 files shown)


.github/workflows/ci.yml
```diff
@@ -10,22 +10,30 @@ on:
...
+      contents: read
+      packages: write # push images to ghcr.io
     env:
+      # Scoped to this job (not workflow-level) so the secret is not exposed to the `pr`
+      # job, which executes PR-controlled code from fork checkouts.
+      NX_CLOUD_ENCRYPTION_KEY: ${{ secrets.NX_CLOUD_ENCRYPTION_KEY }}
       NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN_READ_WRITE }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
```
sagely1, line 36: Since we are pinning the first party SHAs, can you confirm we have SHA pinning set as a policy on the repo?

hallieswan: Good idea! We can change this setting after this PR has been merged

sagely1 (approved):
LGTM!  Let's add the security policy if things look good after the merge.

PR #4000, opened 2026-04-21
feat(bixarena): enable battle example cards with DNA helix in angular app (SMR-749)

rrchai:
## Description

Reworks the battle landing's example prompt cards, wrapping them in a DNA helix backdrop that ties the picker back to the BioArena brand. Clicking a card still submits the prompt straight into the composer, and each card surfaces its category as a small chip when one is set. The cards are extracted into a reusable `bixarena-prompt-card` component , so the upcoming home-page trending cards can share the same building block.

The chip is read-only - a deliberate step back from the earlier prototype that exposed every category as a clickable filter node. Showing all categories on the landing felt like a poor fit for non-expert users who don't think in taxonomies, so the picker stays focused on the prompt itself rather than doubling as a category navigator.

## Related Issue

[SMR-749](https://sagebionetworks.jira.com/browse/SMR-749)

## Changelog

- Replace the static 5-category helix-node picker with an API-driven 3-card grid rendered by a new reusable `bixarena-prompt-card` component
- Turn the DNA helix into a decorative animated backdrop with a spin on refresh and fade-out/fade-in on card swap, honoring `prefers-reduced-motion`
- Extract helix geometry (`buildStrandPath`, `buildRungs`) into a standalone `dna-helix.ts` module
- Tighten landing spacing and pin the composer and disclaimer to the viewport bottom with a responsive breakpoint scale (single column below `$lg`, icon-only CTA between `$lg` and `$xl`)
- Cap the prompt-composer textarea at 8rem with internal scroll so long prompts cannot push the disclaimer off-screen
- Expand test coverage for initial fetch, click emission, refresh, empty and error states, helix geometry, and category label formatting

## Preview

https://github.com/user-attachments/assets/aa88e37f-748a-4169-9ad4-e5fc632f6faf

[SMR-749]: https://sagebionetworks.jira.com/browse/SMR-749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/bixarena/battle/src/lib/battle.component.html
+++ b/libs/bixarena/battle/src/lib/battle.component.html
@@ -9,15 +9,17 @@
         </p>
       </div>
       <bixarena-example-prompts (promptSelect)="onPromptSubmit($event)" />
-      <bixarena-prompt-composer
-        placeholder="Ask anything biomedical..."
-        [maxLength]="state.promptLengthLimit"
-        [disabled]="state.phase() === 'creating'"
-        (promptSubmit)="onPromptSubmit($event)"
-      />
-      <p class="disclaimer">
-        AI may make mistakes. Do not include sensitive or personal information.
-      </p>
+      <div class="composer-group">
+        <bixarena-prompt-composer
+          placeholder="Ask anything biomedical..."
+          [maxLength]="state.promptLengthLimit"
+          [disabled]="state.phase() === 'creating'"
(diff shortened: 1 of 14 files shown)


tschaffter:
PR #4000! 🎉 Love seeing the monorepo still going strong. Congrats team, keep it up 🚀

PR #4001, opened 2026-04-21
fix(sage-monorepo): explicitly pass SYNAPSE_AUTH_TOKEN to e2e-explorer workflow (SMR-752)

hallieswan:
## Description

Fixes a regression from #3992 where removing `secrets: inherit` from the e2e-explorers caller workflow caused `SYNAPSE_AUTH_TOKEN` to be unavailable in the reusable e2e-explorer workflow. The Synapse client fell back to anonymous access, resulting in a 403 on data download. Reusable workflows require the caller to pass secrets explicitly even when the secret is configured on the callee job's environment.

## Related Issue

- [SMR-752](https://sagebionetworks.jira.com/browse/SMR-752)

## Changelog

- Declare `SYNAPSE_AUTH_TOKEN` in the callee's `workflow_call.secrets` block so it can be received from the caller
- Add an explicit `secrets:` mapping in the caller to pass only `SYNAPSE_AUTH_TOKEN` (preserving the least-privilege improvement from #3992)
- Update inline comment to reflect that the secret must be passed explicitly rather than relying on environment-only resolution

[SMR-752]: https://sagebionetworks.jira.com/browse/SMR-752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/e2e-explorer.yaml
+++ b/.github/workflows/e2e-explorer.yaml
@@ -6,6 +6,10 @@ on:
         required: true
         type: string
         description: The name of the Explorer to run e2e tests for
+    secrets:
+      SYNAPSE_AUTH_TOKEN:
+        required: false
+        description: Synapse PAT for explorer data sources
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-explorer.yaml
```diff
@@ -6,6 +6,10 @@ on:
         required: true
         type: string
         description: The name of the Explorer to run e2e tests for
+    secrets:
+      SYNAPSE_AUTH_TOKEN:
+        required: false
```
hallieswan, line 11: Use `required: false` because fork pushes won't necessarily have the secret in their repo context

rrchai (approved):

PR #4017, opened 2026-04-25
feat(agora): add support for sentry post config change, align sentry release info (AG-2031)

sagely1:
## Description
This PR replaces the initial exploratory implementation of Sentry and aligns with the changes made to the infra repos to support better Sentry integration.

## Related Issue

[AG-2031](https://sagebionetworks.jira.com/browse/AG-2031)

## Changelog

- Add `sentryDsn` and `sentryEnvironment` to the explorers config schema, transformer, application YAMLs, and test mocks
- Move Sentry initialization from `main.ts` into the app initializer in `app.config.ts` using the shared `initSentry()` so it runs after runtime config is loaded
- Replace the hostname-to-environment mapping (`getSentryEnvironment` / `hostEnvironmentMap`) with a config-driven `environment` value and remove the old helper
- Skip Sentry init when `dsn` is empty, and warn + skip when `environment` is empty so misconfigured deployments fail loudly instead of silently reporting to the wrong project
- Prefix the `sentry-cli sourcemaps upload --release` value with the project slug (`agora@$SENTRY_RELEASE`, `model-ad@$SENTRY_RELEASE`) to match Sentry's release identifier convention
- Update `init-sentry` unit tests to cover the new dsn/environment guards and remove the obsolete `getSentryEnvironment` tests

## Testing
- Testing will require a follow-up to see that errors are logged in the Sentry UI.

[AG-2031]: https://sagebionetworks.jira.com/browse/AG-2031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -11,7 +11,6 @@
     "NX_CLOUD_ENCRYPTION_KEY": "${localEnv:NX_CLOUD_ENCRYPTION_KEY}",
     "NX_HEAD": "${localEnv:NX_HEAD}",
     "SENTRY_AUTH_TOKEN": "${localEnv:SENTRY_AUTH_TOKEN}",
-    "SENTRY_RELEASE": "${localEnv:SENTRY_RELEASE}",
     "SONAR_PULL_REQUEST_NUMBER": "${localEnv:SONAR_PULL_REQUEST_NUMBER}",
     "SONAR_TOKEN": "${localEnv:SONAR_TOKEN}",
     "UV_CACHE_DIR": "${containerWorkspaceFolder}/.cache/uv"
--- a/apps/agora/app/project.json
+++ b/apps/agora/app/project.json
@@ -154,7 +154,7 @@
       "options": {
         "commands": [
           "sentry-cli sourcemaps inject dist/apps/agora/app",
-          "sentry-cli sourcemaps upload --org sage-bionetworks --project agora --release agora@$SENTRY_RELEASE dist/apps/agora/app",
+          "sentry-cli sourcemaps upload --org sage-bionetworks --project agora --release agora@$SENTRY_RELEASE_SUFFIX dist/apps/agora/app",
           "find dist/apps/agora/app -name '*.map' -delete"
         ],
         "parallel": false
--- a/apps/agora/app/project.json
+++ b/apps/agora/app/project.json
@@ -154,7 +154,7 @@
       "options": {
         "commands": [
           "sentry-cli sourcemaps inject dist/apps/agora/app",
-          "sentry-cli sourcemaps upload --org sage-bionetworks --project agora --release $SENTRY_RELEASE dist/apps/agora/app",
+          "sentry-cli sourcemaps upload --org sage-bionetworks --project agora --release agora@$SENTRY_RELEASE dist/apps/agora/app",
           "find dist/apps/agora/app -name '*.map' -delete"
         ],
         "parallel": false
--- a/apps/model-ad/app/project.json
+++ b/apps/model-ad/app/project.json
@@ -159,7 +159,7 @@
       "options": {
         "commands": [
           "sentry-cli sourcemaps inject dist/apps/model-ad/app",
-          "sentry-cli sourcemaps upload --org sage-bionetworks --project model-ad --release model-ad@$SENTRY_RELEASE dist/apps/model-ad/app",
+          "sentry-cli sourcemaps upload --org sage-bionetworks --project model-ad --release model-ad@$SENTRY_RELEASE_SUFFIX dist/apps/model-ad/app",
           "find dist/apps/model-ad/app -name '*.map' -delete"
         ],
         "parallel": false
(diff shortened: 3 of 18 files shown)


apps/model-ad/app/src/main.ts
```diff
@@ -1,16 +1,5 @@
 import { bootstrapApplication } from '@angular/platform-browser';
-import { getSentryEnvironment } from '@sagebionetworks/explorers/sentry';
-import * as Sentry from '@sentry/angular';
 import { appConfig } from './app/app.config';
 import { AppComponent } from './app/app.component';
 
-Sentry.init({
-  dsn: 'https://bbf0d51ab53013fe73d95348a6bffe61@o4510881207418880.ingest.us.sentry.io/4510896864559104',
```
sagely1, line 8: hard-coded DSN no longer needed

apps/model-ad/app/src/main.ts
```diff
@@ -1,16 +1,5 @@
...
-  environment: getSentryEnvironment({
-    'dev.modeladexplorer.org': 'dev',
-    'stage.modeladexplorer.org': 'stage',
-    'modeladexplorer.org': 'prod',
-  }),
```
sagely1, line 13: environment now passed in via SENTRY_ENVIRONMENT

apps/agora/app/project.json
```diff
@@ -154,7 +154,7 @@
       "options": {
         "commands": [
           "sentry-cli sourcemaps inject dist/apps/agora/app",
-          "sentry-cli sourcemaps upload --org sage-bionetworks --project agora --release $SENTRY_RELEASE dist/apps/agora/app",
+          "sentry-cli sourcemaps upload --org sage-bionetworks --project agora --release agora@$SENTRY_RELEASE dist/apps/agora/app",
```
hallieswan: `$SENTRY_RELEASE` already includes the app prefix in the infra repo -- should we remove the prefix here to prevent a mismatch? 

```suggestion
          "sentry-cli sourcemaps upload --org sage-bionetworks --project agora --release $SENTRY_RELEASE dist/apps/agora/app",
```

apps/model-ad/app/project.json
```diff
@@ -159,7 +159,7 @@
       "options": {
         "commands": [
           "sentry-cli sourcemaps inject dist/apps/model-ad/app",
-          "sentry-cli sourcemaps upload --org sage-bionetworks --project model-ad --release $SENTRY_RELEASE dist/apps/model-ad/app",
+          "sentry-cli sourcemaps upload --org sage-bionetworks --project model-ad --release model-ad@$SENTRY_RELEASE dist/apps/model-ad/app",
```
hallieswan: Same question here: 

```suggestion
          "sentry-cli sourcemaps upload --org sage-bionetworks --project model-ad --release @$SENTRY_RELEASE dist/apps/model-ad/app",
```

libs/explorers/sentry/src/lib/init-sentry.spec.ts
```diff
@@ -1,72 +1,62 @@
...
-      expect.objectContaining({
-        dsn: mockConfig.dsn,
-        release: '1.0.0+abc1234',
-        sendDefaultPii: false,
-      }),
-    );
+    expect(Sentry.init).toHaveBeenCalledWith({
+      dsn: mockConfig.dsn,
+      environment: 'dev',
+      release: '1.0.0+abc1234',
```
hallieswan: Should the test match the expected release format? e.g. something like this:


```suggestion
      release: 'explorer@1.0.0+abc1234',
```

hallieswan (changes requested):
Looks great! Just a couple small questions

apps/agora/app/project.json
sagely1: I believe this might be needed because this is prepending to the release which is version + sha: see `ci.yml`

sagely1: You are right that this is confusing so I will rename this variable to SENTRY_RELEASE_SUFFIX.

apps/model-ad/app/project.json
sagely1: renamed as per above

.devcontainer/devcontainer.json
```diff
@@ -11,7 +11,6 @@
     "NX_CLOUD_ENCRYPTION_KEY": "${localEnv:NX_CLOUD_ENCRYPTION_KEY}",
     "NX_HEAD": "${localEnv:NX_HEAD}",
     "SENTRY_AUTH_TOKEN": "${localEnv:SENTRY_AUTH_TOKEN}",
-    "SENTRY_RELEASE": "${localEnv:SENTRY_RELEASE}",
```
hallieswan, line 14: Do we need a similar `SENTRY_RELEASE_SUFFIX` variable here or is it unnecessary?

hallieswan (approved):
Looks good! Just one question for you

.devcontainer/devcontainer.json
sagely1: Unnecessary since we would need a real token to upload sourcemaps anyway and I don't imagine we need to do that manually.

PR #4042, opened 2026-05-01
feat(qtl): add chiclet, chiclet card component, and port existing chiclets into reusable components  (QTL-18)

sagely1:
## Description
This PR builds a new foundational component (chiclet component) based on a PrimeNG chip component that needs to support dismissable and non-dismissable variants.  The chiclet needs to support different styles - font sizes, html styles, colors, dismissable state.  The chiclet will replace the custom chiclet in the comparison tools (used when adding/removing filters).  This chiclet component also has a non-dismissable variant that will be used in a new ChicletCardComponent (also in this PR) that displays on the QTL Explorer home page (non-interactive).

## Related Issue
[QTL-18](https://sagebionetworks.jira.com/browse/QTL-18)

## Changelog
- Add ChicletComponent that encapsulates desired chiclet behavior and importantly supports variants
- Add ChicletCardComponent to libs/explorers/ui with a required title and chiclets input
- Removed existing chiclets from Agora GCT and refactored to use the shared chiclet components
- Add a shared Chiclet model (label, color) to libs/explorers/models
- Export the new component and model from their library barrel files
- Add Storybook stories covering default and minimal-chiclet variants
- Add unit tests verifying title rendering, chiclet rendering, and inline color styling
- Ensured chiclet labels are not selectable

## Preview
QTL Home Page ChicletCardComponent storybook preview
image

GCT before and after:
Not that the close button is now aligned with text and the text is (intentionally) no longer selectable
| Before | After |
| -------- | -------- |
| image | image |

[QTL-18]: https://sagebionetworks.jira.com/browse/QTL-18?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/explorers/ui/src/lib/components/chiclet/chiclet.component.html
```diff
@@ -0,0 +1,24 @@
+<p-chip
```
hallieswan: The old components had a chiclet-specific aria-label on each clear button (`Clear ${title !== '' ? title : description}`), but there isn't an equivalent on the new chiclet. Right now the remove icon falls back to PrimeNG's global `removeLabel`, so we've lost the per-chiclet context for screen reader users. 

Could we add a `removeAriaLabel` input to `ChicletComponent` and surface it on the actual remove button? Then the `filter-list-item` callsite can pass `[removeAriaLabel]="'Clear ' + (title || description)"` to preserve the prior behavior.

libs/explorers/ui/src/lib/components/chiclet/chiclet.component.html
```diff
@@ -0,0 +1,24 @@
...
+  [style.--p-chip-background]="backgroundColor()"
+  [style.--p-chip-color]="textColor()"
+  [style.--p-chip-remove-icon-color]="closeIconColor()"
+  [style.--p-chip-remove-icon-size]="closeIconSizePx()"
+  (onRemove)="removed.emit()"
+>
+  <span class="chiclet-label unselectable-text" [style.font-size]="fontSize()">
+    @if (label()) {
+      <span [innerHTML]="label()"></span>
+    }
```
hallieswan: Is `innerHTML` needed here? I don't see a consumer (other than the Storybook) that uses this behavior, so what do you think about either just calling `{{ label() }}` directly or removing the `label` input entirely and just relying on projection to simplify this component?

libs/explorers/ui/src/lib/components/chiclet-card/chiclet-card.component.html
```diff
@@ -0,0 +1,12 @@
+<div class="chiclet-card">
+  <h2>{{ title() }}</h2>
+  <div class="chiclets">
+    @for (chiclet of chiclets(); track chiclet.label) {
```
hallieswan: `label` isn't guaranteed to be unique (see the Default story where "chr2:164.3Mb" is repeated with different colors) -- can we track a unique value (like the index) instead? That will ensure that Angular doesn't reuse DOM nodes incorrectly


```suggestion
    @for (chiclet of chiclets(); track $index) {
```

libs/explorers/testing/src/index.ts
```diff
@@ -2,6 +2,7 @@
 export * from './lib/constants';
 export * from './lib/mocks';
 export * from './lib/providers';
+export * from './lib/queries';
```
hallieswan: I found this name a bit misleading, since `queries` generally makes me think about data-fetching queries rather than Testing Library `getBy*/queryBy*` "queries". What do you think about renaming this directory to something like `dom-helpers`?

libs/explorers/testing/src/lib/queries/index.ts
```diff
@@ -0,0 +1 @@
+export * from './chiclet-queries';
```
hallieswan: Similarly, could think about renaming this file as well

libs/explorers/comparison-tool/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './lib/comparison-tool.component';
+export { ComparisonToolFilterListItemComponent } from './lib/comparison-tool-filter-list/comparison-tool-filter-list-item/comparison-tool-filter-list-item.component';
```
hallieswan: `ComparisonToolFilterListItemComponent` is mostly a thin wrapper (a visibility toggle plus a chiclet) so promoting it to the `explorers-comparison-tool` public API feels like the wrong layer. Could we move it into `explorers-ui` as a sibling of `ChicletComponent` (e.g. `DismissibleChicletComponent`)? Then the `comparison-tool` barrel stays narrow and the component is reusable outside the explorers CT

libs/explorers/comparison-tool/CLAUDE.md
```diff
@@ -59,9 +60,12 @@ Bootstrap tests with `provideComparisonToolService()` and `provideComparisonTool
 
 ### Public API
 
-Only two exports from `src/index.ts`:
+The barrel re-exports the root component plus any sub-component with a legitimate cross-package consumer; everything else stays internal so it can be refactored freely.
```
hallieswan, line 62: If we move `comparison-tool-filter-list-item` into the `ui` library, then we could also update this doc to specify that generally components that are reused outside the CT should be moved to a generic library.

libs/explorers/testing/src/lib/queries/chiclet-queries.ts
```diff
@@ -0,0 +1,15 @@
+import { screen, within } from '@testing-library/angular';
+
+/**
+ * Finds the remove (close) button inside the `<explorers-chiclet>` whose
+ * label text matches `labelMatcher`. Useful when multiple chiclets are on
+ * screen and `getByRole('button')` would be ambiguous.
+ */
+export function getChicletRemoveButton(labelMatcher: string | RegExp): HTMLElement {
```
hallieswan: If we add an aria-label to the remove button, then we may not need this helper, since the calls will be much simpler

libs/explorers/models/src/lib/chiclet.ts
```diff
@@ -0,0 +1,4 @@
+export interface Chiclet {
```
hallieswan, line 1: Should we add an optional `textColor` field to this model, and default to white in the chiclet-card when omitted? That handles the QTL home page case as-is, but lets a future caller pass a pale color without ending up with unreadable white-on-white.

libs/explorers/ui/src/lib/components/chiclet/chiclet.component.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * When both are set, `label` renders first, then projected content.
+ */
+export class ChicletComponent {
+  label = input<string>();
+  backgroundColor = input<string>();
+  textColor = input<string>();
+  closeIconColor = input<string>();
+  closeIconSize = input<number>(14);
+  fontSize = input<string>();
```
hallieswan, line 23: Would it be worth aligning these inputs so they both use the same type?

hallieswan (changes requested):
This is looking great! Just a couple suggestions

rrchai (commented):
The chips looks great. No new findings beyond what @hallieswan already raised.

libs/explorers/ui/src/lib/components/chiclet/chiclet.component.html
sagely1: Looks like PrimeNG does not support ariaLabel so we will have to do some overrides to support this feature (more trouble than it's worth) so I will remove PrimeNG chips.

libs/explorers/comparison-tool/src/index.ts
sagely1: I think moving this component might be confusing to have in explorers-ui.  It might be clearer if we split this into something that can go into explorers-ui and have *-comparison-tool-filter-list own the non-presentational elements.

libs/explorers/comparison-tool/CLAUDE.md
sagely1: see above

libs/explorers/ui/src/lib/components/chiclet/chiclet.component.ts
sagely1: Probably not since at this point in development.  Changing to number when mapping to a generic CSS property will enforce specific limitations.

libs/explorers/models/src/lib/chiclet.ts
sagely1: No.  Chiclet is a data shape that `ChicletCardComponent` uses.  The design shows various colors against a constant text color.  It may be better at this point to consider the single text color a concern of the `ChicletCardComponent`.  That said, the property `color` could be confusing so I'll rename it to `backgroundColor`

hallieswan (approved):
Looks good to me!

PR #4058, opened 2026-05-06
fix(bixarena): drop divergent BT entries from leaderboard snapshots

rrchai:
## Description

The daily leaderboard snapshot job was failing with `psycopg.errors.NumericValueOutOfRange` whenever Bradley-Terry produced ±inf ratings on perfectly-separable per-category data — common at low data density on stage and on any newly-seeded prod category. This PR drops divergent entries before insert, skips the snapshot row when no qualifying entries remain, and respects a per-snapshot model floor so a leaderboard with too few survivors is gracefully skipped rather than published. Bundled in: a few visual tweaks to the battle-view composer (full-width to match panels, tighter input height, flat disclaimer) that reclaim vertical space for the streaming LLM responses without weakening transparency.

## Changelog

- Drop entries with non-finite or out-of-range BT scores before insert; clamp bootstrap CI bounds to fit `NUMERIC(10,6)`
- Skip snapshot creation when no qualifying entries remain, with explicit reason routing (`no_qualifying_entries`, `insufficient_qualifying_entries`)
- Wire `--min-total-models` through both single-slug and `--all` CLI paths so the entry floor is consistently applied
- Lower default `--composer-input-min-height` and widen the battle-view composer to match the model-panel grid
- Flatten the disclaimer panel on battle view (no background, smaller text) to recover vertical space for LLM responses
- Add unit tests covering BT divergence drops, CI clamping, and skipped-reason routing

### DIFF
--- a/libs/bixarena/leaderboard-python/bixarena_leaderboard/snapshot_generator.py
+++ b/libs/bixarena/leaderboard-python/bixarena_leaderboard/snapshot_generator.py
@@ -11,6 +11,7 @@ Encapsulates the full snapshot workflow:
 
 import json
 import logging
+import math
 from datetime import UTC, datetime
 
 from bixarena_leaderboard.db_helper import (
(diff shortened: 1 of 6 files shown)


libs/bixarena/leaderboard-python/bixarena_leaderboard/snapshot_generator.py
```diff
@@ -43,47 +44,56 @@ def __init__(self, message: str, summary: dict) -> None:
 
 SNAPSHOT_IDENTIFIER_FORMAT = "snapshot_%Y-%m-%d_%H-%M-%S"
 
+# BT can diverge to ±inf on perfectly separable per-category data. Drop entries
+# with extreme btScore; clamp bootstrap CI bounds so they fit NUMERIC(10, 6).
+_BT_SCORE_MIN_EXCLUSIVE = 0.0  # drop if btScore <= this
+_BT_SCORE_MAX = 3000.0  # drop if btScore > this
+_CI_CLAMP = 3000.0  # bootstrap quantiles clamped to [-_CI_CLAMP, _CI_CLAMP]
```
rrchai, line 51: Drop entries if BT_SCORE <= 0 OR > 3000 OR infinite, also set boundaries for CI. 
This will allow us to show more reasonable scores for per-category leaderboard while the data volume is small.

libs/bixarena/leaderboard-python/bixarena_leaderboard/snapshot_generator.py
```diff
@@ -132,12 +142,61 @@ def generate_snapshot(
...
+        for entry in leaderboard_entries:
+            _clamp_ci_bounds(entry)
+
         snapshot_identifier = datetime.now(UTC).strftime(SNAPSHOT_IDENTIFIER_FORMAT)
         description = (
             f"Leaderboard snapshot with {len(all_evaluations)} evaluations "
             f"and {len(leaderboard_entries)} ranked models"
         )
 
+        if len(leaderboard_entries) < min_total_models:
```
rrchai, line 173: the snapshot has entries < min_total_models will be dropped (not inserting snapshot)

PR #4059, opened 2026-05-06
feat(qtl): add header, hero background, and loading icon colors (QTL-26, QTL-69, QTL-87)

hallieswan:
## Description

This PR fills in more of the xQTL Explorer's UI scaffolding by wiring in the shared `<explorers-header>` with QTL branding, applying the QTL hero background image to all wiki/error/not-found routes, and theming the explorers loading icon with QTL colors. 

It also fixes a CSS-cascade foot-gun discovered along the way: shared `_variables.scss` files were re-emitting `:root` blocks into every component that `@use`'d them, which silently undid app-level CSS custom property overrides as soon as a lazy-loaded route mounted. The fix splits the `:root` emissions into a dedicated `_root.scss` so consuming `_variables.scss` from a component is now a no-op cascade-side.

## Related Issue

- [QTL-26](https://sagebionetworks.jira.com/browse/QTL-26)
- [QTL-69](https://sagebionetworks.jira.com/browse/QTL-69)
- [QTL-87](https://sagebionetworks.jira.com/browse/QTL-87)

## Changelog

- Add `<explorers-header>` to the qtl app shell with `qtl-assets/images/header-logo.svg`, a Home nav link from `qtl-util`, and a temporary empty search slot until a qtl-specific search component is created (tracked in [QTL-30](https://sagebionetworks.jira.com/browse/QTL-30))
- Add `qtl-assets/images/hero-background.svg` and `DEFAULT_HERO_BACKGROUND_IMAGE_PATH` to `qtl-config`, then thread it into the About, News, Terms of Service, Not Found, and Error routes
- Add `QTL_LOADING_ICON_COLORS` to `qtl-config` and provide it via the explorers `LOADING_ICON_COLORS` injection token in `AppComponent`
- Expose `--header-padding-x` and `--header-padding-y` custom properties on the shared explorers header (mirroring the existing `--footer-padding-x` pattern) so per-product padding overrides don't require forking the component
- Add a `--header-link-color` override hook on the shared explorers header (with `--color-action-primary` fallback) so apps can theme the header active/hover link color without leaking the override to other `--color-action-primary` consumers inside the header (e.g. the search input's icon and selected-result row)
- Split `:root { --... }` emissions out of `_variables.scss` into a new `_root.scss` in each shared style lib (explorers, agora, model-ad, qtl); the variables file is now Sass-only and the new file is the sole side-effect emitter, forwarded only by the global index

### Header

image

### Hero background is QTL-specific -- About / News / Terms of Service

image

image

image

### Hero background is QTL-specific -- Not Found / Error

image

image

### Loading icon colors are QTL specific

image

### Active link color is QTL-specific -- Home (active) on home page 

image

### Hover link color is QTL-specific -- Home (hover) on About page

https://github.com/user-attachments/assets/778dcf4c-75bb-4b38-a90a-17c689c57da7

### Mobile Active link color is QTL-specific -- TOS (active) on TOS page

image

[QTL-26]: https://sagebionetworks.jira.com/browse/QTL-26?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QTL-69]: https://sagebionetworks.jira.com/browse/QTL-69?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QTL-87]: https://sagebionetworks.jira.com/browse/QTL-87?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QTL-30]: https://sagebionetworks.jira.com/browse/QTL-30?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/qtl/styles/src/lib/_variables.scss
```diff
@@ -1,12 +1,4 @@
-/* stylelint-disable no-duplicate-selectors */
-@use 'sass:map';
-
 $main-colors: (
   secondary: #2c5182,
+  action-primary: #6e76ae,
```
hallieswan: Originally set `action-primary` to match the header active tab color in the header design. However, based on the [style guide](https://www.figma.com/design/TpCkecXfQ0RGXfVlyEKpxf/QTL-Explorer?node-id=2223-9866&p=f&m=dev), the `action-primary` color is actually different, so this override should only apply to the header links. Will update in the next commit

libs/agora/styles/src/lib/_variables.scss
```diff
@@ -40,12 +39,6 @@ $spacings: (
   'xl': 50px,
 );
 
-:root {
-  @each $key, $value in $spacings {
-    --spacing-#{$key}: #{$value};
-  }
-}
-
```
hallieswan, line 48: Split `_variables.scss` -- the original files were defining Sass variables and emitting `:root { --... }` blocks. Every time a component did `@use '<scope>/styles/src/variables' as vars`, Sass re-emitted those `:root` blocks into the component's compiled CSS. Angular then dropped that CSS into a lazy-loaded stylesheet appended to `<head>` whenever the route mounted -- silently overwriting any app-level CSS custom property override of equal specificity, cascade-order-wise. When QTL set `--color-action-primary` globally, the override worked on the home page, but as soon as we navigated to an explorers-based route (About, News, TOS), the explorers `:root` re-emission won the cascade. (We ultimately moved to a header-scoped `--header-link-color` for the QTL link color, but the underlying cascade foot-gun was still worth fixing.) 

The fix keeps Sass variables in `_variables.scss` (now Sass-only, no CSS output) and moves the emissions into a new `_root.scss` that's forwarded only from the global `index.scss`. Applied the fix across explorers, agora, model-ad, and qtl so we don't trip on the same thing later.

apps/qtl/app/src/app/app.routes.ts
```diff
@@ -60,6 +67,7 @@ export const routes: Route[] = [
       description: 'The link you followed may be broken, or the page may have been removed.',
       robots: 'noindex, follow',
       supportEmail: SUPPORT_EMAIL,
+      backgroundImagePath: DEFAULT_HERO_BACKGROUND_IMAGE_PATH,
```
rrchai, line 70: just curious why it's used a diff variable name here, instead of `heroBackgroundImagePath`?

rrchai (approved):
LGTM - just one clarifying question

apps/qtl/app/src/app/app.routes.ts
hallieswan: Good question! The variable name clarifies how the image will be used on the page -- pages with `backgroundImagePath` (not found and error page) use the image as background for all content, whereas pages with `heroBackgroundImagePath` (about, news, tos) use the image as background for the hero only; other content is displayed on a white background.

sagely1 (approved):
LGTM!

PR #4060, opened 2026-05-07
feat(explorers): update signals implementation and refactor SvgIconComponent (QTL-88)

sagely1:
## Description

Adds a new reusable SvgIconTileComponent that renders an SVG icon inside a colored circular or square background tile with configurable padding. The work is driven by upcoming QTL UI needs but lands as shared explorers infrastructure usable by every explorer app. Modernizes the underlying SvgIconComponent to use Angular signal inputs, adds accessibility attributes, and simplifies and standardizes TooltipButtonComponent's use of SvgIconComponent.

## Related Issue

[QTL-88](https://sagebionetworks.jira.com/browse/QTL-88)

## Changelog

- Add new properties to `SvgIconTileComponent` to support background shape and background color
- Modernize SvgIconComponent to use Angular signal inputs
- Add accessibility attributes
- Add a people SVG icon (more will be added in the future)

### Bug fixes

- `TooltipButtonComponent`: `cfg.enableHoverEffects || true` always evaluated
  to `true` because of the OR fallback. Changed to `?? false` so explicit
  `false` is respected. (This is also the new default, so the binding is
  effectively a no-op now, but the bug was real.). 
- Removed `SvgIconConfig` interface eliminating the dual source of truth between the
config type and the component's input declarations.
- `gene-similar.component.html`: `<explorers-svg-icon name="info-circle">`
  was a silent no-op — `SvgIconComponent` has never had a `name` input. Now
  uses `imagePath="explorers-assets/icons/info-circle.svg"` and renders.  

### Cleanup

- Removed 7 now-redundant `[enableHoverEffects]="false"` bindings across
  `gene-comparison-tool`, `home-card`, and `primary-identifier-controls` by setting a new default.

## Testing
From Figma:
Desired icon for QTL homepage:
image

Storybook Preview:
Default:
image

With custom color:
image

With circular background:
image

With square background:
image

With larger background padding:
image

Model-AD TooltipButtonComponent (is identical before and after): 
| Before                   | After                   |
|--------------------------|-------------------------|
| image | image |
| https://dev.modeladexplorer.org/models/3xTg-AD/biomarkers | https://localhost:8000/models/3xTg-AD/biomarkers |

Agora SimilarGenesComponent (is identical before and after): 
| Before                   | After                   |
|--------------------------|-------------------------|
| image | image |
| https://agora-dev.adknowledgeportal.org/genes/ENSG00000178209/similar | http://localhost:8000/genes/ENSG00000178209/similar |

[QTL-88]: https://sagebionetworks.jira.com/browse/QTL-88?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.html
```diff
@@ -330,7 +330,6 @@ <h1 class="gct-heading h2">Gene Comparison Tool</h1>
                         >
                           <explorers-svg-icon
                             imagePath="explorers-assets/icons/external-link.svg"
-                            [enableHoverEffects]="false"
```
sagely1, line 333: false is the new default for enableHoverEffects

libs/agora/genes/src/lib/components/gene-similar/gene-similar.component.html
```diff
@@ -43,7 +43,10 @@ <h1 class="h2">Similar Genes</h1>
         ></agora-gene-table>
       </div>
       <div>
-        <explorers-svg-icon name="info-circle" class="mr-xs text-gray-600"></explorers-svg-icon>
```
sagely1, line 46: bug: `name` property does not exist

libs/explorers/util/src/lib/svg-icon/svg-icon.component.ts
```diff
@@ -12,27 +11,21 @@ import { SvgIconService } from '@sagebionetworks/explorers/services';
   encapsulation: ViewEncapsulation.None,
 })
 export class SvgIconComponent implements OnInit {
-  @Input() imagePath!: string;
-  @Input() altText = '';
-  @Input() width = 14;
-  @Input() height = 14;
-  @Input() color = 'inherit'; // Default to parent color
-  @Input() enableHoverEffects = true;
```
sagely1, line 20: updated to use the newer `input` syntax

libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.html
hallieswan: Did we flip all of the call sites that didn't set `enableHoverEffects` explicitly before? With the changed default, these sites will stop using hover effects if not explicitly set to `true` now, e.g. `gene-comparison-tool.component.html:159` (cog button) used to get hover effects from the old default and now silently won't

libs/explorers/util/src/lib/modal-link/modal-link.component.ts
```diff
@@ -19,7 +19,7 @@ export class ModalLinkComponent {
   wikiParams = input.required<SynapseWikiParams>();
   iconWidth = input(14);
   iconHeight = input(14);
-  enableHoverEffects = input(true);
+  enableHoverEffects = input(false);
```
hallieswan: As above -- looks like the `explorers-modal-link` default was flipped to align with the new `explorers-svg-icon` default. Did we update the modal-link call sites that didn't set it explicitly? e.g. `gene-evidence-metabolomics.component.html:12` ("Data and Methods" link) used to get hover effects from the old default and now silently won't

libs/explorers/util/src/lib/svg-icon-tile/svg-icon-tile.component.ts
```diff
@@ -0,0 +1,23 @@
...
+export type SvgIconTileShape = 'circle' | 'square';
+
+@Component({
+  selector: 'explorers-svg-icon-tile',
+  imports: [SvgIconComponent],
+  templateUrl: './svg-icon-tile.component.html',
+  styleUrls: ['./svg-icon-tile.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class SvgIconTileComponent {
```
hallieswan: Just curious, how did we land on "tile" as the name for this component vs something like "avatar" or "badge"? Just want to make sure the name doesn't collide with a future card-style tile icon component

libs/explorers/util/src/lib/svg-icon-tile/svg-icon-tile.component.html
```diff
@@ -0,0 +1,15 @@
+<div
```
hallieswan: Did we consider using [PrimeNG's avatar component](https://primeng.org/avatar) for this? Just curious, since it already supports `shape="circle|square"` sizing and content projection for `explorers-svg-icon` to slot in. If we couldn't use `p-avatar` here, then might be worth a comment somewhere documenting why?

libs/explorers/util/src/lib/modal-link/modal-link.component.ts
hallieswan: Actually, when I tested locally, the modal link hover effect still works, even though `enableHoverEffects` is false. It looks like the wrapper button has its own hover rule that supersedes the entire `enableHoverEffects` mechanism for the modal-link svg icon. Should we remove the `enableHoverEffects` input here since it's a no-op?

hallieswan (changes requested):
Looking good! Just a few comments

libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.html
sagely1: > Did we flip all of the call sites that didn't set `enableHoverEffects` explicitly before? With the changed default, these sites will stop using hover effects if not explicitly set to `true` now, e.g. `gene-comparison-tool.component.html:159` (cog button) used to get hover effects from the old default and now silently won't

Yes. I'm still seeing hover effects.  Can you confirm?

libs/explorers/util/src/lib/svg-icon-tile/svg-icon-tile.component.ts
sagely1: Avatar or badge may be more appropriate for people-centric images according to AI.  This seemed to be more generic as we are adding a background shape to an icon.  Open to suggestions of course.

libs/explorers/util/src/lib/svg-icon-tile/svg-icon-tile.component.html
sagely1: For this simple case, it seemed like overkill.  My preference would be to consider documenting dependencies we bring in (not leave out) especially since every version of PrimeNG comes with a maintenance cost and feature parity does not exist across PrimeNG components.

libs/explorers/util/src/lib/modal-link/modal-link.component.ts
sagely1: Since `ModalLinkComponent` and `PopoverLinkComponent` have the same behavior, I can set the values in their corresponding html files.

libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.html
hallieswan: I'm seeing that the hover effect is missing from the following places: 
- GCT significance setting control (cog) and column control
- explorers CT significance setting control (cog) and columns control

There could be more, but I didn't perform a comprehensive check

libs/explorers/util/src/lib/svg-icon-tile/svg-icon-tile.component.ts
hallieswan: Could this just be a shape/backgroundColor variant on `SvgIconComponent` itself? Then we don't need a new name at all

libs/agora/gene-comparison-tool/src/lib/gene-comparison-tool.component.html
sagely1: You're right - I missed those two buttons.  Fixed in the next commit.

libs/explorers/util/src/lib/svg-icon-tile/svg-icon-tile.component.ts
sagely1: You're right.  I think we can just make this a feature of `SvgIconComponent`.  Refactoring in the next commit.

hallieswan (approved):
Looks good to me!

PR #4069, opened 2026-05-12
feat(qtl): add shared LinkBarComponent, update SvgIconComponent, qtl theme (QTL-94)

sagely1:
## Description

Introduces `LinkBarComponent`, a shared explorer UI component that will sit beneath the StatCards block on QTL pages. The component pairs a hyperlink with a circular PrimeNG icon button — both navigate to the same destination — inside a full-width container. Also extends `SvgIconComponent` with a configurable `strokeWidth` so the existing arrow asset and other future assets can render with a thinner stroke, and wires up QTL's PrimeNG theme to use the primary color defined in the QTL style guide.

## Related Issue

[QTL-94](https://sagebionetworks.jira.com/browse/QTL-94)

## Changelog

- Add `LinkBarComponent` to `libs/explorers/ui` with a routed text link and a PrimeNG icon-button arrow
- Add a `strokeWidth` input to `SvgIconComponent` for per-instance stroke override via a CSS custom property
- Configure QTL's PrimeNG primary palette anchored on `#2c5182` in `QtlPreset` (QTL-only theme override)
- Override QTL's solid primary button to lighten on hover/active (better visual feedback against a dark base)
- Add `primary: #2c5182` and a hover-color map to QTL's shared SCSS variables

## Testing

Note: Storybook colors represent the shared baseline that all three apps consume.  The QtlPreset ultimately determines the exact color so the color of the button in storybook will not match the design.  Also note that when this is added to QTL, a parent component (yet to be built) will encapsulate the stat cards and this component and will determine the look of the border.  Therefore, this component will have non-rounded borders.

No mouse hover:
image

With mouse hover:
image

New storybook entries for SvgIconComponent:
Default Stroke:
image

Thin Stroke:
image

Thick Stroke:
image

[QTL-94]: https://sagebionetworks.jira.com/browse/QTL-94?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/qtl/styles/src/lib/_variables.scss
```diff
@@ -1,3 +1,11 @@
+// -------------------------------------------------------------------------- //
+// Colors
+// -------------------------------------------------------------------------- //
+
 $main-colors: (
+  primary: #2c5182,
   secondary: #2c5182,
 );
+$main-colors-hover: (
+  action-primary: #4e6e96,
```
hallieswan: Just curious, where did this color come from? The [style guide](https://www.figma.com/design/TpCkecXfQ0RGXfVlyEKpxf/QTL-Explorer?node-id=2238-14128&m=dev) specifies #2C5182 as the action-primary color

libs/qtl/themes/src/lib/qtlPreset.ts
```diff
@@ -2,7 +2,36 @@ import { definePreset } from '@primeuix/themes';
...
+      100: '#b9c5d5',
+      200: '#95a8c0',
+      300: '#7089aa',
+      400: '#4e6e96',
+      500: '#2c5182',
+      600: '#254570',
+      700: '#1f395b',
+      800: '#182d48',
+      900: '#122034',
+      950: '#0b1421',
```
hallieswan, line 17: How were these colors determined? I didn't see these values in the [style guide](https://www.figma.com/design/TpCkecXfQ0RGXfVlyEKpxf/QTL-Explorer?node-id=2223-9866&p=f&m=dev), but I might've missed them!

hallieswan: Also, if these are the desired colors, do we need to update QTL's _variables file to match?

libs/qtl/themes/src/lib/qtlPreset.ts
```diff
@@ -2,7 +2,36 @@ import { definePreset } from '@primeuix/themes';
...
+    button: {
+      colorScheme: {
+        light: {
+          root: {
+            primary: {
+              hoverBackground: '{primary.400}',
+              hoverBorderColor: '{primary.400}',
+              activeBackground: '{primary.300}',
+              activeBorderColor: '{primary.300}',
+            },
```
hallieswan, line 33: Just curious, how does this differ from the default PrimeNG behavior and how did we land on this style vs the default?

hallieswan: Also, would we want to align the hover/active style for other PrimeNG buttons too, e.g. outlined or text buttons?

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.scss
```diff
@@ -0,0 +1,31 @@
...
+  gap: 16px;
+  padding: 16px 24px;
+  background-color: white;
+  border: 1px solid var(--color-border);
+}
+
+.explorers-link-bar .link-bar-text {
+  flex: 1;
+  text-align: left;
+  color: inherit;
```
hallieswan: Should we align the font style with the design? 


```suggestion
.explorers-link-bar .link-bar-text {
  flex: 1;
  text-align: left;
  color: var(--color-gray-800);
  font-size: 21px;
  font-weight: 600;
  line-height: 24px;
```

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.scss
```diff
@@ -0,0 +1,31 @@
...
+  &:hover {
+    text-decoration: underline;
+    cursor: pointer;
+  }
+}
+
+.explorers-link-bar .p-button {
+  width: 45px;
+  height: 35px;
+  border-radius: 999px;
```
hallieswan, line 33: Just curious, where did this border-radius value come from? The design specifies a border-radius of 50px. However, I'm wondering if we should use the default border-radius provided by the PrimeNG theme, then just override the theme for each app? That would keep button border radiuses consistent across each app. What do you think?

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.scss
```diff
@@ -0,0 +1,31 @@
+:host {
+  --color-arrow: white;
+}
+
+.explorers-link-bar {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 16px;
```
hallieswan: The gap value differs a bit from the design (24px) -- did we decide on another value here after testing?

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.html
```diff
@@ -0,0 +1,16 @@
...
+  <a class="link-bar-text" [routerLink]="link()">
+    <ng-content />
+  </a>
+  <p-button type="button" size="small" (onClick)="navigate()" [ariaLabel]="altText()">
+    <explorers-svg-icon
+      imagePath="explorers-assets/icons/card-arrow.svg"
+      altText="arrow"
+      [color]="'var(--color-arrow)'"
+      [height]="12"
+      [width]="13"
```
hallieswan, line 11: The arrow is 16px in the design -- did 12x13 look better in practice?

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.ts
```diff
@@ -0,0 +1,22 @@
...
+
+@Component({
+  selector: 'explorers-link-bar',
+  imports: [ButtonModule, RouterLink, SvgIconComponent],
+  templateUrl: './link-bar.component.html',
+  styleUrls: ['./link-bar.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class LinkBarComponent {
+  link = input.required<string>();
```
hallieswan, line 14: What do you think about using `routerLink` for this input name instead? Then we'd both match the `routerLink` directive argument and the existing input on `HomeCardComponent`. Also makes clear that it's not an arbitrary URL/href

libs/qtl/styles/src/lib/_variables.scss
hallieswan: We'll also need to add `$main-colors-hover` to `libs/qtl/styles/src/lib/_root.scss`

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.html
```diff
@@ -0,0 +1,16 @@
+<div class="explorers-link-bar">
```
hallieswan, line 1: Should the whole bar be the link? Right now the gap between the text and the arrow isn't clickable, and we duplicate the navigation between `[routerLink]` on the `<a>` and a manual `router.navigate()` on the `<p-button>`

hallieswan (changes requested):
Looking good! Just some questions about aligning with the design

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.html
```diff
@@ -0,0 +1,16 @@
...
+  </a>
+  <p-button type="button" size="small" (onClick)="navigate()" [ariaLabel]="altText()">
+    <explorers-svg-icon
+      imagePath="explorers-assets/icons/card-arrow.svg"
+      altText="arrow"
+      [color]="'var(--color-arrow)'"
+      [height]="12"
+      [width]="13"
+      [strokeWidth]="4"
+      [enableHoverEffects]="false"
```
rrchai: nit: redundant (default is already false)

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.html
sagely1: I thought that initially but when I saw the highlight behavior, there is no visual indication of the bar being a link, but there is for the hyperlink.  It would be strange for hyperlink to have an underline when the parent is hovered.

hallieswan: Got it, I could definitely see that! Should we double check with design, just to make sure we're interpreting the expected behavior correctly?

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.ts
sagely1: Naming it routerLink would be confusing because Angular uses this property name - in the case of the HomeCardComponent, there is an issue with that implementation because Agora imports RouterLink (and Model-AD doesn't) causing a collision.  The better approach is to avoid routerLink here and address the HomeComponent routerLink bug separately - standardizing on `link` would be the safer approach.

libs/qtl/styles/src/lib/_variables.scss
sagely1: This is the hover color (not the action-primary color) - the initial behavior was that hovering over the button was causing the hover effect to go darker which is not ideal.  The component has not been brought into the app so I will remove this since it can be added later when integrated into the home page.

libs/qtl/themes/src/lib/qtlPreset.ts
sagely1: > How were these colors determined? I didn't see these values in the [style guide](https://www.figma.com/design/TpCkecXfQ0RGXfVlyEKpxf/QTL-Explorer?node-id=2223-9866&p=f&m=dev), but I might've missed them!

Good question.  I had AI create these and it is centered around the style guide's #2c5182 at 500 and equally spaced lightness around this value.  This is best practice for PrimeNG components as these tokens are used internally in PrimeNG component templates.  Not having these creates a latent bug that may show up as missing shades.

> Also, if these are the desired colors, do we need to update QTL's _variables file to match?

No, we don't need to match these 11 hues in our simplified color system.  We are more used to thinking in terms of primary, secondary, etc. and not these values.

libs/qtl/themes/src/lib/qtlPreset.ts
sagely1: > Just curious, how does this differ from the default PrimeNG behavior and how did we land on this style vs the default?

Without this, mousing over buttons will cause the dark color used for buttons to go darker - our style guide specifies a very saturated, dark primary color and this made the hover effect hard to notice.  Best practice for accessibility dictate going lighter but it looks like PrimeNG always defaults to darker.

> Also, would we want to align the hover/active style for other PrimeNG buttons too, e.g. outlined or text buttons?

Good question.  It can be added in the future but since this change doesn't need these variants, I will just add the styles we need for now.

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.scss
sagely1: Added in next commit!

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.scss
sagely1: > Just curious, where did this border-radius value come from? The design specifies a border-radius of 50px. However, I'm wondering if we should use the default border-radius provided by the PrimeNG theme, then just override the theme for each app? That would keep button border radiuses consistent across each app. What do you think?

50px is a magic number - so 999px ensures we see the pill shape in Figma regardless of sizing (e.g. as we zoom into the page).  Your thought of keeping border radius consistent is interesting but I think we revisit it in the future once we've added more buttons and validated the intended look.

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.scss
sagely1: No this is copy paste error.  Fixed in next commit.

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.html
sagely1: > The arrow is 16px in the design -- did 12x13 look better in practice?
I am seeing that the icon size has a width of 13px in [Figma](https://www.figma.com/design/TpCkecXfQ0RGXfVlyEKpxf/QTL-Explorer?node-id=2953-41711&m=dev).

hallieswan: Ah, I misread the design! Thanks for double checking

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.scss
hallieswan: Got it! Most of QTL's other primary buttons use 50px border-radius (e.g. "Filter Results" and filter chips on the eQTL CT, search chips on the QTL homepage), so might be worth using that as the eventual QTL default when we revisit

libs/explorers/ui/src/lib/components/link-bar/link-bar.component.html
sagely1: I will be doing a design pass with Adam once this is integrated with the Stat Cards so he can make the final call at that point but it'll be easy enough to implement once he sees it.

hallieswan (approved):

hallieswan (approved):

PR #4073, opened 2026-05-18
feat(explorers): adopt DB v8 collation for case-insensitive sorting (AG-2010, MG-912)

rrchai:
## Description

Replace the `$toLower` + `$addFields` workaround (needed because DocumentDB 5.0 lacked collation) with native pipeline-level collation (`{locale: "en", strength: 2}`) now supported in DocumentDB 8.0. This simplifies CT repository code, improves query performance, and makes future CTs easier to build without per-column boilerplate.

> [!NOTE]
> Collation applies to all string comparisons in the pipeline (both `$sort` and `$match`). The `modified_genes` filter in Model Overview has a known case variant (`PSEN1`/`Psen1`) — discussed and accepted in [MG-993](https://sagebionetworks.jira.com/browse/MG-993?focusedCommentId=333026). Data restructure will address this long-term.

> [!NOTE]
> DocumentDB 8.0 requires collation to be configured (via `createIndex`) before it can be used in queries. Data-manager PRs must be deployed first:
> - https://github.com/Sage-Bionetworks/agora-data-manager/pull/158
> - https://github.com/Sage-Bionetworks/model-ad-data-manager/pull/43

## Related Issue

- [AG-2010](https://sagebionetworks.jira.com/browse/AG-2010)
- [MG-912](https://sagebionetworks.jira.com/browse/MG-912)

## Changelog

- Add `CASE_INSENSITIVE` collation to CT aggregation pipeline and count query in `ComparisonToolRepositorySupport`
- Remove `toLowerExpr()` method and all `getComputedSortFieldExpressions()` entries that only existed for case-insensitive sort
- Rename `arrayToLoweredStringExpr` → `arrayToStringExpr` (remove `$toLower` wrapper, collation handles case)
- Add `.collation()` to all 7 Mongoose sort queries in Agora Node.js API
- Add collated indexes to `collections-indexes.json` and local data loaders for both apps
- Add `name` → `name.link_text` sort alias in `CustomTranscriptomicsRepositoryImpl`
- Update `.claude/rules/ct-repository.md` for collation-based sorting

## Testing

- Start model-ad stack (`npx nx build-image model-ad-apex && npx nx serve-detach model-ad-apex`), navigate to Model Overview CT, verify sorting works on all columns
- Verify null/empty rows still sort to the end (isEmpty mechanism unchanged)
- In Transcriptomics CT, sort by `gene_symbol` and confirm `mt-*` genes appear in alphabetical M position (not at the end)
- Filter by `PSEN1` in Modified Gene dropdown — confirm it returns both `PSEN1` and `Psen1` models (expected behavior after collation, accepted in MG-993)
- Verify search/typeahead still works (not affected by collation)

[MG-993]: https://sagebionetworks.jira.com/browse/MG-993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AG-2010]: https://sagebionetworks.jira.com/browse/AG-2010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MG-912]: https://sagebionetworks.jira.com/browse/MG-912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/agora/data/src/main.py
```diff
@@ -154,6 +155,21 @@ def create_collections_indexes(
...
+    DocumentDB 8.0 requires a collation to be used in at least one collection
+    or index before it can be referenced in queries. This creates a dedicated
+    config collection with the collation to make it available cluster-wide.
+    Standard MongoDB does not require this step but is unaffected by it.
+    """
+    if "_collation_config" in db.list_collection_names():
+        logger.debug("Collation already registered")
+        return
+    db.create_collection("_collation_config", collation=Collation("en", strength=2))
+    logger.debug("Registered case-insensitive collation")
```
hallieswan: The local image uses MongoDB, not DocumentDB, so just curious if this is really required here? However, do we need to add this step to either the infra repos or the data manager scripts?

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/CustomTranscriptomicsRepositoryImpl.java
```diff
@@ -55,31 +55,17 @@ protected Class<TranscriptomicsDocument> getDocumentClass() {
...
   @Override
   protected Map<String, ComputedSortField> getComputedSortFieldExpressions() {
     return Map.of(
       GENE_SYMBOL_FIELD,
-      ComputedSortField.of(toLowerExpr(DISPLAY_GENE_SYMBOL_FIELD)).withPrerequisite(
+      ComputedSortField.of("$" + DISPLAY_GENE_SYMBOL_FIELD).withPrerequisite(
         buildDisplayGeneSymbolField()
-      ),
-      "name",
-      ComputedSortField.of(toLowerExpr("name.link_text")),
```
hallieswan, line 70: Does the collation handle mapping `name` to `name.link_text`? Or should we add an entry to `getSortFieldAliases`?

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
```diff
@@ -167,7 +171,11 @@ protected final Page<T> executePagedAggregation(Criteria matchCriteria, Pageable
         getDocumentClass()
       );
 
-      long total = mongoTemplate.count(new Query(matchCriteria), getCollectionName());
+      long total = mongoTemplate.count(
+        new Query(matchCriteria)
+          .collation(Collation.of("en").strength(2)),
+        getCollectionName()
+      );
```
hallieswan, line 184: Is this collation needed for the count or could we drop it?

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
```diff
@@ -158,7 +159,10 @@ protected final Page<T> executePagedAggregation(Criteria matchCriteria, Pageable
       // Permits disk spillover when the $sort working set exceeds 100MB; only activates
       // when needed -- required for deep pagination on large collections
       Aggregation aggregation = Aggregation.newAggregation(operations).withOptions(
-        AggregationOptions.builder().allowDiskUse(true).build()
+        AggregationOptions.builder()
+          .allowDiskUse(true)
+          .collation(Collation.of("en").strength(2))
```
hallieswan: Does adding collation at the aggregation-level change how all stages are performed? For example, will the pinned item filter comparisons be case-insensitive now, so "5xFAD" would be equivalent to "5XFAD"?

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
hallieswan, line 94: Should this comment be updated to reflect that `$toLowerExpr` is no longer needed? We also probably need to update `.claude/rules/ct-repository.md` to capture this change

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
hallieswan: If we decide to keep this, could we pull the `Collation.of("en".strength(2))` into a constant, so it can be reused for the aggregation and the count?

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
hallieswan: Could this method be updated to remove `$toLower`, since collation should handle that now?

apps/agora/api/src/components/comparison.ts
```diff
@@ -162,6 +162,7 @@ export async function getRnaComparisonGenes(model: string) {
     })
       .lean()
       .sort({ hgnc_symbol: 1, tissue: 1 })
+      .collation(CASE_INSENSITIVE_COLLATION)
```
hallieswan, line 165: It looks this newly adds case insensitive sort across queries in the agora-api -- just want to double check that was by design? Have we double checked that's the desired behavior across all of these endpoints?

hallieswan (changes requested):
This is looking good! Just a few questions for you

apps/agora/data/src/main.py
rrchai: good catch! the registration is only needed for DocumentDB and I put here was just for consistency. However, I should put the registration codes into data manager repo and I think I will drop this for MongoDB here - will open a separate PR for this in data manager repos soon.

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/CustomTranscriptomicsRepositoryImpl.java
rrchai: good catch! the `collation` will not automatically sort by `name.link_text` for this nested object - will fix in next commit(s).

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
rrchai: > Is this collation needed for the count or could we drop it?

Yes, needed for consistency. This ensures the count uses the same case-insensitive matching as the aggregation, so they can't diverge if data ever contains case variants (rare case, but defensive).

> could we pull the Collation.of("en".strength(2)) into a constant, so it can be reused for the aggregation and the count?

Good idea, will do in next commit(s).

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
rrchai: > Does adding collation at the aggregation-level change how all stages are performed?

Yes, collation applies to all stages including `$match`. The actual pinned filter values sent to API come from the row data itself, but yes the matching will be case-insensitive, so "5xFAD" and "5XFAD" would both match the same rows. This looks a correct behavior to me - is there any case (e.g. case variants represent two distinct values) you would think could be an issue?

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
rrchai: good call, will do in next commit(s).

rrchai: good call, the `$toLower` is redundant wrapper now - will drop it and update the function in next commit(s).

apps/agora/api/src/components/comparison.ts
rrchai: Yes, by design for consistency with the api-next sorting. I didn't expect this will have unexpected behavior other than maybe order change, but lmk if I miss anything.

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
hallieswan: Good question! Previously filter matching was exact-case, but with this change it will be case-insensitive for all filters -- pinned items, data filters (age, model type, etc.). However, I'm not sure whether that will ever be a problem in the actual data. 

@JessterB -- do you know whether it's possible that the UID for any two pinned items will differ only on casing across any current or future CTs? And more broadly, will we ever have two data filter values that differ only by casing and should return different rows?

apps/agora/api/src/components/comparison.ts
hallieswan: Got it! I'm not sure whether the current order is intentional or will be affected by case-insensitive sorting -- @JessterB do you know whether string fields included in the changes to `agora/api` (e.g. `hgnc_symbol`, `tissue`, `model`, etc...) are consistently cased across the data? If yes, then there shouldn't be any change in the sort. If no, then previously uppercase values sorted before lowercase ones (e.g. 'AZY' before 'abc'), but with collation the sort is purely alphabetical (e.g. 'abc' before 'AZY'). Is that acceptable?

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
JessterB: > across any current or future CTs?

While it is difficult to predict the future for all of time, our UIDs today are case-insensitive.

But say we encounter this scenario in the future when we try to build a cross-organism RNA CT with UID: gene_symbol, while the data contains a mixture of results for human and mouse genes: "APP", "App"

Solution: Add another value to the UID that makes it unique
UID: gene_symbol + organism
"app, human" != "app, mouse"

If there isn't a native field in the data to use for this, we can add one in the data pipeline
UID: gene_symbol + guid
"app, abc1234"!= "app,def876"

apps/model-ad/data/src/main.py
```diff
@@ -140,9 +141,14 @@ def create_collections_indexes(
             collection = db.get_collection(collection_name)
             for index in indexes:
                 collection.create_index(list(index.items()))
+            for index in collection_index_data.get("collatedIndexes", []):
+                key = list(index.items())
+                name = "_".join(f"{k}_{v}" for k, v in key) + "_ci"
+                collection.create_index(
+                    key, name=name, collation=Collation("en", strength=2)
+                )
```
rrchai, line 149: Add an explicit `_ci` name suffix for collation indexes because the DB auto-generates the same name for indexes with the same key (regardless of collation), causing a conflict when both binary and collated versions coexist (collation index will be ignored).

apps/agora/data/src/data/collections-indexes.json
```diff
@@ -6,15 +6,17 @@
       { "hgnc_symbol": 1 },
       { "alias": 1 },
       { "total_nominations": -1, "hgnc_symbol": 1 }
-    ]
+    ],
+    "collatedIndexes": [{ "hgnc_symbol": 1 }]
```
rrchai: We need both binary and collated indexes on `hgnc_symbol` (binary serves regex queries, collated serves .collation() queries). For simplicity, since only a few collated indexes are needed, I added all of them as additional indexes alongside the existing ones rather than modifying existing indexes to use collation directly. The `_ci` suffix avoids the name conflict when both share the same key.

apps/model-ad/data/src/main.py
sagely1: @rrchai what do you think about using `_index` instead for the name.  Since `ci` is a common acronym for continuous integration?

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
hallieswan: Thanks @JessterB! That's really helpful. 

@rrchai if you haven't already, what do you think about updating the docs to specify that all CT query params are now case-insensitive (`items`, each filter, sort, etc...), across the API descriptions of the various apps, the java docstring for the `ComparisonToolRepositorySupport` base class, and the `ct-repository` rule? Just thinking this might be the type of non-obvious change that we might want to document clearly, so we're aware of it when defining future CTs

rrchai: Yes, I've updated the `ComparisonToolRepositorySupport` javadoc and `ct-repository.md` rule. For the OpenAPI descriptions, I'd defer to a follow-up since the spec describes query params, and the javadoc already documents the DB filtering behavior.

apps/model-ad/data/src/main.py
rrchai: that makes sense, I will change to use `_collated` suffix to distinct from the regular index, in next commit(s).

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
```diff
@@ -167,32 +177,32 @@ protected final Page<T> executePagedAggregation(Criteria matchCriteria, Pageable
...
       log.error("Error executing aggregation on collection {}", getCollectionName(), e);
       throw e;
     }
   }
 
-  /** Convenience: {@code $toLower("$" + fieldPath)} as a MongoDB expression Document. */
-  protected static Object toLowerExpr(String fieldPath) {
-    return new Document("$toLower", "$" + fieldPath);
-  }
-
```
rrchai, line 182: no longer needed, collation handles case

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
```diff
@@ -167,32 +177,32 @@ protected final Page<T> executePagedAggregation(Criteria matchCriteria, Pageable
...
+  protected static Object arrayToStringExpr(String arrayField) {
     Document reduce = new Document(
       "$reduce",
       new Document()
         .append("input", "$" + arrayField)
         .append("initialValue", "")
         .append("in", new Document("$concat", List.of("$$value", "\u0000", "$$this")))
     );
-    return new Document("$toLower", reduce);
+    return reduce;
```
rrchai, line 204: Rename `arrayToLoweredStringExpr()` → `arrayToStringExpr()` — removes the `$toLower` wrapper from the `$reduce` expression, collation handles case on the reduced string

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/CustomDiseaseCorrelationRepositoryImpl.java
```diff
@@ -47,26 +46,6 @@ protected Class<DiseaseCorrelationDocument> getDocumentClass() {
...
-      ComputedSortField.of(toLowerExpr("sex")),
-      "model_type",
-      ComputedSortField.of(toLowerExpr("model_type")),
-      "matched_control",
-      ComputedSortField.of(toLowerExpr("matched_control")),
-      "cluster",
-      ComputedSortField.of(toLowerExpr("cluster"))
-    );
-  }
-
```
rrchai, line 69: no longer needed, collation now handles automatically

apps/agora/api/src/components/comparison.ts
rrchai: FYI: there's only one lowercase symbol (`snoZ196`) for `hgnc_symbol ` in agora, so this will shift from the bottom of the list to the S section.

apps/agora/data/src/main.py
```diff
@@ -146,9 +147,14 @@ def create_collections_indexes(
             collection = db.get_collection(collection_name)
             for index in indexes:
                 collection.create_index(list(index.items()))
+            for index in collection_index_data.get("collatedIndexes", []):
```
rrchai, line 150: On local MongoDB 8.0, collation works without needing indexes. This is for local dev consistency with prod - same for model-ad's.

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
```diff
@@ -167,32 +177,32 @@ protected final Page<T> executePagedAggregation(Criteria matchCriteria, Pageable
...
-  protected static Object toLowerExpr(String fieldPath) {
-    return new Document("$toLower", "$" + fieldPath);
-  }
-
   /**
-   * Convenience: reduces an array field into a single lowercased string with a NUL separator,
-   * producing a stable lexicographic sort key. Avoids MongoDB's "parallel arrays" limitation.
+   * Reduces an array field into a single NUL-separated string, producing a stable lexicographic
+   * sort key. Avoids MongoDB's "parallel arrays" limitation. Case-insensitive comparison is
+   * handled by pipeline-level collation.
```
hallieswan: Since `\0` is ignorable under the ICU collation, this method no longer produces a sort key that's unique per distinct array -- `["foo", "bar"]` -> `\0foo\0bar` now compares as equal to `["foobar"]` -> `\0foobar`, whereas they were distinguished in the previous binary comparison. Any tied rows can have indeterminate order across pages, so a row could appear on two pages or none. For this PR, I'm thinking that we could just tweak this comment, since "stable lexicographic sort key" no longer holds with collation?

Separately, this also made me realize that the indeterminate order for ties across pages was pre-existing across all sorts. I'm thinking we could add `_id` as a final tiebreaker, so all sorts are deterministic. Would you be willing to create a follow-up ticket to track the work?

hallieswan (approved):
Looks good! Just one small comment

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
rrchai: Good catch, I will update the comment in next commit. 

> Separately, this also made me realize that the indeterminate order for ties across pages was pre-existing across all sorts. I'm thinking we could add _id as a final tiebreaker, so all sorts are deterministic. Would you be willing to create a follow-up ticket to track the work?

Sounds good a idea to resolve the existing indeterminate order. Tickets are here: [AG-2181](https://sagebionetworks.jira.com/browse/AG-2181) and [MG-1034](https://sagebionetworks.jira.com/browse/MG-1034)

hallieswan (approved):

PR #4075, opened 2026-05-18
feat(explorers): refactor SvgIconComponent for performance (QTL-99)

sagely1:
## Description

Replaces the runtime HTTP-fetch model for SVG icons with a build-time registry that inlines icon contents into the bundle. The previous implementation depended on HttpClient, returned an Observable<SafeHtml>, and could not deliver icons synchronously — leading to a flash of missing icons on first paint and empty markup in SSR-rendered HTML. The new approach exposes a synchronous getSvg() returning SafeHtml directly, eliminates HttpClient from the dependency graph of every consumer, and ensures SSR output already contains the inline SVG markup.

## Related Issue

[QTL-99](https://sagebionetworks.jira.com/browse/QTL-99)

## Changelog

- Add tools/generate-svg-icon-registry.js and the pnpm generate:svg-icon-registry script to produce libs/explorers/services/src/lib/svg-icon-registry.gen.ts from libs/explorers/assets/icons/*.svg
- Refactor SvgIconService to read from the generated registry synchronously and drop the HttpClient dependency
- Tighten path validation in isValidImagePath to only accept explorers-assets/icons/<file>.svg (no longer accepts agora-assets/ or model-ad-assets/ prefixes)
- Update SvgIconComponent so svgContent is a computed signal bound directly to [innerHTML]
- Update the SvgIconServiceStub, component tests, and service tests to use the synchronous API and remove provideHttpClient()
- Add a CI step (Verify svg-icon registry is up-to-date) and a lint-staged / pre-commit hook that regenerates and stages the registry whenever a source SVG changes

## Testing

- Open any explorers app (e.g. nx serve qtl-app or nx serve model-ad-app) and confirm SVG icons render immediately on first paint, with no flash of missing icons
- Run pnpm generate:svg-icon-registry and confirm libs/explorers/services/src/lib/svg-icon-registry.gen.ts is unchanged afterwards (i.e. the committed file is in sync)
- Note: Tested added and removing an icon would update the gen file (svg-icon-registry.ts).  See commit history for details

[QTL-99]: https://sagebionetworks.jira.com/browse/QTL-99?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rrchai (commented):
Would you consider directly using Angular built-in `loader` - [ref](https://medium.com/netanelbasal/simplifying-file-imports-in-angular-with-new-loader-options-4e94a0eb18af), so that we could drop the generator script and the CI verification step?

sagely1:
> Would you consider directly using Angular built-in `loader` - [ref](https://medium.com/netanelbasal/simplifying-file-imports-in-angular-with-new-loader-options-4e94a0eb18af), so that we could drop the generator script and the CI verification step?

@rrchai The `loader` option is a valid, simpler alternative — simpler to implement and an OOB solution.  However it shifts icon registration/maintenance onto humans. The implemented approach keeps the filesystem as the single source of truth: adding/removing an SVG regenerates the registry automatically, and CI runs a check to ensure the registry is up to date.  There may be more pros/cons to consider as the number of icons increases but I don't expect we will for a while considering we only have ~15 total for all 3 explorers and there may only be 4 or 5 new icons in QTL.

rrchai (approved):
lgtm

.github/workflows/ci.yml
```diff
@@ -50,6 +50,12 @@ jobs:
           devcontainer exec --workspace-folder ../sage-monorepo bash -c ". ./dev-env.sh \
             && workspace-install-affected"
 
+      - name: Verify svg-icon registry is up-to-date
+        run: |
+          devcontainer exec --workspace-folder ../sage-monorepo bash -c ". ./dev-env.sh \
+            && pnpm generate:svg-icon-registry \
+            && git diff --exit-code libs/explorers/services/src/lib/svg-icon-registry.gen.ts"
```
hallieswan: Non-blocking nit: could consider adding a --check flag to the generator script so this path doesn't have to be hardcoded in this workflow, the husky hook, and lint-staged. e.g. with the flag, this could be simplified to something like this:


```suggestion
      - name: Verify svg-icon registry is up-to-date
        run: |
          devcontainer exec --workspace-folder ../sage-monorepo bash -c ". ./dev-env.sh \
            && pnpm generate:svg-icon-registry --check
```

hallieswan (approved):
Looks good to me!

hallieswan (approved):

PR #4061, opened 2026-05-08
feat(sage-monorepo): add Atlassian mcp (SMR-772)

rrchai:
## Description

This PR wires up the official Atlassian remote MCP server via a project-scoped `.mcp.json` at the repo root - as the pattern Anthropic [recommends](https://code.claude.com/docs/en/devcontainer#enforce-organization-policy) for sharing MCP servers across a team in a devcontainer:

> To make [MCP servers](https://code.claude.com/docs/en/mcp) available inside the container, define them at [project scope](https://code.claude.com/docs/en/mcp#mcp-installation-scopes) in a .mcp.json file at the repository root so they are checked in alongside your dev container configuration. Install any binaries that local stdio servers depend on in your Dockerfile, and add remote server domains to your network allowlist.

The mcp specific rule `.claude/rules/atlassian-mcp.md` is added to document the cloud ID, OAuth flow, and the rule that ticket lookups also pull related context (sub-tasks, linked issues, parent/epic).

**Why `.mcp.json` and not the `atlassian@claude-plugins-official` plugin?**

I tried the plugin first and verified empirically that Claude Code's project-scoped `enabledPlugins` does **not** auto-install HTTP-MCP plugins on rebuild — the plugin gets pre-fetched into the cache but is immediately marked `.orphaned_at` and never registers, leaving each new dev to manually run `/plugin install`. Also, the atlassian plugin is a wrapper of atlassian mcp with extra skills that we don't need under the hood, so it makes sense to only expose atlassian mcp.

Also, this PR fixes `.devcontainer/devcontainer.json` `postCreateCommand` race: when the Claude Code VS Code extension creates `~/.claude` first, `ln -sfn` silently puts the symlink inside that directory, so Claude Code reads/writes to the ephemeral home - the history dropdown (chat) can't auto-fetch prior sessions after rebuild.

## Related Issue

[SMR-772](https://sagebionetworks.jira.com/browse/SMR-772)

## Changelog

- Add `.mcp.json` at repo root pointing at the Atlassian remote MCP server (`https://mcp.atlassian.com/v1/mcp`), project-scoped and team-shared.
- Add `.claude/rules/atlassian-mcp.md` documenting cloud ID, first-time OAuth, common tools (`getJiraIssue`, `searchJiraIssuesUsingJql`, `getAccessibleAtlassianResources`), and the always-pull-related-context rule for ticket lookups.
[...]

PR #4078, opened 2026-05-20
feat(sage-monorepo): add figma plugin and rule (SMR-785)

hallieswan:
## Description

Enables the Figma MCP plugin and adds `.claude/rules/figma.md` so Claude translates Figma designs into idiomatic code for the explorers-based UI stack (explorers, agora, model-ad, qtl). The rule codifies token mapping, component reuse, and a "close enough" principle to avoid the new dependencies, pixel-pinned values, and overly prescriptive PrimeNG overrides that the raw Figma MCP output tends to introduce.

## Related Issue

- [SMR-785](https://sagebionetworks.jira.com/browse/SMR-785)

## Changelog

- Enable the `figma@claude-plugins-official` plugin in `.claude/settings.json`
- Add `.claude/rules/figma.md` defining Figma -> code conventions for the explorers-based UI stack (explorers, agora, model-ad, qtl)

## Testing

- Approve the new `figma@claude-plugins-official` plugin when prompted on first Claude Code launch
- Run `/mcp auth figma` and complete the OAuth flow
- Start a new Claude Code session in this repo and run `/memory` -- confirm `.claude/rules/figma.md` is listed as loaded
- Pick a small Figma node from a QTL/Agora/Model-AD design and ask Claude to implement it. Confirm Claude reads `libs/explorers/styles/src/lib/_variables.scss` and uses semantic tokens (e.g., `var(--color-text-secondary)`) over raw scale entries
- In the same task, confirm Claude reuses existing components (e.g., `LoadingIconComponent`) and PrimeNG props/variants instead of custom CSS animations or pixel-pinned overrides
- Confirm Claude reuses an existing icon before asking for a new asset, and asks for the file (rather than fetching a Figma asset URL) when no match exists
- Confirm `package.json` is unchanged after generation -- no new dependencies introduced

[SMR-785]: https://sagebionetworks.jira.com/browse/SMR-785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.claude/rules/figma.md
```diff
@@ -0,0 +1,83 @@
+# Figma -> Code
+
+**Scope:** these rules apply only to the explorers-based UI stack -- `libs/explorers/`, `libs/agora/`, `libs/model-ad/`, `libs/qtl/`, and `apps/{agora,model-ad,qtl}/`. They do not apply to bixarena or any other scope (different styling, components, and conventions). If you're working outside the listed scopes, ignore this file.
```
hallieswan, line 18: Path-scoped rules only trigger on file Reads in matching paths, so a Figma-to-new-component workflow that goes straight from MCP -> Write never loads the rule.

.claude/settings.json
```diff
@@ -24,6 +24,7 @@
   },
   "enabledPlugins": {
     "claude-md-management@claude-plugins-official": true,
-    "skill-creator@claude-plugins-official": true
+    "skill-creator@claude-plugins-official": true,
+    "figma@claude-plugins-official": true
```
rrchai: consider to sort by alpha

rrchai (approved):
LGTM - just one nit

rrchai:
One thing worth testing is whether the plugin properly installs the Figma MCP at the project level (for new developer) after rebuilding the devcontainer. I ran into a similar issue with the Jira plugin before: https://github.com/Sage-Bionetworks/sage-monorepo/pull/4061. You might consider to config Figma MCP as well (in `.mcp.json`). However, I don't consider it's a blocker - I leave it up to you.

.claude/settings.json
hallieswan: Good call! Done in the next commit

sagely1 (approved):
LGTM

hallieswan:
> One thing worth testing is whether the plugin properly installs the Figma MCP at the project level (for new developer) after rebuilding the devcontainer. I ran into a similar issue with the Jira plugin before: #4061. You might consider to config Figma MCP as well (in `.mcp.json`). However, I don't consider it's a blocker - I leave it up to you.

Good call! That was a bug. I've added Figma to `.mcp.json` to fix. I've left the plugin entry in `settings.json` as well, since it ships the `/figma-*` slash commands that the bare MCP server doesn't include. So `.mcp.json` handles registration, the plugin handles skills.

rrchai (approved):

PR #4084, opened 2026-05-28
feat(explorers): implement cache invalidation (MG-873)

sagely1:
## Description

The QTL, Agora, and Model-AD Spring Boot APIs cache MongoDB-backed query results in Caffeine to reduce DB load, but nothing invalidated those caches when the underlying data is updated — users saw stale results until the API was restarted. This PR introduces a shared `api-cache-invalidation` library that polls the `dataversion` Mongo document every 60 seconds and clears all caches the moment `data_version` changes. The `/data-version` endpoint itself is no longer cached so it always reflects the live state and can drive both clients and the new poller.

## Related Issue

- [MG-873](https://sagebionetworks.jira.com/browse/MG-873)

## Changelog

- Add new shared library `api-cache-invalidation` with Spring Boot auto-configuration, a `DataVersionProvider` SAM interface, and a `CacheInvalidationService` that polls `dataversion` every 60s and clears all named caches on change
- Remove `@Cacheable` from `DataVersionService.loadDataVersion()` in `qtl/api`, `agora/api-next`, and `model-ad/api-next` so `/data-version` always returns the live document
- Wire the shared library into all three Spring Boot APIs via a 3-line `DataVersionProvider` `@Bean` adapter (no `@EnableScheduling` needed in apps — the autoconfig brings it)
- Collapse ~830 LOC of duplicated invalidation code and tests across the three apps into one shared implementation

## Testing

Run the steps below once per Spring Boot API (`agora-api-next`, `model-ad-api-next`). The commands below show the agora variant; substitute names/ports per the table at the end.

1. **Run the API with the new code.** `SPRING_DATA_MONGODB_HOST=localhost pnpm nx serve <name>-api-next` in a separate terminal. Leave it running and watch the log.
2. **Verify startup poll.** Wait until `curl -fs http://localhost:<port>/actuator/health` returns `200`. The Gradle log should show **`Initial data_version observed: <N>`** within ~1s of startup.
3. **Warm the cache.** Hit a cacheable endpoint twice and confirm the second call is markedly faster:
   - agora: `curl -s -o /dev/null -w "%{time_total}s\n" http://localhost:3334/v1/drugs/CHEMBL2105759`
   - model-ad: `curl -s -o /dev/null -w "%{time_total}s\n" http://localhost:3334/v1/models/LOAD1.Ceacam1KO`
4. **Bump the data_version in Mongo.**
   ```
   docker exec <name>-mongo mongosh -u root -p changeme --authenticationDatabase admin <db> \
     --eval 'db.dataversion.updateOne({}, {$set: {data_version: "999"}})'
   ```
   (`<db>` is `qtl`, `agora`, or `model-ad`.)
5. **Verify `/data-version` is no longer cached.** Immediately `curl http://localhost:<port>/v1/data-version` — it must return `data_version: "999"` without waiting for the 60s poll.
6. **Wait ≤ 60s and verify the eviction log.** The Gradle log should print **`data_version changed`** on the next scheduled poll.
7. **Verify cold-read after eviction.** Re-hit the warmed endpoint from step 3. Locally it returns to roughly the uncached timing (~8ms vs ~4ms cached).
8. **Verify the cache re-warms.** Hit the same endpoint once more — back to ~4ms.
9. **Cleanup.** Restore `data_version` to its original value, stop the Gradle process, and `docker start <name>-api-next` to bring the prebuilt container back.

### Per-app verification matrix

| App | Stack port | API port | Mongo db | Cacheable endpoint for steps 5/9/10 |
| --- | --- | --- | --- | --- |
| agora-api-next | 8000 (apex) | 3334 | `agora` | `GET /v1/drugs/CHEMBL2105759` |
| model-ad-api-next | 8000 (apex) | 3334 | `model-ad` | `GET /v1/models/LOAD1.Ceacam1KO` |

### Measured local-stack numbers

Times below are end-to-end `curl` against the local stack. The Gradle log lines were observed exactly when the test plan above expects them. Poll cadence verified at 60s in every case.

| App | Initial poll | Cold first call | Cached call | Post-eviction cold call | Re-warmed call |
| --- | --- | --- | --- | --- | --- |
| agora-api-next | `data_version: 83` | **67 ms** | **4 ms** | **8 ms** | **4 ms** |
| model-ad-api-next | `data_version: 8` | **71 ms** | **4 ms** | **8 ms** | **4 ms** |

[MG-873]: https://sagebionetworks.jira.com/browse/MG-873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/explorers/api-cache-invalidation/src/main/java/org/sagebionetworks/cacheinvalidation/DataVersionProvider.java
```diff
@@ -0,0 +1,8 @@
+package org.sagebionetworks.cacheinvalidation;
```
hallieswan: What do you think about using a more specific package name here? e.g. 


```suggestion
package org.sagebionetworks.explorers.cacheinvalidation;
```

libs/explorers/api-cache-invalidation/src/main/java/org/sagebionetworks/cacheinvalidation/CacheInvalidationService.java
```diff
@@ -0,0 +1,46 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.cache.Cache;
+import org.springframework.cache.CacheManager;
+import org.springframework.scheduling.annotation.Scheduled;
+
+@Slf4j
+@RequiredArgsConstructor
+public class CacheInvalidationService {
+
+  private static final long POLL_INTERVAL_MS = 60_000L;
```
hallieswan: Nit: what do you think about renaming this variable to `POLL_DELAY_MS` to make it clear that we're using a `fixedDelay` (i.e. 60s after the previous execution completes) rather than a fixed wall-clock 60s interval?


```suggestion
  private static final long POLL_DELAY_MS = 60_000L;
```

libs/explorers/api-cache-invalidation/src/main/java/org/sagebionetworks/cacheinvalidation/CacheInvalidationService.java
```diff
@@ -0,0 +1,46 @@
...
+      log.info("Initial data_version observed: {}", currentVersion);
+      return;
+    }
+    if (!previous.equals(currentVersion)) {
+      log.info("data_version changed {} -> {}; clearing caches", previous, currentVersion);
+      clearAllCaches();
+    }
+  }
+
+  private void clearAllCaches() {
```
hallieswan: Could consider adding a comment here to clarify that caches are cleared sequentially rather than atomically, but that's okay for our current use case, e.g. something like: 


```suggestion
  // Caches are cleared sequentially, not atomically — a request can observe
  // a partially-invalidated state during this window. Acceptable because
  // data version changes are rare events, and any stale entry is corrected
  // on the next poll cycle at most 60 seconds later.
  private void clearAllCaches() {
```

libs/explorers/api-cache-invalidation/src/main/java/org/sagebionetworks/cacheinvalidation/CacheInvalidationService.java
```diff
@@ -0,0 +1,46 @@
...
+public class CacheInvalidationService {
+
+  private static final long POLL_INTERVAL_MS = 60_000L;
+
+  private final DataVersionProvider dataVersionProvider;
+  private final CacheManager cacheManager;
+  private final AtomicReference<String> lastSeenVersion = new AtomicReference<>();
+
+  @Scheduled(fixedDelay = POLL_INTERVAL_MS)
+  public void pollForDataVersionChange() {
```
hallieswan: Should we wrap the body of this method in a `try/catch`? If it throws an uncaught exception, Spring will silently stop scheduling future executions, so a single transient MongoDB error would permanently disable the poller until the next restart.

hallieswan (commented):
This is looking great! Just a couple suggestions

hallieswan (approved):
Looks good to me!

rrchai (approved):
LGTM!

PR #4091, opened 2026-06-02
fix(sage-monorepo): configure allowedHosts for SSR CommonEngine (AG-2113)

rrchai:
## Description

Configures explicit `allowedHosts` on `CommonEngine` to restore SSR on deployed environments after the `@angular/ssr` 20.3.17 bump broke it. The `'*'` wildcard is not supported in 20.3.x (only Angular 21+), so explicit subdomain patterns are used instead.

## Related Issue

- [AG-2113](https://sagebionetworks.jira.com/browse/AG-2113)

## Changelog

- Configure explicit `allowedHosts` on `CommonEngine` in all 4 Angular SSR apps (agora, model-ad, bixarena, qtl)

## Testing

- After merge, verify agora-dev SSR is working: open https://agora-dev.adknowledgeportal.org and confirm the browser console shows `[ConfigLoader] Using configuration transferred from server` (not `No transferred state found, loading from YAML files`)
- Confirm no CORS errors in the browser console when loading the page
- Repeat for dev.modeladexplorer.org

[AG-2113]: https://sagebionetworks.jira.com/browse/AG-2113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/agora/app/server.ts
+++ b/apps/agora/app/server.ts
@@ -12,7 +12,9 @@ export function app(): express.Express {
   const browserDistFolder = resolve(serverDistFolder, '../browser');
   const indexHtml = join(serverDistFolder, 'index.server.html');
 
-  const commonEngine = new CommonEngine();
+  const commonEngine = new CommonEngine({
+    allowedHosts: ['*'],
+  });
 
   server.set('view engine', 'html');
   server.set('views', browserDistFolder);
--- a/apps/agora/app/server.ts
+++ b/apps/agora/app/server.ts
@@ -13,7 +13,7 @@ export function app(): express.Express {
   const indexHtml = join(serverDistFolder, 'index.server.html');
 
   const commonEngine = new CommonEngine({
-    allowedHosts: ['*'],
+    allowedHosts: ['localhost', '*.adknowledgeportal.org'],
   });
 
   server.set('view engine', 'html');
--- a/apps/bixarena/app/server.ts
+++ b/apps/bixarena/app/server.ts
@@ -11,7 +11,9 @@ export function app(): express.Express {
   const browserDistFolder = resolve(serverDistFolder, '../browser');
   const indexHtml = join(serverDistFolder, 'index.server.html');
 
-  const commonEngine = new CommonEngine();
+  const commonEngine = new CommonEngine({
+    allowedHosts: ['*'],
+  });
 
   server.set('view engine', 'html');
   server.set('views', browserDistFolder);
--- a/apps/bixarena/app/server.ts
+++ b/apps/bixarena/app/server.ts
@@ -12,7 +12,7 @@ export function app(): express.Express {
   const indexHtml = join(serverDistFolder, 'index.server.html');
 
   const commonEngine = new CommonEngine({
-    allowedHosts: ['*'],
+    allowedHosts: ['localhost', '*.bioarena.io', 'bioarena.io', '*.elb.amazonaws.com'],
   });
 
   server.set('view engine', 'html');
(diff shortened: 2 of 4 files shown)


apps/agora/app/server.ts
```diff
@@ -12,7 +12,9 @@ export function app(): express.Express {
   const browserDistFolder = resolve(serverDistFolder, '../browser');
   const indexHtml = join(serverDistFolder, 'index.server.html');
 
-  const commonEngine = new CommonEngine();
+  const commonEngine = new CommonEngine({
+    allowedHosts: ['*'],
```
sagely1: This seems overly permissive but instead of configuring `allowedHosts` as *, what if we bumped to a more recent version.  It's possible we don't have to configure `allowedHosts` in 20.3.26.  It's still recommended to configure explicitly for security.

sagely1 (changes requested):
Let's try a higher version of the package first.

apps/agora/app/server.ts
rrchai: Unfortunately, the issue remains in `20.3.26` (in prod) - confirmed in #4093 with the same error `[ConfigLoader] Browser configuration (loaded from YAML, no SSR)`

>  It's still recommended to configure explicitly for security

Setting allowedHosts as * restores the same behavior before this was introduced (the app is protected by ALB + Caddy in a private subnet). We can do explicit lists, but that means we need to maintain the list per app. Happy to go with explicit hosts if you prefer - just wanted to flag the trade-off. What do you think?

rrchai: Actually, I realize allowedHosts: ['*'] is actually not supported until Angular 21!
This makes only explicit work now :). I will add explicit list for allowedHosts per app in this PR, and we could revisit if needed to change when we upgrade Angular.

apps/bixarena/app/server.ts
```diff
@@ -11,7 +11,9 @@ export function app(): express.Express {
   const browserDistFolder = resolve(serverDistFolder, '../browser');
   const indexHtml = join(serverDistFolder, 'index.server.html');
 
-  const commonEngine = new CommonEngine();
+  const commonEngine = new CommonEngine({
+    allowedHosts: ['localhost', '*.bioarena.io', 'bioarena.io', '*.elb.amazonaws.com'],
```
rrchai, line 15: bixarena does not have a custom domain for dev

apps/qtl/app/server.ts
```diff
@@ -12,7 +12,9 @@ export function app(): express.Express {
   const browserDistFolder = resolve(serverDistFolder, '../browser');
   const indexHtml = join(serverDistFolder, 'index.server.html');
 
-  const commonEngine = new CommonEngine();
+  const commonEngine = new CommonEngine({
+    allowedHosts: ['localhost'],
```
rrchai: not sure what domain will look like for qtl, only localhost for now

hallieswan: @rrchai domains for QTL are defined in [QTL-51](https://sagebionetworks.jira.com/browse/QTL-51), if we want to proactively add them. Otherwise, would you mind adding a follow-up ticket to add the QTL domains here?

rrchai: cool! Thanks - I will add them in this pr using '*.adknowledgeportal.org' (same pattern as agora)

hallieswan (approved):
Looks good to me!

sagely1 (approved):
LGTM!

PR #4092, opened 2026-06-02
feat(qtl): update fonts, remove duplication (QTL-105)

sagely1:
## Description

QTL adds a new font family.  This PR adds the new font, updates the version of the existing font-family and removes duplication in font handling and unneeded styles.

## Related Issue

[QTL-105](https://sagebionetworks.jira.com/browse/QTL-105)

## Changelog

- Move font loading and `--font-family-primary` declaration into per-app `_fonts.scss`
  (agora, model-ad, qtl) so each app owns its font choice
- Remove the shared font import from `libs/explorers/styles` and delete its `_fonts.scss`
- Consolidate base `html, body` typography rules into `libs/explorers/styles/src/lib/_general.scss`
  using `var(--font-family-primary, sans-serif)`, removing duplicated declarations from each
  per-product `_general.scss`
- Replace the hardcoded `'DM Sans Variable'` reference in the explorers tooltip with
  `var(--font-family-primary, sans-serif)`
- Bump `@fontsource-variable/dm-sans` from `5.2.5` to `5.2.8`
- Adds support for `--font-family-secondary` in QTL's `_font.scss`

## Testing

- Inspect `:root` in DevTools for each app and verify `--font-family-primary` resolves to
  `'DM Sans Variable', sans-serif`
- Hover an element with a PrimeNG tooltip in any explorer app and confirm the tooltip text
  uses DM Sans Variable
- Confirm there is no visible regression in heading/body spacing across the three apps
  (margins/line-height should match `main`)

[QTL-105]: https://sagebionetworks.jira.com/browse/QTL-105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/explorers/styles/src/lib/components/_tooltip.scss
```diff
@@ -1,5 +1,5 @@
 .p-tooltip {
-  font-family: 'DM Sans Variable', sans-serif;
+  font-family: var(--font-family-primary, sans-serif);
```
sagely1, line 2: no longer hard-coded

package.json
```diff
@@ -26,7 +26,8 @@
     "@angular/platform-server": "20.3.1",
     "@angular/router": "20.3.1",
     "@angular/ssr": "20.3.17",
-    "@fontsource-variable/dm-sans": "5.2.5",
+    "@fontsource-variable/dm-sans": "5.2.8",
```
sagely1, line 29: update version

libs/explorers/styles/src/lib/components/_fonts.scss
```diff
@@ -1,5 +0,0 @@
-@use '@fontsource-variable/dm-sans';
-
-@font-face {
-  font-family: 'DM Sans Variable', sans-serif;
-}
```
hallieswan, line 5: This causes the explorers-storybook to lose the default font-family -- should we keep this font-family as the explorers default, and just override where needed per-app instead?
For example, see the "With Swatch" story:

[docs site](https://sage-bionetworks.github.io/sage-monorepo/storybook/?path=/story/explorers_ui-simpletablecomponent--with-swatch) | this PR (nx start explorers-storybook)
:---: | :---: 
Image | Image

libs/qtl/styles/src/lib/components/_fonts.scss
```diff
@@ -1 +1,5 @@
 @use '@fontsource-variable/dm-sans';
+
+:root {
+  --font-family-primary: 'DM Sans Variable', sans-serif;
```
hallieswan: The PR description notes that QTL uses a new font-family, but this looks like the same font-family that's used in Agora and Model AD, and `@fontsource/dm-serif-text` is added to `package.json` but isn't `@use`d anywhere. Is that by design? If the new font is only used for some components, would we define it as `--font-family-secondary` or something like that?

libs/explorers/styles/src/lib/_general.scss
```diff
@@ -1,10 +1,9 @@
...
 html,
 body {
-  font-family: var(--font-family-primary);
+  font-family: var(--font-family-primary, sans-serif);
+  font-size: var(--font-size-md);
+  line-height: 1.3;
+  font-weight: 400;
 }
 
 /* Remove default margin and padding */
```
hallieswan, line 9: The old per-app `_general.scss` files reset `margin: 0; padding: 0` on both `html` and `body`. This block covers `body` but not `html` -- should we add `html` here to preserve the explicit reset?

hallieswan (changes requested):
Looks good! Just a couple small questions

libs/qtl/styles/src/lib/components/_fonts.scss
sagely1: As far as I can see for now, it looks like we will need this new font-family as part of the QTL home page header but the existing font-family can be used as a default.  I'll add `font-family-secondary` in the next commit.

libs/explorers/styles/src/lib/components/_fonts.scss
sagely1: Good point - I think we can assume DM Sans for the default.  Added in next commit.

libs/explorers/styles/src/lib/_general.scss
sagely1: added in next commit

libs/agora/styles/src/lib/components/_fonts.scss
```diff
@@ -1 +1,5 @@
 @use '@fontsource-variable/dm-sans';
+
+:root {
+  --font-family-primary: 'DM Sans Variable', sans-serif;
+}
```
hallieswan: Nit: we could probably remove this override, since we set the same value in explorers now

libs/model-ad/styles/src/lib/components/_fonts.scss
```diff
@@ -1 +1,5 @@
 @use '@fontsource-variable/dm-sans';
+
+:root {
+  --font-family-primary: 'DM Sans Variable', sans-serif;
+}
```
hallieswan: Same nit here -- we could probably remove this in favor of the explorers default

hallieswan (approved):
Looks good! Just two small optional suggestions

hallieswan (approved):

PR #4094, opened 2026-06-02
feat(agora): redesign teams page with card-based layout (AG-2111)

hallieswan:
## Description

The Agora contributing teams page has been redesigned to use a card-based layout matching the new Figma design. Each team is now displayed as a distinct white card with rounded corners, sitting directly on the hero background image. A new purely presentational `TeamCardComponent` was introduced to represent each team, making it independently testable and Storybook-ready without any service dependencies.

## Related Issue

- [AG-2111](https://sagebionetworks.jira.com/browse/AG-2111)

## Changelog

- Add `TeamCardComponent` -- presentational card showing team title, description, and member avatars with PI label
- Add Storybook story for `TeamCardComponent` with Default, WithImages, NoProgram, and NoMembers variants
- Move image fetching to `TeamListComponent`, scoped per team and passed down as a per-card image map
- Merge teams page into a single hero section so cards render on the background image
- Update team member avatar placeholder to new SVG icon design
- Remove `TeamMemberListComponent` and its unused route files (replaced by `TeamCardComponent`)

## Testing

- Navigate to the Contributing Teams page (`/teams`) and confirm each team renders as a white rounded card on the hero background image
- Confirm the card title shows `Program: Team Full Name` when a program exists, and just `Team Full Name` when it does not
- Confirm team member avatars load -- members with a real photo show it; others show the placeholder SVG icon
- Confirm the Primary Investigator label appears in bold below the PI's name and that the PI is listed first
- Confirm member names with a URL render as clickable links
- Open Storybook (`nx start agora-storybook`) and verify all four `Teams/TeamCard` stories render correctly

### Preview

> [!NOTE]
> Team images are not downloaded in the local build, so the actual team member images are not displayed. The storybook story below demonstrates that images are still shown when they are loaded.

prod | this PR 
:---: | :---: 
image | image

`nx start agora-storybook`

See that the Agora social image is displayed when provided as a team member's image: 

image

[AG-2111]: https://sagebionetworks.jira.com/browse/AG-2111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/agora/teams/src/lib/team-card/team-card.component.ts
```diff
@@ -0,0 +1,35 @@
+import { Component, computed, input } from '@angular/core';
+import { Team, TeamMember } from '@sagebionetworks/agora/api-client';
+
+@Component({
+  selector: 'agora-team-card',
+  imports: [],
+  templateUrl: './team-card.component.html',
+  styleUrl: './team-card.component.scss',
+})
+export class TeamCardComponent {
```
hallieswan, line 10: A purely presentational component -- now teams-list handles fetching the team info and team member images

libs/agora/teams/src/lib/team-list/team-list.routes.ts
hallieswan, line 1: Remove these unused routes

libs/agora/teams/src/lib/team-member-list/team-member-list.routes.ts
hallieswan, line 1: Component and routes are unused now

libs/agora/teams/src/lib/teams.component.scss
```diff
@@ -0,0 +1,7 @@
+.section {
+  min-height: 100vh;
+}
```
hallieswan, line 3: Ensure background hero image takes up entire height while teams data is loading

libs/agora/assets/images/team-member.svg
```diff
@@ -1,3 +1,12 @@
...
+<svg width="47" height="50" viewBox="0 0 47 50" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M45.7379 49.5068H0.616002C0.275784 49.5068 0 49.231 0 48.8908V40.4064C0 33.2729 5.80353 27.4695 12.9371 27.4695H17.4973C17.8375 27.4695 18.1133 27.7453 18.1133 28.0858C18.1133 28.4264 17.8375 28.7018 17.4973 28.7018H12.9371C6.48291 28.7018 1.23201 33.9526 1.23201 40.4068V48.2752H45.1219V40.4068C45.1219 33.953 39.871 28.7022 33.4168 28.7022H28.8622C28.522 28.7022 28.2462 28.4268 28.2462 28.0862C28.2462 27.7456 28.522 27.4698 28.8622 27.4698H33.4168C40.5504 27.4698 46.3539 33.2733 46.3539 40.4068V48.8912C46.3539 49.2314 46.0781 49.5072 45.7376 49.5072L45.7379 49.5068Z" fill="white"/>
```
sagely1, line 2: consider using `currentColor` instead of hard-coded colors

libs/agora/teams/src/lib/team-list/team-list.component.ts
```diff
@@ -1,29 +1,54 @@
...
+  private readonly logger = inject(LoggerService);
+  private readonly teamService = inject(TeamService);
+
   _teams: Team[] = [];
-  get teams(): Team[] {
-    return this._teams;
-  }
+  imagesByTeam: Partial<Record<string, Record<string, string>>> = {};
 
   @Input() set teams(teams: Team[]) {
```
sagely1: It might be a good time to update this to the new `input` syntax for consistency

libs/agora/teams/src/lib/team-list/team-list.component.ts
```diff
@@ -1,29 +1,54 @@
...
 export class TeamListComponent {
+  private readonly destroyRef = inject(DestroyRef);
+  private readonly logger = inject(LoggerService);
+  private readonly teamService = inject(TeamService);
+
   _teams: Team[] = [];
-  get teams(): Team[] {
-    return this._teams;
-  }
+  imagesByTeam: Partial<Record<string, Record<string, string>>> = {};
```
sagely1: these could be private readonly as well

libs/agora/teams/src/lib/team-card/team-card.component.html
```diff
@@ -0,0 +1,32 @@
+<div class="team-card">
+  <div class="team-card-textbox">
+    <h2 class="team-card-title">{{ title() }}</h2>
+    <p class="team-card-description">{{ description() }}</p>
+  </div>
+  <div class="team-card-members">
+    @for (member of sortedMembers(); track $index) {
```
sagely1, line 7: $index seems unnecessary.  $member.name may be better for tracking

libs/agora/assets/images/team-member.svg
hallieswan: I considered this! However, I think `currentColor` won't get picked up when the SVG is used as a `background-image` -- we could work around it with a CSS mask, but I don't think it's worth the added complexity here, especially since the previous `team-member.svg` also hardcoded its fill and the other SVGs in the `images/` asset directories all hardcode their colors too. What do you think?

libs/agora/teams/src/lib/team-list/team-list.component.ts
hallieswan: The tests access `imagesByTeam` directly, so `private readonly` would break them -- one test writes to it to set up pre-state and others read it to assert on the result. We'd need to refactor the tests to make that work. Happy to make that change if you feel strongly that this should be `private readonly`

libs/agora/teams/src/lib/team-card/team-card.component.html
hallieswan: `member.name` isn't guaranteed to be unique (and `TeamMember` doesn't have a unique id field), so I used `index` to guard against a team with members with the same name. I can use `member.name` if we aren't worried about duplicates

libs/agora/teams/src/lib/team-list/team-list.component.ts
hallieswan: Good idea! Will do in the next commit

libs/agora/assets/images/team-member.svg
sagely1: You're right.  Seems like a side-quest for flexibility we aren't needing today.  Let's leave the hardcoding and look into it if you ever need to customize it.

sagely1 (changes requested):

libs/agora/teams/src/lib/team-list/team-list.component.ts
```diff
@@ -1,29 +1,54 @@
...
 
   @Input() set teams(teams: Team[]) {
     this._teams = teams;
-  }
+    this.imagesByTeam = {};
```
sagely1: as part of the `input` refactor, we may want to use an `effect`

libs/agora/teams/src/lib/team-list/team-list.component.ts
sagely1: I think if these are converted to `input` we get `readonly` implicitly which is good enough.

rrchai (approved):
LGTM

rrchai:
By having another look, I notice the avatars of members have uneven spacings - members with longer names is wider, and this seems a pre-existing behavior. Would we want to address this in this PR, a follow-up, or was it intentional by design?

Screenshot 2026-06-03 at 1 11 49 PM

hallieswan:
> By having another look, I notice the avatars of members have uneven spacings - members with longer names is wider, and this seems a pre-existing behavior. Would we want to address this in this PR, a follow-up, or was it intentional by design?

@rrchai good call! The spacing is uneven in the [design](https://www.figma.com/design/dgrq8yWWywKQI3rserIR60/Agora-4.0?node-id=678-5950&m=dev), but it's less noticeable since the non-primary members all have the same name. Since this is pre-existing behavior (and we may need additional input from design about what an "equal-spaced" layout would look like), would you be alright if we merge this PR, and separately, I'll reach out to design to see if this is intentional or if we need a follow-up?

sagely1 (approved):
LGTM

PR #4099, opened 2026-06-03
fix(explorers): fix Sentry sourcemap upload (MG-902)

sagely1:
## Description

Sentry stack traces for Agora, Model-AD, and QTL resolved to minified chunk paths instead of
the original TypeScript source, because the deployed bundles and the uploaded sourcemaps
carried mismatched Debug IDs. The Nx `build-image` task depends on the cached `build` output,
so when it ran after the separate `sentry-sourcemaps-inject` step, Nx restored the
pre-injection bundles from cache and stripped the Debug IDs back out before packaging the
image — publishing bundles with no Debug ID while Sentry held maps that did. This change makes
`build-image` depend on `sentry-sourcemaps-inject` so injection happens inside the same Nx task
graph (after the cache restore), guaranteeing the published image and the uploaded sourcemaps
share the same content-derived Debug IDs.

## Related Issue

[MG-902](https://sagebionetworks.jira.com/browse/MG-902)

## Changelog

- Make the `build-image` target depend on `sentry-sourcemaps-inject` in `apps/agora/app`, `apps/model-ad/app`, and `apps/qtl/app`, so Debug IDs are injected within the same task graph and survive the cached `build` restore
- Split the single `sentry-sourcemaps` Nx target into `sentry-sourcemaps-inject` (depends on `build`) and `sentry-sourcemaps-upload` (depends on `sentry-sourcemaps-inject`) across the three apps
- Point the post-image Sentry step in `.github/workflows/ci.yml` and `.github/workflows/release.yml` at `sentry-sourcemaps-upload`
- Rely on the `build-image` → `inject` dependency rather than a manual ordering step in the workflows

## Testing

- Trigger a CI run on `main` (push event) and confirm the `build-image` task graph runs `build` → `sentry-sourcemaps-inject` → `build-image` (visible in the Nx task output)
- Trigger a Sentry error on the Sentry test page
- In Sentry, open a new error event from the edge deployment and confirm the stack frame resolves to a `.ts` source file (e.g. `app.component.ts:42`) instead of `/chunk-XXXXX.js`

[MG-902]: https://sagebionetworks.jira.com/browse/MG-902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/agora/app/project.json
+++ b/apps/agora/app/project.json
@@ -149,17 +149,23 @@
       },
       "dependsOn": []
     },
-    "sentry-sourcemaps": {
+    "sentry-sourcemaps-inject": {
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "sentry-cli sourcemaps inject dist/apps/agora/app"
+      },
+      "dependsOn": ["build"]
+    },
+    "sentry-sourcemaps-upload": {
       "executor": "nx:run-commands",
       "options": {
         "commands": [
-          "sentry-cli sourcemaps inject dist/apps/agora/app",
           "sentry-cli sourcemaps upload --org sage-bionetworks --project agora --release agora@$SENTRY_RELEASE_SUFFIX dist/apps/agora/app",
           "find dist/apps/agora/app -name '*.map' -delete"
         ],
         "parallel": false
       },
-      "dependsOn": ["build"]
+      "dependsOn": ["sentry-sourcemaps-inject"]
     },
     "scan-image": {
       "executor": "nx:run-commands",
--- a/apps/agora/app/project.json
+++ b/apps/agora/app/project.json
@@ -168,7 +168,10 @@
       "dependsOn": ["sentry-sourcemaps-inject"]
     },
     "build-image": {
-      "dependsOn": ["sentry-sourcemaps-inject"]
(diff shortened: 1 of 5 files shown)


apps/agora/app/project.json
```diff
@@ -149,17 +149,23 @@
...
       "dependsOn": []
     },
-    "sentry-sourcemaps": {
+    "sentry-sourcemaps-inject": {
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "sentry-cli sourcemaps inject dist/apps/agora/app"
+      },
+      "dependsOn": ["build"]
+    },
```
rrchai, line 158: Out of curiosity, currently both `build-image` and `sentry-sourcemaps-inject` depend on build task. Will `build-image` task hits the nx cache on build and cause an issue on the injection?

sagely1: I think you're right.  I'll make the fix in the next commit.

rrchai (approved):
LGTM

rrchai (approved):

hallieswan (approved):
Looks good to me!

PR #4103, opened 2026-06-09
feat(explorers): add opt-in row selection and hover highlighting to comparison tool (QTL-49)

hallieswan:
## Description

Adds opt-in row selection and hover highlighting to the comparison tool. When enabled, the first row is automatically selected on load; clicking a different row moves the selection to it; the selection persists across sorting and pagination and resets to the first row only when filters remove it from all results. Hovering a row highlights it independently of the selection. Both behaviors are off by default and can be enabled per CT instance.

## Related Issue

- [QTL-49](https://sagebionetworks.jira.com/browse/QTL-49)

## Changelog

- Add opt-in row selection: auto-selects the first row on load, persists across sort and pagination, resets to first row when filters remove the selection from all results
- Add opt-in row hover highlighting that appears on mouseenter and clears on mouseleave
- Clicking a heatmap circle also updates the selected row
- Add `notifySelectedRowValidity(isInResults: boolean)` to `ComparisonToolService` -- consumers must call this after each filtered fetch to trigger selection reset when the selected row is no longer in results
- Introduce `--color-comparison-tool-row-highlight` CSS custom property (via `$comparison-tool-colors` in explorers `_variables.scss`) for per-product override
- Add unit tests and Storybook stories for both behaviors
- Add scaffolding to `EqtlComparisonToolComponent` for wiring up row selection when CT data fetching is implemented (QTL-113)

## Testing

> [!NOTE]
> Pin/unpin and sort/pagination behavior with row selection will be tested fully once the eQTL CT data fetching is implemented. Tracked in [QTL-113](https://sagebionetworks.jira.com/browse/QTL-113).

- Open the CT Demo story in Storybook with `rowSelectionEnabled` and `rowHoverEnabled` toggled on and confirm the first unpinned row is highlighted on load.
- Click a different row and confirm the highlight moves to it; click the same row again and confirm nothing changes.
- Click a heatmap circle and confirm the popover opens and the row selection moves to that row.
- Click within the primary column (model name / pin button area) and confirm the row selection moves.
- Hover over any row and confirm the blue highlight appears only while the cursor is on that row.
- Toggle `rowSelectionEnabled` off in the Storybook controls and confirm no rows are highlighted, clicking does nothing, and heatmap circle popovers still open.

### Preview

`nx start explorers-storybook`

https://github.com/user-attachments/assets/5d8322f8-507d-455a-a504-ef4564bed7a6

[QTL-49]: https://sagebionetworks.jira.com/browse/QTL-49?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QTL-113]: https://sagebionetworks.jira.com/browse/QTL-113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/comparison-tool/src/lib/comparison-tool-table/base-table/base-table.component.html
+++ b/libs/explorers/comparison-tool/src/lib/comparison-tool-table/base-table/base-table.component.html
@@ -22,11 +22,21 @@
     </ng-template>
 
     <ng-template pTemplate="body" let-rowData>
-      <tr>
+      @let rowId = rowData[viewConfig().rowIdDataKey];
+      <tr
+        (click)="onRowClick(rowId)"
+        (mouseenter)="onRowMouseEnter(rowId)"
+        (mouseleave)="onRowMouseLeave()"
+        [class.selected]="
+          viewConfig().rowSelectionEnabled && comparisonToolService.selectedRowId() === rowId
+        "
+        [class.hovered]="
+          viewConfig().rowHoverEnabled && comparisonToolService.hoveredRowId() === rowId
+        "
+      >
         @for (column of selectedColumns(); track column) {
           @let columnType = column.type;
           @let cellData = rowData[column.data_key];
-          @let rowId = rowData[this.viewConfig().rowIdDataKey];
           <td
             [style]="{ width: columnWidths()[column.data_key] || 'auto' }"
             [class]="columnType"
--- a/libs/explorers/comparison-tool/src/lib/comparison-tool-table/base-table/base-table.component.html
+++ b/libs/explorers/comparison-tool/src/lib/comparison-tool-table/base-table/base-table.component.html
@@ -22,7 +22,7 @@
     </ng-template>
 
     <ng-template pTemplate="body" let-rowData>
-      @let rowId = rowData[viewConfig().rowIdDataKey];
+      @let rowId = String(rowData[viewConfig().rowIdDataKey]);
       <tr
         (click)="onRowClick(rowId)"
         (mouseenter)="onRowMouseEnter(rowId)"
--- a/libs/explorers/comparison-tool/src/lib/comparison-tool.component.stories.ts
+++ b/libs/explorers/comparison-tool/src/lib/comparison-tool.component.stories.ts
@@ -5,6 +5,7 @@ import { provideRouter } from '@angular/router';
 import {
   ComparisonToolConfig,
   ComparisonToolViewConfig,
+  HeatmapCircleClickTransformFnContext,
   LegendPanelConfig,
   SynapseWikiParams,
 } from '@sagebionetworks/explorers/models';
@@ -38,6 +39,9 @@ type StoryArgs = Omit<
   visualizationOverviewVisible?: boolean;
   // Sidebar
   hasSidebar?: boolean;
+  // Row interaction
+  rowSelectionEnabled?: boolean;
+  rowHoverEnabled?: boolean;
 };
 
 // Inner component that provides services - will be remounted when configs change
@@ -89,6 +93,9 @@ class ComparisonToolInnerComponent {
   visualizationOverviewVisible = input<boolean>();
   // Sidebar input
   hasSidebar = input<boolean>();
+  // Row interaction inputs
+  rowSelectionEnabled = input<boolean>();
+  rowHoverEnabled = input<boolean>();
 
   private readonly comparisonToolService = inject(ComparisonToolService);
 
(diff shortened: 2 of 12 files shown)


libs/explorers/comparison-tool/src/lib/comparison-tool.component.stories.ts
```diff
@@ -129,6 +136,24 @@ class ComparisonToolInnerComponent {
       viewConfig.allowPinnedImageDownload = this.allowPinnedImageDownload();
       viewConfig.defaultSort = this.defaultSort();
       viewConfig.linkExportField = this.linkExportField();
+      viewConfig.rowSelectionEnabled = this.rowSelectionEnabled();
+      viewConfig.rowHoverEnabled = this.rowHoverEnabled();
+      viewConfig.heatmapCircleClickTransformFn = ({
```
hallieswan, line 143: Enable heatmap circle popovers in demo story to show that the heatmap popover opens AND the row is selected on click

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -142,6 +146,13 @@ export class ComparisonToolService<T> {
   });
 
   constructor() {
+    effect(() => {
+      if (this.isLoadingTableData()) return;
+      if (!this.viewConfig().rowSelectionEnabled) return;
+      if (this.selectedRowIdSignal() !== null) return;
+      this.autoSelectFirstRow();
+    });
```
hallieswan, line 155: Use an effect to `autoSelectFirstRow` rather than calling when setting pinned or unpinned data, so that we auto-select the same row regardless of whether pinned or unpinned data returns first

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -586,6 +602,48 @@ export class ComparisonToolService<T> {
...
+  }
+
+  setHoveredRowId(id: string | null): void {
+    if (!this.viewConfig().rowHoverEnabled) return;
+    this.hoveredRowIdSignal.set(id);
+  }
+
+  /** Call after a filtered fetch to report whether the selected row exists in the full result set;
+   * resets to first row if not */
+  notifySelectedRowValidity(isInResults: boolean): void {
```
hallieswan, line 639: We only want to remove the current selection if does not appear anywhere in the full result set, but the CT route only returns one page of data at a time. This method allows consumers to specify whether the selected row exists in the full result set. See [QTL-112](https://sagebionetworks.jira.com/browse/QTL-112) for plan to adjust the QTL CT route to return this boolean flag

libs/explorers/styles/src/lib/_variables.scss
```diff
@@ -33,6 +33,9 @@ $extra-colors: (
   shadow: #c7c5c5,
   tooltip: #63676c,
 );
+$comparison-tool-colors: (
+  row-highlight: #ebf2fa,
+);
```
hallieswan, line 38: Add new block for specifying CT colors -- allows us to add other CT specific colors here, so they're easier to override per app

libs/qtl/eqtl-comparison-tool/src/lib/eqtl-comparison-tool.component.ts
```diff
@@ -23,10 +23,77 @@ import { SidebarComponent } from './components/sidebar/sidebar.component';
...
+  //     .subscribe({
+  //       next: (response) => {
+  //         this.comparisonToolService.setUnpinnedData(response.data);
+  //         this.comparisonToolService.totalResultsCount.set(response.page.totalElements);
+  //         // NOTE: selectedRowId and notifySelectedRowValidity are specific to the
+  //         // selected row feature (rowSelectionEnabled) -- other CTs do not use them.
+  //         // The backend must check whether selectedRowId exists anywhere in the full
+  //         // filtered result set (not just the current page) and return
+  //         // selectedRowInResults: boolean in the response.
+  //         this.comparisonToolService.notifySelectedRowValidity(response.selectedRowInResults);
```
hallieswan, line 75: Demonstrates how the new `notifySelectedRowValidity` method will be used

libs/explorers/comparison-tool/src/lib/comparison-tool-table/base-table/base-table.component.html
```diff
@@ -22,11 +22,21 @@
     </ng-template>
 
     <ng-template pTemplate="body" let-rowData>
-      <tr>
+      @let rowId = rowData[viewConfig().rowIdDataKey];
```
sagely1: Might consider using `String()` something like this:

`@let rowId = String(rowData[viewConfig().rowIdDataKey])`

If rowData[rowIdDataKey] corresponds to a numeric, it could cause a equality to evaluate as false on line 31.  A numeric ID gives "123" === 123 → false.

sagely1 (changes requested):
Looks great! Just one suggestion.

libs/explorers/comparison-tool/src/lib/comparison-tool-table/base-table/base-table.component.html
hallieswan: Good call! Done in the next commit

sagely1 (approved):
LGTM

PR #4105, opened 2026-06-10
feat(explorers): conditionally hide filter button and search input in comparison tool (QTL-116, QTL-117)

hallieswan:
## Description

New QTL comparison tools will not need a filter panel or a search input, but both UI elements were always rendered regardless of config. This PR makes the Filter Results button reactive -- it hides when the currently selected config has no filters and reappears when switching to one that does, with the filter panel also auto-closing on such transitions. It also adds a `showTableSearch` option to `ComparisonToolViewConfig` so callers can opt out of rendering the search filterbox input entirely.

## Related Issue

- [QTL-116](https://sagebionetworks.jira.com/browse/QTL-116)
- [QTL-117](https://sagebionetworks.jira.com/browse/QTL-117)

## Changelog

- Hide Filter Results button when the currently selected `ComparisonToolConfig` has an empty `filters` array
- Auto-close the filter panel when switching to a config with no filters
- Fix header layout so the title stays centered when the Filter Results button is absent
- Add `showTableSearch?: boolean` to `ComparisonToolViewConfig` (default `true`) to allow hiding the search input
- Update Storybook Demo story to demonstrate reactive filter button visibility across dropdown selections
- Add unit tests for filter button visibility, filter panel auto-close, and search input visibility

## Testing

- Open the Storybook `ComparisonToolComponent` Demo story -- the Filter Results button should be visible for the default dropdown selection (configs with filters)
- In the Demo story, switch the first dropdown to the second option, then switch the second dropdown to its last option -- the Filter Results button should disappear
- Open the filter panel, then switch to a dropdown option whose config has no filters -- confirm the filter panel closes automatically
- In the Storybook controls panel, toggle `showTableSearch` to `false` -- the search input should disappear from the controls bar; toggling back to `true` should restore it
- Verify the header title remains horizontally centered regardless of whether the Filter Results button is visible

### Preview

`nx start explorers-storybook`

Filter results button is shown when current config has filters and hidden otherwise: 

https://github.com/user-attachments/assets/d5ea2102-86c2-4adf-96cf-decbc2f0c064

Table search filterbox can be optionally hidden via viewConfig option: 

https://github.com/user-attachments/assets/3b38fec3-851e-413c-86d6-83da72666113

[QTL-116]: https://sagebionetworks.jira.com/browse/QTL-116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QTL-117]: https://sagebionetworks.jira.com/browse/QTL-117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/services/src/lib/comparison-tool.service.ts
+++ b/libs/explorers/services/src/lib/comparison-tool.service.ts
@@ -643,6 +643,10 @@ export class ComparisonToolService<T> {
       pinnedItems: currentPins,
       pageNumber: this.FIRST_PAGE_NUMBER,
     });
+
+    if (!newFilters.length && this.isFilterPanelOpenSignal()) {
+      this.isFilterPanelOpenSignal.set(false);
+    }
   }
 
   private normalizeSelection(selection: string[], configs: ComparisonToolConfig[]): string[] {
--- a/libs/explorers/services/src/lib/comparison-tool.service.ts
+++ b/libs/explorers/services/src/lib/comparison-tool.service.ts
@@ -64,6 +64,7 @@ export class ComparisonToolService<T> {
     rowIdDataKey: '_id',
     allowPinnedImageDownload: true,
     linkExportField: 'link_url',
+    showTableSearch: true,
   };
 
   // Private State Signals
--- a/libs/explorers/services/src/lib/comparison-tool.service.ts
+++ b/libs/explorers/services/src/lib/comparison-tool.service.ts
@@ -143,6 +143,14 @@ export class ComparisonToolService<T> {
   });
 
   constructor() {
+    // Close the filter panel when the current config has no filters to avoid showing an empty panel.
+    effect(() => {
+      const config = this.currentConfig();
+      if (config !== null && !config.filters?.length && this.isFilterPanelOpenSignal()) {
+        this.isFilterPanelOpenSignal.set(false);
+      }
+    });
+
     // Automatically sync state changes to URL.
     // This effect re-runs whenever any of the tracked signals change,
     // keeping the URL in sync with the component's state.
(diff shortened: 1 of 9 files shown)


libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -643,6 +644,10 @@ export class ComparisonToolService<T> {
       pinnedItems: currentPins,
       pageNumber: this.FIRST_PAGE_NUMBER,
     });
+
+    if (!newFilters.length && this.isFilterPanelOpenSignal()) {
+      this.isFilterPanelOpenSignal.set(false);
+    }
```
hallieswan: Close the filter panel if open when the user changes dropdowns to a view without any filters

rrchai (approved):
LGTM

libs/explorers/services/src/lib/comparison-tool.service.ts
sagely1: Consider adding this as a comment as I think it is helpful for the reader of this code

sagely1: Also, consider adding an effect instead.  Something like:
```
constructor() {
  effect(() => {
    if (!this.currentConfig()?.filters?.length && this.isFilterPanelOpenSignal()) {
      this.isFilterPanelOpenSignal.set(false);
    }
  });
}
```

sagely1 (changes requested):
Looks great!  Just a few suggestions.

libs/explorers/services/src/lib/comparison-tool.service.ts
hallieswan: Good idea! Will do in the next commit

sagely1 (approved):
LGTM

PR #4106, opened 2026-06-12
feat(qtl): add home page background and add opacity to card components (QTL-20)

sagely1:
## Description

The QTL home page was missing the decorative background from the design. This adds the
background SVG behind the home content and wires up the responsive behavior so it reads well
across viewports. To match the design's layered look, the stat cards and the example-searches
chiclet card are made translucent so the background shows through, with the stat cards returning
to fully opaque on hover for legibility.

## Related Issue

- [QTL-20](https://sagebionetworks.jira.com/browse/QTL-20) — Add home page background

## Changelog

- Add the home page background SVG and render it behind the home content (anchored top-center, natural scale)
- Anchor the background to the top-left below the desktop breakpoint so the left edge stays visible as the viewport narrows
- Reduce the home page heading to 24px, and the chiclet card title to 16px, below the desktop breakpoint
- Make the QTL stat cards and chiclet card translucent (driven by a single shared token) so the background shows through, with stat cards returning to full opacity on hover
- Add opt-in `--stat-card-background-color` / `--chiclet-card-background-color` to the shared stat card and chiclet card, defaulting to their existing colors so Agora and Model-AD are unaffected
- Define a QTL `$desktop-breakpoint` token and point the home and footer media queries at it

## Testing

- Open the QTL home page at a desktop width and confirm the hexagon background renders behind the content, anchored to the top, and is subtle (low opacity)
- Narrow the window below 992px and confirm the background re-anchors to the left edge (right side crops off), the main heading shrinks to 24px, and the "Example searches" chiclet card title shrinks to 16px
- Confirm the stat cards and the example-searches chiclet card are translucent so the background is faintly visible through them, and that the two match
- Hover over a stat card and confirm its background becomes fully opaque while the existing scale/shadow hover animation still plays

### Preview

Transparent Chiclet Card
image

Transparent Stat Card
image

Mobile Preview
image

Desktop Preview
image

[QTL-20]: https://sagebionetworks.jira.com/browse/QTL-20?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/qtl/styles/src/lib/_variables.scss
```diff
@@ -23,3 +23,9 @@ $main-colors: (
   chiclet-blue: #3f51b5,
   chiclet-purple: #9d67d6,
 );
+
+// -------------------------------------------------------------------------- //
+// Breakpoints
+// -------------------------------------------------------------------------- //
+
+$desktop-breakpoint: 992px;
```
hallieswan: Could we map this variable to the `lg-breakpoint` value, since the values match? 

```suggestion
$desktop-breakpoint: variables.$lg-breakpoint;
```

libs/explorers/ui/src/lib/components/chiclet-card/chiclet-card.component.scss
```diff
@@ -8,14 +10,19 @@
...
   border: 1px solid var(--color-border);
-  background: rgb(255 255 255 / 80%);
+  background: var(--chiclet-card-background-color, rgb(255 255 255 / 80%));
   box-shadow: 0 6px 24px 0 rgb(0 0 0 / 5%);
 
   h2 {
     font-weight: bold;
     font-size: 24px;
     line-height: 28px;
     margin: 0 0 20px;
```
hallieswan: It looks like this style might have changed a bit in the latest [design](https://www.figma.com/design/TpCkecXfQ0RGXfVlyEKpxf/QTL-Explorer?node-id=2953-41569&m=dev) -- 

```suggestion
  h2 {
    color: var(--color-action-primary);
    font-weight: 600;
    font-size: 21px;
    line-height: 24px;
    margin: 0;
```

libs/explorers/ui/src/lib/components/chiclet-card/chiclet-card.component.scss
hallieswan: This gap also looks like it changed a bit in the latest [design](https://www.figma.com/design/TpCkecXfQ0RGXfVlyEKpxf/QTL-Explorer?node-id=2924-39967&m=dev) -- 


```suggestion
  gap: 16px;
```

libs/qtl/home/src/lib/home.component.scss
```diff
@@ -1,10 +1,32 @@
...
+}
+
 .section {
   display: flex;
   flex-direction: column;
   gap: 40px;
   position: relative;
   color: var(--color-primary);
-  padding: 80px clamp(30px, 5vw, 160px);
+  padding: 100px clamp(30px, 5vw, 160px);
```
hallieswan: The [design](https://www.figma.com/design/TpCkecXfQ0RGXfVlyEKpxf/QTL-Explorer?node-id=2953-41550&m=dev) shows 80px padding for the top/bottom of the home page -- should this change only apply to left/right of the page?

libs/explorers/ui/src/lib/components/chiclet-card/chiclet-card.component.scss
```diff
@@ -8,14 +10,19 @@
...
+
+    @media (width < #{vars.$lg-breakpoint}) {
+      font-size: 16px;
+      line-height: 20px;
+    }
```
hallieswan: What do you think about applying this change in the home page stylesheet instead? This breakpoint change feels like a concern of the home page, not of the generic component. It also depends on the `lg-breakpoint` variable, whereas other changes on the home page depend on the new `desktop-breakpoint` variable

libs/explorers/ui/src/lib/components/chiclet-card/chiclet-card.component.scss
```diff
@@ -1,3 +1,5 @@
+@use 'explorers/styles/src/variables' as vars;
+
 .chiclet-card {
   display: flex;
   padding: 40px;
```
hallieswan: It also looks like this padding decreases to 16px in mobile in the latest [design](https://www.figma.com/design/TpCkecXfQ0RGXfVlyEKpxf/QTL-Explorer?node-id=2924-39966&m=dev).

hallieswan (commented):
Looks great! Just a couple small questions. 

I also noticed that the latest chiclet card design might've drifted a bit from the current implementation -- not sure if that's something you're already tracking, but I added comments for what I noticed. Up to you whether we tackle them as part of this PR or in the future instead!

libs/qtl/styles/src/lib/_variables.scss
sagely1: good idea - we can import the explorer's variables.

libs/explorers/ui/src/lib/components/chiclet-card/chiclet-card.component.scss
sagely1: Ah, so it has.  I'll update this along with the gap I think also changed.

libs/qtl/home/src/lib/home.component.scss
sagely1: You're right.  The design is currently showing 80px (top, bottom) and 94px (left, right).

libs/explorers/ui/src/lib/components/chiclet-card/chiclet-card.component.scss
sagely1: Good idea.  I'll add some variables and a fallback.

libs/qtl/home/src/lib/home.component.scss
```diff
@@ -1,10 +1,38 @@
...
+  background-position: center top;
+  background-repeat: no-repeat;
+  background-size: auto;
+}
+
+@media (width <= #{variables.$desktop-breakpoint}) {
+  #home-page {
+    background-position: left top;
+  }
+}
```
hallieswan: Nit: should we colocate this block with the other mobile breakpoint block, so we can easily see all the mobile styles in one place? Also, it looks like this breakpoint rule is slightly different than the other block (<= vs <) -- is that by design?

libs/explorers/ui/src/lib/components/chiclet-card/chiclet-card.component.scss
hallieswan: Nit: should we update this padding in the home page mobile style block to 16px, so it matches the mobile design? We could add a similar var here that we can override in the home component, e.g. 


```suggestion
  padding: var(--chiclet-card-padding, 40px);
```

hallieswan (commented):
Looks great! Just two small questions for you

libs/explorers/ui/src/lib/components/chiclet-card/chiclet-card.component.scss
```diff
@@ -18,6 +21,10 @@
     line-height: var(--chiclet-card-title-line-height, 24px);
     margin: 0;
   }
+
+  @media (width < #{vars.$lg-breakpoint}) {
+    padding: 16px;
+  }
```
hallieswan: We update the other chiclet styles for the home page mobile design in `libs/qtl/home/src/lib/home.component.scss` -- what do you think about colocating this change there as well since it may be more of a concern of the QTL home page than the generic chiclet-card component?

hallieswan (approved):
Looks good to me!

PR #4107, opened 2026-06-12
feat(explorers): add optional subheader support to header dropdown nav menus (MG-943)

hallieswan:
## Description

The shared `explorers-header` component previously only supported flat lists of links within dropdown menus. This PR extends `NavigationLink` with an optional `isSubheader` flag, allowing dropdown children to act as non-clickable section headers that group child links beneath them. Existing dropdowns (e.g., Agora Nominations) are unaffected since the feature is opt-in.

## Related Issue

- [MG-943](https://sagebionetworks.jira.com/browse/MG-943)

## Changelog

- Add `isSubheader?: boolean` to `NavigationLink` to support non-clickable section headers inside dropdown menus
- Update desktop dropdown to render subheader groups via PrimeNG `MenuItem.items`; subheader labels and flat items render bold (700), child links under a subheader render at normal weight (400)
- Update mobile hamburger menu to render subheaders as non-clickable bold labels with an additional indent level for their child links (normal weight)
- Update `isDropdownActive()` to recurse into subheader grandchildren so the dropdown trigger highlights correctly when a grouped route is active
- Fix mobile nav font weights: top-level links and section labels bold (700), submenu child links normal (400)
- Throw a descriptive error when a dropdown mixes subheader and flat children, which is unsupported due to a PrimeNG rendering limitation
- Add `HeaderComponent` Storybook story demonstrating standalone links, a flat dropdown, and a dropdown with subheaders in a single nav config

## Testing

- In Storybook (`nx start explorers-storybook`, port 4402), open **UI / Header** and verify all three nav variants render correctly
- Click **Menu** -- all child links should be bold and navigable
- Click **Menu (subheaders)** -- group labels should be bold and non-clickable, child links should be normal weight and navigable
- Resize to mobile and open the hamburger menu -- top-level links and subheader labels should appear at one indent (bold, labels non-clickable), their child links at a second indent (normal weight, clickable)

### Preview

| view | type | design | storybook | 
:---: | :---: | :---: | :---:
desktop | no subheaders | MG-943_desktop_noSubheaders_design | image
desktop | subheaders | MG-943_desktop_subheaders_design | image
mobile | no subheaders and subheaders | MG-943_mobile_design | image

[MG-943]: https://sagebionetworks.jira.com/browse/MG-943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/styles/src/lib/components/_menu.scss
+++ b/libs/explorers/styles/src/lib/components/_menu.scss
@@ -1,0 +1,10 @@
+.header-dropdown-menu {
+  .p-menu-item-label,
+  .p-menu-submenu-label {
+    font-weight: 700;
+  }
+
+  .p-menu-submenu-label ~ .p-menu-item .p-menu-item-label {
+    font-weight: 400;
+  }
+}
--- a/libs/explorers/styles/src/lib/components/_menu.scss
+++ b/libs/explorers/styles/src/lib/components/_menu.scss
@@ -2,9 +2,10 @@
   .p-menu-item-label,
   .p-menu-submenu-label {
     font-weight: 700;
+    color: var(--color-text);
   }
 
-  .p-menu-submenu-label ~ .p-menu-item .p-menu-item-label {
+  .header-dropdown-subheader-child .p-menu-item-label {
     font-weight: 400;
   }
 }
--- a/libs/explorers/ui/src/lib/components/header/header.component.scss
+++ b/libs/explorers/ui/src/lib/components/header/header.component.scss
@@ -76,6 +76,7 @@
 
   a {
     font-size: var(--font-size-lg);
+    font-weight: 700;
     color: var(--color-text);
     text-decoration: none;
   }
@@ -122,7 +123,6 @@
     a {
       position: relative;
       display: flex;
-      font-weight: 700;
       padding: 8px 0;
       color: var(--color-text-secondary);
       transition: var(--transition-duration);
@@ -218,6 +218,7 @@
     .mobile-dropdown-label {
       display: block;
       font-size: var(--font-size-lg);
+      font-weight: 700;
       color: var(--color-text);
       padding: 10px 0;
       cursor: default;
@@ -226,6 +227,10 @@
 
     .mobile-submenu {
       margin-left: 1.25rem;
+
+      .mobile-submenu a {
+        font-weight: 500;
+      }
     }
 
     &.show {
--- a/libs/explorers/ui/src/lib/components/header/header.component.ts
+++ b/libs/explorers/ui/src/lib/components/header/header.component.ts
@@ -64,30 +64,45 @@ export class HeaderComponent implements OnInit {
 
   isDropdownActive(link: NavigationLink): boolean {
     if (!link.children) return false;
-    return link.children.some((child) =>
-      child.routerLink
-        ? this.router.isActive(child.routerLink.join('/'), {
-            paths: 'subset',
-            queryParams: 'ignored',
-            fragment: 'ignored',
-            matrixParams: 'ignored',
-          })
-        : false,
-    );
+    return link.children.some((child) => {
+      const candidates = child.isSubheader && child.children ? child.children : [child];
+      return candidates.some((c) =>
+        c.routerLink
+          ? this.router.isActive(c.routerLink.join('/'), {
+              paths: 'subset',
+              queryParams: 'ignored',
+              fragment: 'ignored',
+              matrixParams: 'ignored',
+            })
+          : false,
+      );
+    });
   }
 
   private buildDropdownMenuItems(links: NavigationLink[]) {
     this.dropdownMenuItems.clear();
     for (const link of links) {
       if (link.children) {
-        this.dropdownMenuItems.set(
-          link.label,
-          link.children.map((child) => ({
-            label: child.label,
-            routerLink: child.routerLink,
-          })),
-        );
+        this.dropdownMenuItems.set(link.label, this.toMenuItems(link.children));
       }
     }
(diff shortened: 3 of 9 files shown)


libs/explorers/styles/src/lib/components/_menu.scss
```diff
@@ -0,0 +1,11 @@
+.header-dropdown-menu {
```
hallieswan, line 1: Add `_menu.scss` to style PrimeNG's `p-menu` popup overlay, which is appended to `document.body` outside the component's view encapsulation scope -- follows same pattern as `_popover.scss`

libs/explorers/ui/src/lib/components/header/header.component.ts
```diff
@@ -64,30 +64,59 @@ export class HeaderComponent implements OnInit {
...
+  private toMenuItems(children: NavigationLink[]): MenuItem[] {
+    const hasSubheaders = children.some((c) => c.isSubheader);
+    const hasFlat = children.some((c) => !c.isSubheader);
+    // PrimeNG renders all top-level MenuItem entries as group headers when any entry has nested
+    // `items`, making flat siblings non-clickable. Mixed layouts are therefore not supported.
+    if (hasSubheaders && hasFlat) {
+      throw new Error(
+        'HeaderComponent: mixing subheader and flat children in the same dropdown is not supported. All children should be either subheaders or flat links.',
+      );
+    }
```
hallieswan: Throw an error so we catch invalid header navigation link configs at dev/test time rather than shipping a silently broken UI (non-clickable links) with a warning message

libs/explorers/ui/src/lib/components/header/header.component.scss
```diff
@@ -226,6 +227,10 @@
 
     .mobile-submenu {
       margin-left: 1.25rem;
+
+      .mobile-submenu a {
+        font-weight: 500;
```
sagely1, line 232: Shouldn't mobile subheaders be 400 to match desktop?

libs/explorers/styles/src/lib/components/_menu.scss
sagely1: This change seems to have caused the Nominations dropdowns to no longer be bold.  The design has bold menu items here: https://www.figma.com/design/dgrq8yWWywKQI3rserIR60/Agora-4.0?node-id=508-17718&m=dev

libs/explorers/ui/src/lib/components/header/header.component.scss
hallieswan: The font-weight specified in the [mobile design](https://www.figma.com/design/GtUYqZBxoOnPtjV1hLcTgw/Model-AD-Explorer-Designs?node-id=4403-4119&m=dev) is different than the font-weight specified in the [desktop design](https://www.figma.com/design/GtUYqZBxoOnPtjV1hLcTgw/Model-AD-Explorer-Designs?node-id=4775-7711&m=dev)

libs/explorers/ui/src/lib/components/header/header.component.ts
sagely1: This check only runs for desktop but not for mobile.  Could we potentially do this error checking earlier for both screen sizes?

sagely1 (changes requested):
Looks good! Just a few suggestions.

libs/explorers/styles/src/lib/components/_menu.scss
hallieswan: I'm seeing that the Nominations dropdowns are still bold when I build locally (`agora-build-images && agora-docker-start`) -- would you mind sharing a screenshot of what you're seeing in your build? I'm wondering if it's possible that older styles are cached for you?

image

libs/explorers/ui/src/lib/components/header/header.component.ts
hallieswan: Good call! Done in the next commit

libs/explorers/styles/src/lib/components/_menu.scss
```diff
@@ -0,0 +1,11 @@
...
+  .p-menu-item-label,
+  .p-menu-submenu-label {
+    font-weight: 700;
+    color: var(--color-text);
+  }
+
+  .header-dropdown-subheader-child .p-menu-item-label {
+    font-weight: 400;
```
sagely1, line 9: What do you think of using font-weight: normal instead of 400?  Another option is to remove this style if it defaults to this weight.

libs/explorers/ui/src/lib/components/header/header.component.scss
sagely1: Oh interesting.  Thanks for clarifying.

libs/explorers/styles/src/lib/components/_menu.scss
hallieswan: Good question! The style is actually need here to override the parent selector, which sets `.p-menu-item-label` to `font-weight: 700`, so without the override the subheader child labels would inherit that weight. I used `400` because numeric values are more common across explorers projects (`font-weight: normal` only appears once), but happy to switch to `normal` if you'd prefer. Might be worth a larger refactor ticket though?

sagely1: Looks like 400 is more prevalent than normal.  Let's stick with numeric values as the corresponding preset values are not the most intuitive.

sagely1 (approved):
LGTM

PR #4109, opened 2026-06-15
feat(explorers): add column_width override support for explorer CTs (AG-2092, AG-2132, AG-2133)

sagely1:
## Description

Comparison Tool columns are currently auto-sized depending on their content, which means sorting on a column with dramatically different widths (depending on cell content) will cause large shifts when sorting.  This was most dramatically on Agora's nomination CTs. This PR adds an optional `column_width` to the CT column config so a column can opt out of content-based auto-sizing and keep a constant pixel width regardless of sort order. The custom width comes from `column_width` in `ui_config`.

## Related Issue

- [AG-2092](https://sagebionetworks.jira.com/browse/AG-2092)
- [AG-2132](https://sagebionetworks.jira.com/browse/AG-2132)
- [AG-2133](https://sagebionetworks.jira.com/browse/AG-2133)

## Changelog

- Add optional `column_width` field to the `ComparisonToolConfigColumn` OpenAPI schema for both Agora and Model-AD, and regenerate the Angular/Java clients
- Map `column_width` through the document → DTO mapper and persist it on the config document
- Route columns with an explicit `column_width` to a fixed width (applied verbatim, bypassing the auto-size measurement and MIN/MAX clamp) via a new `resolveFixedColumnWidths` helper
- Add unit tests for the helper split logic and the Java config mappers
- Bump Agora `DATA_VERSION` to 85 to pick up the config release that defines the first fixed-width column

## Testing

- Open the Agora Nominated Targets comparison tool and confirm the "Nominating Teams" column renders at the configured fixed width (300px).  Sorting ascending or descending should not result in a column shift.
- Resize the browser window and confirm fixed-width columns stay at their configured width while auto-sized columns recalculate

### Preview

Before:

https://github.com/user-attachments/assets/3cac4f1c-9103-4b2a-8c2a-b25f978cf281

After:

https://github.com/user-attachments/assets/b95468b9-65e7-4ea5-a4c4-b53d559227dd

[AG-2092]: https://sagebionetworks.jira.com/browse/AG-2092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AG-2132]: https://sagebionetworks.jira.com/browse/AG-2132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AG-2133]: https://sagebionetworks.jira.com/browse/AG-2133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/model-ad/api-next/src/main/resources/openapi.yaml
```diff
@@ -478,6 +484,7 @@ components:
             is_hidden: true
             link_url: link_url
             sort_tooltip: sort_tooltip
+            column_width: 0
```
hallieswan: Nit: do you know how this example is auto-generated? Should we provide a real example value in the `ComparisonToolConfigColumn.yaml` definition for `column_width`, so the example generated is a value we'd actually expect someone to use? Just thinking that a value of 300 would make more sense than 0

libs/model-ad/api-description/src/components/schemas/ComparisonToolConfigColumn.yaml
```diff
@@ -35,6 +35,11 @@ properties:
   is_hidden:
     type: boolean
     description: Indicates whether the column is shown in the comparison table UI.
+  column_width:
+    type: integer
+    format: int32
+    description: Optional fixed width for the column in pixels. When set, the UI uses this
+      width verbatim instead of auto-sizing the column from content.
```
hallieswan, line 44: Nit: should we add a `minimum` here, so the user can't set a negative `column_width`? If we wanted to enforce a positive width (e.g. strictly greater than 0), we could also use something like this: 


```suggestion
  column_width:
    type: integer
    format: int32
    minimum: 0
    exclusiveMinimum: true # Means the value must be > 0
    description: Optional fixed width for the column in pixels. When set, the UI uses this
      width verbatim instead of auto-sizing the column from content.
```

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.helpers.ts
```diff
@@ -129,6 +129,27 @@ export function restoreCellStyles(
...
+  columnsToMeasure: ComparisonToolColumn[];
+} {
+  const fixedWidthByColumn: Record<string, string> = {};
+  const columnsToMeasure: ComparisonToolColumn[] = [];
+  for (const column of columns) {
+    if (column.column_width != null) {
+      fixedWidthByColumn[column.data_key] = `${column.column_width}px`;
+    } else {
+      columnsToMeasure.push(column);
+    }
```
hallieswan, line 150: Should we check that the column width is positive here?

hallieswan (commented):
This is looking great! Just a couple small questions for you

apps/model-ad/api-next/src/main/resources/openapi.yaml
sagely1: Agreed.  Since 0 is not an ideal value for a column, I'll update the examples.

libs/model-ad/api-description/src/components/schemas/ComparisonToolConfigColumn.yaml
sagely1: Yes, for documentation purposes we would probably want to specify: `minimum: 1`

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.helpers.ts
sagely1: Good idea.  I will also add a Sentry warning if it is invalid.

hallieswan (approved):
Looks great!

PR #4116, opened 2026-06-29
feat(agora): add visualization overview content for nominated comparison tools (AG-1981)

sagely1:
## Description

Every comparison tool shows a "Visualization Overview" tutorial, but in Agora that content was provided once at the app level, so the legacy Gene Comparison Tool (GCT) and the new Nominated Targets/Drugs CTs all showed the GCT's tutorial. This change introduces a per-CT override so the Nominated CTs get their own tutorial (modeled on Model-AD's, with the heatmap-specific panel omitted since these tools have no heatmap) while the GCT keeps its existing walkthrough. It also unifies where each product's overview content lives, so the three explorer apps follow one convention.

## Related Issue

- [AG-1981](https://sagebionetworks.jira.com/browse/AG-1981) — Create visualization overview content

## Changelog

- Add an optional `visualizationOverviewPanes` override to `ComparisonToolViewConfig`; the overview panel now prefers a CT's own panes and falls back to the app-level `EXPLORERS_CONFIG` when none are set
- Add shared visualization overview content for the Nominated Targets and Nominated Drugs CTs (3 panels, text from Model-AD, heatmap panel omitted) and wire it into both CTs
- Keep the GCT's existing tutorial via the app-level fallback
- Standardize the content location: move each product's overview content into `libs/<product>/config` (Agora, Model-AD, QTL) and export it from the config barrel
- Add nomination overview assets (`nomination-overview-1.png`, `nomination-overview-2.mp4`, `nomination-overview-3.mp4`)
- Add unit tests for override-vs-fallback precedence and for the Nominated CT content wiring

## Testing

- Open the **Nominated Targets** CT (`/comparison/targets`) → click **Visualization Overview** at the bottom → confirm 3 panels ("Comparison Tool Tutorial", "Search & Filter", "Build, share and export custom comparisons"), with an image on panel 1 and videos on panels 2–3, and working Back/Next/Close navigation
- Repeat for the **Nominated Drugs** CT (`/comparison/drugs`) → confirm the same shared content renders
- Open the **GCT** (`/genes/comparison`) → click **Visualization Overview** → confirm its original tutorial is unchanged (not the new Nominated content)
- Confirm **Model-AD** overviews still render exactly as before (content only moved location)
- Toggle **"Don't show this again"**, reload, and confirm the overview stays hidden (single app-wide preference is unchanged)

### Preview

This is pane 1 of the Nominated Drugs and Nominated Targets CTs.  Before GCT content was the same across all CTs.  Now the newer CTs have unique assets (images and video across 3 panes).

| Before                                                                                          | After                                                                                                  |
| ----------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------ |
| image | image |

[AG-1981]: https://sagebionetworks.jira.com/browse/AG-1981?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/agora/config/src/lib/nominated-cts-visualization-overview.content.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Visualization overview content shared by Agora's Nominated Targets and Nominated Drugs
+ * comparison tools. The text is copied verbatim from Model-AD's tutorial (the heatmap-specific
+ * "Customize your view" panel is intentionally omitted since these CTs have no heatmap).
+ *
+ * TODO: Two references describe features disabled for the Nominated CTs and should be revised
+ * after stakeholder review:
+ *   - Panel 1 mentions the Legend link, but these CTs set `legendEnabled: false` (no Legend link).
+ *   - Panel 3 mentions "download an image of the pinned results", but these CTs set
+ *     `allowPinnedImageDownload: false` (no image download control).
```
hallieswan: Could we include the id for the ticket that tracks the follow-up work described in this comment?

libs/agora/config/src/lib/nominated-cts-visualization-overview.content.ts
```diff
@@ -0,0 +1,33 @@
+import { VisualizationOverviewPane } from '@sagebionetworks/explorers/models';
+
+/**
+ * Visualization overview content shared by Agora's Nominated Targets and Nominated Drugs
+ * comparison tools. The text is copied verbatim from Model-AD's tutorial (the heatmap-specific
+ * "Customize your view" panel is intentionally omitted since these CTs have no heatmap).
```
hallieswan: Do we need this context in the comment? The visualization overview copy may drift between Model AD and Agora in the future, then this comment would become stale

hallieswan (commented):

libs/agora/assets/images/nomination-overview-1.png
hallieswan, line 1: Should this image be updated to one that doesn't include a legend link, since we don't mention the legend in the visualization overview anymore?

libs/agora/config/src/lib/nominated-cts-visualization-overview.content.ts
```diff
@@ -0,0 +1,21 @@
...
+  },
+  {
+    heading: 'Search & Filter',
+    content: `<p>Quickly find results of interest using the tool’s embedded search box. Enter a value to get a list of matching results, or perform a bulk search using a list of comma-separated full values.</p>
+      <p>You can also use the tool’s supported filters to find results of interest. The filter region near the top of the page shows which filters are currently applied. You can clear specific filters, or all of the filters at once.</p>
+      <div class="video-container"><video autoplay muted loop playsinline preload="none" controls><source src="agora-assets/videos/nomination-overview-2.mp4" type="video/mp4"></video></div>`,
+  },
+  {
+    heading: 'Build, share and export custom comparisons',
+    content: `<p>Pin up to 50 rows of results so that you can easily compare them. Once you've built your custom result set, you can share it with your colleagues by Share URL, download an image of the pinned results, and export the underlying data for further analysis.</p>
```
hallieswan: Should we update this sentence since the nominated targets/drugs CTs don't allow downloading an image of the pinned results?


```suggestion
    content: `<p>Pin up to 50 rows of results so that you can easily compare them. Once you've built your custom result set, you can share it with your colleagues by Share URL, and export the underlying data for further analysis.</p>
```

JessterB: Good catch!

libs/agora/assets/images/nomination-overview-1.png
sagely1: That's right.  Updating screenshot to a CT without a Legend in the next commit.

libs/agora/config/src/lib/nominated-cts-visualization-overview.content.ts
sagely1: Fixing in the next commit!

hallieswan (approved):
Looks good to me!

PR #4118, opened 2026-06-30
feat(explorers): keep comparison tool controls visible at narrow widths (AG-2142)

sagely1:
## Description

In the comparison tool, both the table controls (column picker, significance, and category selectors) and the pin-header controls (Download, Clear All, Pin All) are right-aligned within elements that stretch to the full width of the wide heatmap table. Once the viewport is narrower than the table (around 1600px), those controls scroll off the right edge and can only be reached with the horizontal scroll bar. This change makes the controls horizontally sticky so they float with the scroll and stay reachable at narrow widths, until there is genuinely no room left for them.

## Related Issue

- [AG-2142](https://sagebionetworks.jira.com/browse/AG-2142) — CT: Make Column picker & Pin header controls horizontally sticky

## Changelog

- Make the comparison tool table controls (column picker, significance, category selectors) horizontally sticky so they stay visible as the window narrows
- Make the pin-header controls (Download, Clear All, Pin All) horizontally sticky for the same reason
- Move controls down 20px according to original Figma design

## Testing

- Open the comparison tool (e.g. Agora gene comparison) in a wide window (>1600px) and confirm the controls look and behave exactly as before.
- Narrow the browser window below ~1600px (or zoom in) and confirm the column picker, significance, and category selector controls remain visible at the right edge instead of disappearing off-screen.
- Pin at least one row to reveal the "Pinned Results" header, then narrow the window and confirm the Download and Clear All buttons stay visible.
- Apply a search term or filter to reveal the "Matching/Filtered Results" header and confirm the Pin All button stays visible at narrow widths.
- Scroll the table horizontally to its far right and confirm the floating controls settle back into their resting position without a visible jump.

## Agora
| Before                                                                                              | After                                                                                            |
| --------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------ |
| image | image |

## Model-AD
Desktop Width (controls at the far right)
image

Small width (controls wrap around dropdowns, pinned controls visible)
image

[AG-2142]: https://sagebionetworks.jira.com/browse/AG-2142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/explorers/comparison-tool/src/lib/comparison-tool-controls/comparison-tool-controls.component.scss
```diff
@@ -23,13 +21,25 @@ $border-style: 1px solid var(--color-gray-300);
...
+    explorers-comparison-tool-category-selectors {
+      flex: 1 1 auto;
+    }
+
+    .controls-right-inner {
+      display: flex;
+      flex-direction: row;
+      flex-shrink: 0;
+      align-items: start;
+      margin-top: 20px;
```
hallieswan, line 39: This margin is applied on narrow and wide screens, so these controls are pushed down relative to the current design on wide screens. What do you think about using `row-gap` on the parent container (`.controls-right`) instead, so the extra space is only added when the rows are wrapped?

[Current](https://sage-bionetworks.github.io/sage-monorepo/storybook/explorers/iframe.html?globals=&id=comparison-tool-comparisontoolcomponent--demo&viewMode=story): 
Image

This PR: 
Image

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.scss
```diff
@@ -25,6 +25,11 @@
           font-weight: 700;
           flex-grow: 1;
         }
+
+        &:not(:first-child) {
+          position: sticky;
+          right: 30px;
```
hallieswan: Is this 30px sticky offset linked to the `margin-right` set on `.table-divider-inner`? If so, what do you think about creating a variable to tie them together, so they don't drift in the future?

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.scss
```diff
@@ -25,6 +25,11 @@
           font-weight: 700;
           flex-grow: 1;
         }
+
+        &:not(:first-child) {
```
hallieswan: What do you think about using a more specific selector here, so we don't inadvertently target other elements added to this div in the future? If we added the same class to div that wraps the "Pin All" button and the div that wraps the "Download" and "Clear All" button, then we could use that instead. What do you think?

hallieswan (changes requested):
This is such a great improvement, especially for CTs that include the sidebar! Just a couple small suggestions

libs/explorers/comparison-tool/src/lib/comparison-tool-controls/comparison-tool-controls.component.scss
sagely1: This was a visual bug to match the design: see [Figma](https://www.figma.com/design/GtUYqZBxoOnPtjV1hLcTgw/Model-AD-Explorer-Designs?node-id=371-8805&m=dev)

hallieswan: Oh good catch! Nice fix

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.scss
sagely1: Good idea.  We can probably add more class names in general to CT components for clarity.  Fixed in next commit.

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.scss
sagely1: Good idea.  Fixed in next commit.

hallieswan (approved):
Looks good to me!

PR #4125, opened 2026-07-01
feat(agora): add new badge handling logic for drug details (AG-2082)

sagely1:
## Description

[AG-2074](https://sagebionetworks.jira.com/browse/AG-2074) converted the drug `combined_with` field from a flat single-value pair into an array of `{ common_name, chembl_id }` objects so the data can represent combination therapies of three or more partner drugs. This PR is the frontend follow-up: it updates the drug details hero badges to render each combination-therapy nomination as a single badge with human-readable copy and places each badge on its own line.

## Related Issue

- [AG-2082](https://sagebionetworks.jira.com/browse/AG-2082) — FE: drug details hero - update badges to handle multiple drug combined_with entries
  - Related: [AG-2074](https://sagebionetworks.jira.com/browse/AG-2074) — ADT: convert drug_info.combined_with fields to object (data model change this builds on)

## Changelog

- Aggregate each combination-therapy nomination into a single "Nominated Combination Therapy with …" badge listing all of its partner drugs
- Format the partner list as human-readable copy: `A`, `A & B`, and `A, B & C` (comma-separated with an ampersand before the last, no Oxford comma)
- Render each badge on its own line instead of comma-separating them inline
- Guard the drug details hero and nomination-details components against a missing/undefined `combined_with`, treating it as a solo nomination
- Add Storybook variants (`SingleDrugCombination`, `MultiDrugCombinations`) and unit tests covering the new formatting and the missing-field case

## Testing

- Open a drug that was nominated as a standalone therapy and confirm a **Nominated Drug** badge shows.
- Confirm that when a drug has both a solo and a combination nomination, the two badges render on separate lines.
- Testing more combinations is not possible given that multiple drug partners are not in the data release today.

### Preview

| Before | After |
| ------ | ----- |
| image | image

Storybook showcasing multiple drug partners:
image

[AG-2074]: https://sagebionetworks.jira.com/browse/AG-2074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AG-2082]: https://sagebionetworks.jira.com/browse/AG-2082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AG-2074]: https://sagebionetworks.jira.com/browse/AG-2074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.html
+++ b/libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.html
@@ -2,13 +2,18 @@
   <section>
     <h1>{{ drug().common_name }}</h1>
     <p>
-      @for (badge of badges(); track badge.label; let last = $last) {
-        @if (badge.link) {
+      @for (badge of badges(); track $index; let last = $last) {
+        @if (badge.partners) {
           {{ badge.label }}
-          <a [routerLink]="badge.link">{{ badge.linkText }}</a
-          >{{ last ? '' : ',' }}
+          @for (partner of badge.partners; track partner.link) {
+            <a [routerLink]="partner.link">{{ partner.linkText }}</a
+            >{{ partner.separator }}
+          }
         } @else {
-          {{ badge.label }}{{ last ? '' : ',' }}
+          {{ badge.label }}
+        }
+        @if (!last) {
+          <br />
         }
       }
     </p>
--- a/libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.html
+++ b/libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.html
@@ -3,14 +3,12 @@
     <h1>{{ drug().common_name }}</h1>
     <p>
       @for (badge of badges(); track $index; let last = $last) {
+        {{ badge.label }}
         @if (badge.partners) {
-          {{ badge.label }}
           @for (partner of badge.partners; track partner.link) {
             <a [routerLink]="partner.link">{{ partner.linkText }}</a
             >{{ partner.separator }}
           }
-        } @else {
-          {{ badge.label }}
         }
         @if (!last) {
           <br />
--- a/libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.spec.ts
+++ b/libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.spec.ts
@@ -11,6 +11,9 @@ async function setup(drugOverrides: Partial<Drug> = {}) {
   });
 }
 
+const normalizeWhitespace = (text: string | null | undefined) =>
+  (text ?? '').replace(/\s+/g, ' ').trim();
+
 describe('DrugDetailsHeroComponent', () => {
   it('should display drug name', async () => {
     await setup();
@@ -43,7 +46,24 @@ describe('DrugDetailsHeroComponent', () => {
       expect(screen.getByText(/Nominated Combination Therapy with/)).toBeInTheDocument();
     });
 
-    it('should display a combination badge per constituent for multi-drug combinations', async () => {
+    it('should separate multiple badges with a line break', async () => {
+      await setup();
+      const badgeParagraph = screen.getByText(/Nominated Drug/).closest('p');
+      expect(badgeParagraph?.querySelector('br')).toBeInTheDocument();
+    });
+
+    it('should treat a nomination with a missing combined_with as a solo nomination', async () => {
+      await setup({
+        drug_nominations: [
+          // combined_with omitted at runtime despite the generated model typing it as required
+          { ...drugMock.drug_nominations[0], combined_with: undefined as never },
+        ],
+      });
+      expect(screen.getByText(/Nominated Drug/)).toBeInTheDocument();
+      expect(screen.queryByText(/Nominated Combination Therapy with/)).not.toBeInTheDocument();
+    });
+
+    it('should combine two partners into a single badge with an ampersand', async () => {
       await setup({
         drug_nominations: [
           {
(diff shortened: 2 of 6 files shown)


libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.stories.ts
```diff
@@ -22,3 +22,29 @@ export const DrugDetailsHero: Story = {
...
+export const MultiDrugCombinations: Story = {
+  args: {
+    drug: {
+      ...drugMock,
+      drug_nominations: [
+        drugMock.drug_nominations[0],
+        {
+          ...drugMock.drug_nominations[1],
+          combined_with: [
+            { common_name: 'Letrozole', chembl_id: 'CHEMBL1' },
```
hallieswan: What do you think about using a different drug name in both of the `combined_with` blocks, so it's clear that the main drug (Letrozole) is not included in the drug badge text?

libs/agora/drug-details/src/lib/components/drug-details-nomination-details/drug-details-nomination-details.component.spec.ts
```diff
@@ -41,6 +41,16 @@ describe('DrugDetailsNominationDetailsComponent', () => {
     expect(screen.getByText(`Why was ${drugMock.common_name} nominated?`)).toBeInTheDocument();
   });
 
+  it('should display nomination label without combined drug when combined_with is missing', async () => {
+    await setup({
+      drug_nominations: [
+        // combined_with omitted at runtime despite the generated model typing it as required
+        { ...drugMock.drug_nominations[0], combined_with: undefined as never },
```
hallieswan: Is this a real use case that we need to handle -- are there drugs where `combined_with` is missing? If so, should we update the API model so the type is accurate?

libs/agora/drug-details/src/lib/components/drug-details-nomination-details/drug-details-nomination-details.component.ts
```diff
@@ -12,7 +12,7 @@ export class DrugDetailsNominationDetailsComponent {
   drug = input.required<Drug>();
 
   getNominationLabel(nomination: NominatedDrugEvidence): string {
-    const partners = nomination.combined_with.map(
+    const partners = (nomination.combined_with ?? []).map(
```
hallieswan: Is this guard necessary? The `NominatedDrugEvidence` model defines `combined_with` as required and not-nullable. As far as I can tell, the [source data](https://www.synapse.org/Synapse:syn73880976) only has empty arrays for `combined_with` or arrays with `CombinedWith` objects, so the model matches the data

libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.spec.ts
```diff
@@ -55,11 +75,48 @@ describe('DrugDetailsHeroComponent', () => {
...
       expect(donepezilLink).toHaveAttribute('href', '/drugs/CHEMBL502');
     });
+
+    it('should combine three or more partners with commas and a trailing ampersand', async () => {
+      await setup({
+        drug_nominations: [
+          {
+            ...drugMock.drug_nominations[1],
+            combined_with: [
+              { common_name: 'Letrozole', chembl_id: 'CHEMBL1' },
```
hallieswan: Can we use a different drug name here, so the `combined_with.common_name` doesn't match the drug's `common_name`? Right now the test can't distinguish between using the `combined_with` values vs the base drug name in the drug badge

libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.html
```diff
@@ -2,13 +2,18 @@
...
           {{ badge.label }}
-          <a [routerLink]="badge.link">{{ badge.linkText }}</a
-          >{{ last ? '' : ',' }}
+          @for (partner of badge.partners; track partner.link) {
+            <a [routerLink]="partner.link">{{ partner.linkText }}</a
+            >{{ partner.separator }}
+          }
         } @else {
-          {{ badge.label }}{{ last ? '' : ',' }}
+          {{ badge.label }}
```
hallieswan: Could we hoist `{{ badge.label }}` outside the `@if` block? It looks like we display the same `badge.label` in the if/else blocks -- we're only conditionally displaying the partner badges below it, so could just use an `@if` for that

libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.spec.ts
```diff
@@ -43,7 +46,24 @@ describe('DrugDetailsHeroComponent', () => {
...
+    it('should treat a nomination with a missing combined_with as a solo nomination', async () => {
+      await setup({
+        drug_nominations: [
+          // combined_with omitted at runtime despite the generated model typing it as required
+          { ...drugMock.drug_nominations[0], combined_with: undefined as never },
+        ],
+      });
+      expect(screen.getByText(/Nominated Drug/)).toBeInTheDocument();
+      expect(screen.queryByText(/Nominated Combination Therapy with/)).not.toBeInTheDocument();
+    });
```
hallieswan: Same question here as in [this comment](https://github.com/Sage-Bionetworks/sage-monorepo/pull/4125/changes#r3514691493)

libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.spec.ts
```diff
@@ -55,11 +75,48 @@ describe('DrugDetailsHeroComponent', () => {
           },
         ],
       });
+      const badge = screen.getByText(/Nominated Combination Therapy with/);
+      expect(normalizeWhitespace(badge.textContent)).toBe(
+        'Nominated Combination Therapy with Irinotecan & Donepezil',
+      );
```
hallieswan: I think we can use [`toHaveTextContent`](https://github.com/testing-library/jest-dom#tohavetextcontent) here instead, which normalizes whitespace by default. That would allow us to remove the custom `normalizeWhitespace` helper too:

```suggestion
      const badge = screen.getByText(/Nominated Combination Therapy with/);
      expect(badge).toHaveTextContent(
        'Nominated Combination Therapy with Irinotecan & Donepezil',
      );
```

hallieswan (commented):
Looking good! Just a few questions for you

libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.html
sagely1: Good idea.  Fixed in next commit.

libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.spec.ts
sagely1: Done in next commit.

libs/agora/drug-details/src/lib/components/drug-details-nomination-details/drug-details-nomination-details.component.ts
sagely1: You're right.  No need to guard for this condition.  Removing in next commit.

libs/agora/drug-details/src/lib/components/drug-details-nomination-details/drug-details-nomination-details.component.spec.ts
sagely1: No, we can remove this.  Fixed in next commit.

libs/agora/drug-details/src/lib/components/drug-details-hero/drug-details-hero.component.stories.ts
sagely1: Good idea.  Fixed in next commit.

hallieswan (approved):
Looks good to me!

PR #4126, opened 2026-07-01
feat(agora): add support for max clinical trial phase filter (AG-2156)

sagely1:
## Description

The Agora mv117 data release adds a **Max Clinical Trial Phase** filterset to the Nominated Drugs Comparison Tool.  This PR wires the filter through end to end so users can narrow nominated drugs by clinical trial phase.

## Related Issue

- [AG-2156](https://sagebionetworks.jira.com/browse/AG-2156) — Add `maximum_clinical_trial_phase` filter to backend and wire up in frontend
  - Related: [AG-2109](https://sagebionetworks.jira.com/browse/AG-2109) — ui_config: add clinical phase filterset (deployed the filterset definition to Synapse)
  - Parent context: [AG-2076](https://sagebionetworks.jira.com/browse/AG-2076) — Update Agora to support changes in mv117

## Changelog

- Add `maximumClinicalTrialPhase` query parameter to the Nominated Drug search OpenAPI schema and regenerate the Java DTO and Angular client
- Include `maximumClinicalTrialPhase` in the `NominatedDrugService` `@Cacheable` key so results are not served from a cache entry that ignores the phase selection
- Map the `phase` filterset selection to the `maximumClinicalTrialPhase` query field in the Nominated Drugs CT component
- Extend backend repository/service tests and add a frontend test covering the new filter

## Testing

- Open the Nominated Drugs Comparison Tool and confirm the **Max Clinical Trial Phase** filter (short name "Phase") appears in the filter panel with options Preclinical, Phase I–IV, and Unknown.
- Select a single phase (e.g. Preclinical) and confirm the result set narrows to only drugs with that phase and the URL gains `?phase=Preclinical`.
- Select multiple phases and confirm results include drugs matching any selected phase.
- Switch between different phase selections in succession and confirm results update each time (guards against the cache-key regression where a stale cached page was returned for every phase).
- Call `GET /v1/comparison-tools/drugs?maximumClinicalTrialPhase=Preclinical&sortFields=total_nominations&sortOrders=-1` and confirm `page.totalElements` and the returned rows reflect only Preclinical drugs; an unknown value returns zero rows.

### Preview

Screenshot showing new filter enabled and correct filter being applied.
image

[AG-2156]: https://sagebionetworks.jira.com/browse/AG-2156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AG-2109]: https://sagebionetworks.jira.com/browse/AG-2109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AG-2076]: https://sagebionetworks.jira.com/browse/AG-2076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/agora/app/e2e/helpers/comparison-tool.ts
+++ b/apps/agora/app/e2e/helpers/comparison-tool.ts
@@ -1,13 +1,20 @@
 import { Page, expect } from '@playwright/test';
-import { ComparisonToolConfig } from '@sagebionetworks/agora/api-client';
+import {
+  ComparisonToolConfig,
+  ComparisonToolConfigFilter,
+} from '@sagebionetworks/agora/api-client';
 import { expectComparisonToolTableLoaded } from '@sagebionetworks/explorers/testing/e2e';
 import { baseURL } from '../../playwright.config';
 import {
   COMPARISON_TOOL_API_PATHS,
   COMPARISON_TOOL_CONFIG_PATH,
+  COMPARISON_TOOL_DEFAULT_PAGE_SIZE,
   COMPARISON_TOOL_DEFAULT_SORTS,
   COMPARISON_TOOL_PATHS,
 } from './constants';
+
+const snakeToCamel = (value: string): string =>
+  value.replace(/_([a-z])/g, (_match, char: string) => char.toUpperCase());
 
 export const navigateToComparison = async (
   page: Page,
@@ -68,4 +75,53 @@ export const fetchComparisonToolConfig = async (
   expect(response.ok()).toBeTruthy();
   const data = (await response.json()) as ComparisonToolConfig[];
   return data;
+};
+
+const fetchFilteredPageCount = async (
+  page: Page,
+  name: string,
+  filter: ComparisonToolConfigFilter,
+  value: string,
+): Promise<number> => {
+  const params = new URLSearchParams();
+  params.append('itemFilterType', 'exclude');
+  params.append('pageSize', COMPARISON_TOOL_DEFAULT_PAGE_SIZE.toString());
+
+  // sortFields and sortOrders are required by the API
+  const defaultSort = COMPARISON_TOOL_DEFAULT_SORTS[name];
+  for (const sort of defaultSort) {
+    params.append('sortFields', sort.field);
+    params.append('sortOrders', sort.order.toString());
+  }
+
+  // The config's query_param_key targets the frontend route; the API filters on the
+  // camelCased data_key (e.g. maximum_clinical_trial_phase -> maximumClinicalTrialPhase).
+  params.append(snakeToCamel(filter.data_key), value);
+
+  const response = await page.request.get(`${baseURL}/api/v1/${COMPARISON_TOOL_API_PATHS[name]}`, {
+    params,
+  });
+  expect(response.ok()).toBeTruthy();
+  const data = (await response.json()) as { page: { totalPages: number } };
+  return data.page.totalPages;
+};
+
+/**
+ * Finds the first (filter, value) pair whose applied result still spans more than one
+ * page. Pagination-reset tests need such a pair; some filters (e.g. Max Clinical Trial
+ * Phase) have no single value large enough to paginate.
+ */
+export const findFilterValueSpanningMultiplePages = async (
+  page: Page,
+  name: string,
+  filters: ComparisonToolConfigFilter[],
+): Promise<{ name: string; value: string }> => {
+  for (const filter of filters) {
+    for (const value of filter.values) {
+      if ((await fetchFilteredPageCount(page, name, filter, value)) > 1) {
+        return { name: filter.name, value };
+      }
+    }
+  }
+  throw new Error(`No filter value spans multiple pages for "${name}"`);
 };
--- a/apps/agora/app/e2e/helpers/comparison-tool.ts
+++ b/apps/agora/app/e2e/helpers/comparison-tool.ts
@@ -3,18 +3,16 @@ import {
   ComparisonToolConfig,
   ComparisonToolConfigFilter,
 } from '@sagebionetworks/agora/api-client';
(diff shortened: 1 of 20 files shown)


apps/agora/app/e2e/helpers/constants.ts
```diff
@@ -34,6 +34,8 @@ export const COMPARISON_TOOL_API_PATHS: Record<string, string> = {
 
 export const COMPARISON_TOOL_CONFIG_PATH = 'comparison-tools/config';
 
+export const COMPARISON_TOOL_DEFAULT_PAGE_SIZE = 10;
```
hallieswan: This matches the default explorers value in `@sagebionetworks/explorers/services` -- can we import that value directly instead?

apps/agora/app/e2e/helpers/comparison-tool.ts
```diff
@@ -69,3 +76,52 @@ export const fetchComparisonToolConfig = async (
...
+  const params = new URLSearchParams();
+  params.append('itemFilterType', 'exclude');
+  params.append('pageSize', COMPARISON_TOOL_DEFAULT_PAGE_SIZE.toString());
+
+  // sortFields and sortOrders are required by the API
+  const defaultSort = COMPARISON_TOOL_DEFAULT_SORTS[name];
+  for (const sort of defaultSort) {
+    params.append('sortFields', sort.field);
+    params.append('sortOrders', sort.order.toString());
+  }
```
hallieswan: This is identical to the sort params built in `fetchComparisonToolData` -- what do you think about pulling into a small helper method? Could potentially be moved into the explorers-testing too, if it would be useful in model-ad and qtl

apps/agora/app/e2e/helpers/comparison-tool.ts
```diff
@@ -1,14 +1,21 @@
...
 import {
   COMPARISON_TOOL_API_PATHS,
   COMPARISON_TOOL_CONFIG_PATH,
+  COMPARISON_TOOL_DEFAULT_PAGE_SIZE,
   COMPARISON_TOOL_DEFAULT_SORTS,
   COMPARISON_TOOL_PATHS,
 } from './constants';
 
+const snakeToCamel = (value: string): string =>
+  value.replace(/_([a-z])/g, (_match, char: string) => char.toUpperCase());
```
hallieswan: What do you think about using [`camelCase`](https://lodash.com/docs/#camelCase) from lodash instead? I think the hand-rolled method misses a couple edge cases, such as digit after an underscore (`phase_1` -> `phase_1` instead of `phase1`) and multiple underscores (`foo__bar` -> `foo_Bar` instead of `fooBar`)

apps/agora/app/e2e/helpers/comparison-tool.ts
```diff
@@ -69,3 +76,52 @@ export const fetchComparisonToolConfig = async (
   const data = (await response.json()) as ComparisonToolConfig[];
   return data;
 };
+
+const fetchFilteredPageCount = async (
+  page: Page,
+  name: string,
+  filter: ComparisonToolConfigFilter,
+  value: string,
+): Promise<number> => {
```
hallieswan, line 90: This is pretty similar to `fetchComparisonToolData` -- could that method be extended to append additional query params, then `fetchFilteredPageCount` could be simplified to just call `fetchComparisonToolData` with the extra page size and filter query params, and return the total pages from the response? e.g. maybe something along these lines? 

```
const fetchFilteredPageCount = async (...): Promise<number> => {
    const extra = new URLSearchParams();
    extra.append('pageSize', COMPARISON_TOOL_DEFAULT_PAGE_SIZE.toString());
    extra.append(camelCase(filter.data_key), value);
    const data = await fetchComparisonToolData<{ page: { totalPages: number } }>(page, name, [], extra);
    return data.page.totalPages;
  };
```

libs/agora/nominated-drugs-comparison-tool/src/lib/nominated-drugs-comparison-tool.component.spec.ts
```diff
@@ -50,9 +54,20 @@ async function setup() {
...
 
+const emptyQuery: ComparisonToolQuery = {
+  categories: [],
+  pinnedItems: [],
+  pageNumber: 0,
+  pageSize: 100,
+  multiSortMeta: [],
+  searchTerm: null,
+  filters: [],
+};
```
hallieswan: Nit: could consider moving this into a mock in explorers-testing, so it can be reused by other CTs/products if needed

hallieswan (commented):
Looks good! Just a couple questions

apps/agora/app/e2e/helpers/constants.ts
sagely1: No, not directly because it would break e2e.  We could use the Angular-free constants in comparison-tool.ts from lib/explorers/constants to import to the e2e suite instead.  Moving in next commit!

apps/agora/app/e2e/helpers/comparison-tool.ts
sagely1: I'll add a parameter to fetchFilteredPageCount() and that should make it easier for future tests.

apps/agora/app/e2e/helpers/comparison-tool.ts
sagely1: Good call.  Fixed in next commit.

hallieswan (approved):
Looks good to me!

PR #4128, opened 2026-07-02
fix(explorers): preserve horizontal scroll position when sorting comparison tool (AG-2160)

sagely1:
## Description

In the comparison tool (Nominated Targets, Nominated Drugs, and the Model-AD tables), scrolling the table horizontally and then clicking a column sort button snapped the table back to the far left, forcing users to re-scroll after every sort. Sorting triggers a column-width remeasurement that briefly pulls cells out of flow, collapsing the table so the browser clamps the scroll container's `scrollLeft` to zero. This change captures and restores the horizontal scroll position around that remeasurement so the user's view stays put when they interact with the table.

## Related Issue

- [AG-2160](https://sagebionetworks.jira.com/browse/AG-2160)

## Changelog

- Preserve the comparison tool's horizontal scroll position across the column-width recalculation that runs on sort and data changes

## Testing

- Open the Nominated Targets comparison tool in Agora and scroll the table fully to the right
- Click a sort button on a right-side column and confirm the horizontal scroll position is preserved (no jump to the left) and the sort applies correctly
- Repeat with pinned rows present and with an active search/filter to confirm both the pinned and unpinned tables behave
- Verify the same behavior on a Model-AD comparison tool table to confirm the shared fix works cross-product
- Resize the browser window while scrolled horizontally and confirm the scroll position is not disrupted

### Preview

Before

https://github.com/user-attachments/assets/7d028737-cd27-47fb-80ed-0aa6538d58a2 

After

https://github.com/user-attachments/assets/83bfa72a-e35c-4776-b140-5015f078ec77

[AG-2160]: https://sagebionetworks.jira.com/browse/AG-2160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
+++ b/libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
@@ -95,7 +95,17 @@ export class ComparisonToolTableComponent implements AfterViewInit {
       // so wait for fonts to be ready before calculating column widths
       // to prevent incorrect measurements
       document.fonts.ready.then(() => {
+        // Measurement briefly pulls cells out of flow, collapsing the table so the
+        // browser clamps the scroll container's scrollLeft. Save and restore it so
+        // sorting/data changes don't reset the user's horizontal scroll position.
+        const scrollContainer = this.tableElement()?.nativeElement.closest(
+          '.comparison-tool-body',
+        ) as HTMLElement | null;
+        const savedScrollLeft = scrollContainer?.scrollLeft ?? 0;
         this.columnWidths.set(this.calculateNonPrimaryColumnWidths());
+        if (scrollContainer) {
+          scrollContainer.scrollLeft = savedScrollLeft;
+        }
       });
     }
   }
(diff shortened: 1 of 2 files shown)


libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.ts
```diff
@@ -95,7 +95,17 @@ export class ComparisonToolTableComponent implements AfterViewInit {
       // so wait for fonts to be ready before calculating column widths
       // to prevent incorrect measurements
       document.fonts.ready.then(() => {
+        // Measurement briefly pulls cells out of flow, collapsing the table so the
+        // browser clamps the scroll container's scrollLeft. Save and restore it so
+        // sorting/data changes don't reset the user's horizontal scroll position.
+        const scrollContainer = this.tableElement()?.nativeElement.closest(
+          '.comparison-tool-body',
```
hallieswan: Would it be possible to pull this into constant so it doesn't drift from the template? e.g. something similar to what was done in [comparison-tool-table.constants.ts](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.constants.ts#L11-L14)

hallieswan (commented):
Looks great! Just one small suggestion

rrchai (approved):
lgtm

hallieswan (approved):

PR #4131, opened 2026-07-08
feat(explorers): add reusable explorer toggle card (MG-933)

sagely1:
## Description

MODEL-AD is expanding beyond mouse models to include marmoset models, so the home page needs a way to switch between species-specific tile sets. This PR adds a generic, reusable `ToggleCardComponent` that lets a user switch between two views, and wires it into the Model-AD home page to toggle between the mouse tiles and marmoset tiles. The component lives in `explorers/ui` so any explorer product (Agora, QTL, …) can reuse it. Marmoset content ships as placeholders in this PR; the real marmoset tiles are tracked in MG-932.

Add a toggle card component that encapsulates a PrimeNG select button that can choose between two different templates.  Each template will hold mouse or marmoset tiles (aka cards).

Note: Adding this toggle card component resulted in a shift that moved the content below the existing tiles further down.  Marmoset tiles are expected to have only two tiles so content will be shorter than normal.  This results in stats (in white showing against a white background).  This will be addressed as a separate ticket (MG-997) in the future.

## Related Issue

- [MG-933](https://sagebionetworks.jira.com/browse/MG-933) — Create generic Split Button component
  - Related (not implemented in this PR): [MG-932](https://sagebionetworks.jira.com/browse/MG-932) — Home Page: Add marmoset tiles (follow-up that replaces the placeholders added here)

## Changelog

- Add reusable `ToggleCardComponent` to `explorers/ui`: a segmented toggle built on PrimeNG `SelectButton`, themeable via CSS custom properties and responsive via CSS container queries.
- Add Storybook stories demonstrating the component in both selected states.
- Integrate the toggle into the Model-AD home page to switch between the mouse tile set and marmoset tiles, wrapped in a rounded panel matching the design.
- Add mouse/marmoset icon SVGs and a `--color-secondary-dark` Model-AD color token for the selected segment.
- Add marmoset placeholder tiles as a temporary stand-in until MG-932.
- Add unit tests for the new component and extend the home page spec to cover toggling.

## Testing

- Open the Model-AD home page and confirm **Mouse Models** is selected by default, showing the four mouse tiles (Model Overview, Model Search, Differential Expression, Disease Correlation).
- Click **Marmoset Models** and confirm the tiles swap to the marmoset placeholders and the mouse tiles are no longer shown; click back and confirm the mouse tiles return.
- Confirm the selected segment renders as the dark purple with white text and the unselected segment is white with purple text, with a divider between them.
- Verify accessibility: the toggle group exposes an accessible name (`aria-label="Select a model species"`), and the segments are operable and selectable via keyboard.
- Narrow the browser window and confirm that for the mobile view, each split button text ("Mouse Models" and "Marmoset Models") wraps in the mobile view.
- In Storybook, open **UI/ToggleCardComponent** and confirm the `FirstSelected` and `ManySelected` stories render correctly.

## Preview

"Mouse Models" selected (desktop):
image

"Marmoset Models" (desktop):
image

"Mouse Models" selected (mobile):
image

"Marmoset Models" (mobile):
image

Background issue to be fixed in [MG-997](https://sagebionetworks.jira.com/browse/MG-997) - placement differs in desktop and mobile / species combos.
image

[MG-933]: https://sagebionetworks.jira.com/browse/MG-933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/model-ad/home/src/lib/home.component.scss
```diff
@@ -58,12 +58,31 @@ explorers-home-card {
   margin: 0;
 }
 
+.models-panel {
+  display: flex;
+  flex-direction: column;
+  gap: 21px;
+  margin: 40px auto 20px;
```
hallieswan: Just curious, where did this margin come from? In the design, it looks like there should only be 50px between the `text-marquee` and the `models-panel`, but in the current implementation, there is the existing 60px bottom padding on the `text-marquee` and this 20px top margin on the `models-panel` for a total of 80px between the `text-marquee` and the `models-panel`. Maybe the design has a drifted a bit since this implementation was started?

libs/model-ad/home/src/lib/home.component.ts
```diff
@@ -27,8 +32,31 @@ export class HomeComponent {
 
   readonly backgroundImage = signal(this.backgroundImageDesktop);
 
+  readonly selectedSpecies = signal<Species>('mouse');
+
+  readonly speciesOptions: (ToggleCardOption & { value: Species })[] = [
+    {
+      label: 'Mouse Models',
+      value: 'mouse',
+      imagePath: 'model-ad-assets/images/mouse-model.svg',
```
hallieswan: Can we use the [existing asset](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/model-ad/assets/images/mouse-with-brain-network.svg) used in the Mouse "Model Overview" tile? 


```suggestion
      imagePath: 'model-ad-assets/images/mouse-with-brain-network.svg.svg',
```

libs/model-ad/home/src/lib/home.component.scss
```diff
@@ -58,12 +58,31 @@ explorers-home-card {
...
+  padding: 40px 20px;
```
hallieswan: Just curious where this padding came from? The latest design uses 59px for top/bottom padding

libs/model-ad/home/src/lib/home.component.scss
```diff
@@ -134,6 +153,22 @@ explorers-home-card {
...
+  }
+
+  .species-toggle {
+    // Keep the left edge aligned with the cards, but bleed off the right so the
+    // segments scroll off-screen; drop the pill frame to sit flat edge-to-edge.
+    --toggle-card-border-radius: 0;
+    --toggle-card-shadow: none;
+
+    margin-inline-end: calc(50% - 50vw);
+  }
```
hallieswan: I noticed that the "Marmoset Models" text wraps in mobile, but the "Mouse Models" text doesn't. The design looks like the text of both toggle buttons should wrap in mobile -- 

[design](https://www.figma.com/design/GtUYqZBxoOnPtjV1hLcTgw/Model-AD-Explorer-Designs?node-id=5005-4312&m=dev) | pr 
:---: | :---: 
Image | Image

libs/explorers/ui/src/lib/components/toggle-card/toggle-card.component.html
```diff
@@ -0,0 +1,24 @@
+<div class="toggle-card-scroll">
+  <p-selectButton
+    styleClass="toggle-card"
+    [options]="options()"
```
hallieswan, line 4: Cool that this button can support more than two options! That'll be really useful when other model types are added to Model AD. 

What do you think about changing the name to reflect that it's not a binary "toggle", but can actually support N options? A couple ideas below - although I'm not totally convinced by any of these, naming is hard!

- OptionsCardComponent
- SegmentedSelectorCardComponent
- ButtonStripCardComponent
- TabbedCardComponent --> although, would need to consider whether to include `role="tab"` etc... so this is semantically correct

libs/model-ad/home/src/lib/home.component.scss
hallieswan: The PR description also includes this test: "Narrow the browser window and confirm each segment stacks its icon above the label and the label text is never clipped.". I don't see this behavior when I narrow the screen, and I don't see the icon stacking in the design, so maybe this test is just an artifact in the PR description?

libs/model-ad/config/src/lib/constants.ts
```diff
@@ -6,6 +6,13 @@ export const HELP_URL =
 
 export const SUPPORT_EMAIL = 'modeladexplorer@sagebionetworks.org';
 
+export const SPECIES = ['mouse', 'marmoset'] as const;
+export type Species = (typeof SPECIES)[number];
+
+export function isSpecies(value: unknown): value is Species {
+  return typeof value === 'string' && (SPECIES as readonly string[]).includes(value);
+}
```
hallieswan: Nit: what do you think about using `MODEL_ORGANISM` instead of `SPECIES` here? Jess specified using `modelOrganism` as the query parameter to distinguish between mouse and marmoset models in the model details pages (see [MG-929](https://sagebionetworks.jira.com/browse/MG-929)), so future models may not be separate species?

libs/explorers/ui/src/lib/components/toggle-card/toggle-card.component.stories.ts
```diff
@@ -0,0 +1,46 @@
...
+  args: {
+    value: 'mouse',
+  },
+};
+
+export const SecondSelected: Story = {
+  args: {
+    value: 'marmoset',
+  },
+};
```
hallieswan: Nit: these stories are redundant -- the user can easily toggle the values themselves, so could probably have just a single Demo story or could add a second story with many more options, so that each story demonstrates sufficiently distinct behavior

hallieswan (commented):
This is looking great! Just a few questions for you

libs/model-ad/home/src/lib/home.component.scss
sagely1: 60px is in the v1.1 design: https://www.figma.com/design/GtUYqZBxoOnPtjV1hLcTgw/Model-AD-Explorer-Designs?node-id=4239-14523&m=dev.  Let me know if I am not looking at the wrong one.

libs/model-ad/home/src/lib/home.component.scss
sagely1: I noticed that the overall widths for the paragraph above the toggle card component are not aligned.  Fixed in the next commit.

libs/model-ad/home/src/lib/home.component.scss
sagely1: Good catch.  I'll update the PR description.  Fixed in next commit.

libs/explorers/ui/src/lib/components/toggle-card/toggle-card.component.html
sagely1: "Toggle" to me is not just binary so my thought is this is an example of a multi-stage toggle but I agree naming is hard due to the subjective nature of coming up with a precise name.  Given this is toggling 2 options it seems to handle the current scenario.  Future modifications can consider other names like SpitButtonCardComponent but I leave that up to the future dev since I am not constraining the number of options.

hallieswan (approved):
Looks great!!

libs/model-ad/home/src/lib/home.component.scss
hallieswan: That's the right one! The 60px between the toggle card and the description looks great in the latest set of changes

PR #4145, opened 2026-07-24
feat(model-ad): update model-ad home page to handle dynamic content (MG-997, MG-935)

sagely1:
## Description

The model-organism toggle card added vertical height to the tiles, pushing the stats (Institutions / Genes / Models) down past the hero "swoosh." The stats numbers no longer sat on the arc, leaving light text on a light background with no contrast. A single fixed background arc couldn't track the stats because the content height above them now varies with the selected organism (mouse vs. marmoset tiles differ in height).  It also makes the methodology copy and stat values organism-specific, which also produces differing content heights the new anchoring has to absorb.

The existing implementation did not provide consistent contrast at all widths but this PR addresses the contrast from small mobile widths through desktop widths.

For example, notice the defect below due to the arcs overlapping into dark text below:
(Example at 800px width)
image

Desktop:
The background is split into two independently anchored pieces so the top of the page and the stats band can each stay put regardless of the content height between them:

Upper arc — anchored to the top of #home-page as its background image.
Lower arc — a full-bleed layer (.stats-arc-bg) inside .stats-container, so it tracks the stats wherever they land.

Each breakpoint uses its own artwork (the desktop 2-band arc and the mobile 6-loop arc), swapped by a BreakpointObserver at 850px.

Mobile:
The background is also split into two independently anchored pieces as well with the only difference being that the upper mobile svg asset has two arcs per the Figma design.

Note:
After speaking with Design (@Shysta), the Figma design for mobile does not need more than 3 total non-intersecting arcs and pixel-perfect placement is not needed as long as the arc for the stats is always in the correct place (providing contrast for the light colored text against the darker background arc).

## Related Issue
[MG-997](https://sagebionetworks.jira.com/browse/MG-997)

## Changelog

- Split the arc SVGs into home-arc-bg-upper*.svg (top) and home-arc-bg-lower*.svg (behind stats), for both desktop and mobile. The mobile upper arc was reduced to the two visible bands (top band + second arc).
- home.component.ts — replaced the single backgroundImage signal with isMobile state plus upperArcImage / lowerArcImage computed signals.
- home.component.html — bind the upper arc on #home-page and add the .stats-arc-bg layer as the first child of .stats-container.
- home.component.scss —
.stats-arc-bg: full-bleed layer behind the stats (z-index: -1), height scales with width so the ribbon's curve is fully revealed.
- Mobile upper arc scaled up (background-size: 1800px auto) so only a gentle slice of each loop shows (the intrinsic curve is too steep on a phone), anchored with a fixed offset (background-position: -600px 0) rather than center so framing is identical at every width and the second arc's left edge isn't cut at some widths.
- Removed the obsolete <385px breakpoint block that recolored the stats text as a contrast workaround.
- Cleanup — deleted the now-unused home-arc-bg.svg and home-arc-bg-mobile.svg.

## Testing

- Load the home page on desktop with **Mouse Models** selected and confirm the stats numbers sit on the lower swoosh with clear contrast, and there is no empty gap at the top of the page.
- Toggle to **Marmoset Models** and confirm the stats stay aligned on the lower swoosh (the tile count drops from 4 to 2).
- Confirm the methodology paragraph and stat values change with the toggle — mouse shows the MODEL-AD text with 20K+ genes / 15+ models, marmoset shows the MARMO-AD text with 30K+ genes / 1+ models.
- Verify mobile widths for both mouse and marmoset models: the stats sit on the lower arc with proper contrast, the upper arc(s) render above the stats, and the arcs render without comprimising contrast on desktop and mobile.
- Scroll past the stats and confirm the lower arc layer stays behind the methodology section and does not block clicks or links below it.

## Preview
| Before (desktop)                                                                                     | After (desktop)                                                                                          |
| ------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------- |
| image | image |

## Desktop:
Upper arc:
image

Lower arc:
image

## Mobile:
Upper arcs:
image

Lower arc:
image

[MG-997]: https://sagebionetworks.jira.com/browse/MG-997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/model-ad/home/src/lib/home.component.ts
```diff
@@ -57,29 +73,53 @@ export class HomeComponent {
...
+        "MODEL-AD comprises two research centers with complementary approaches to generating new mouse models that more faithfully recapitulate features of Alzheimer's disease in humans. Mouse models are phenotyped using standardized neuropathology, 'omics, and behavioral measures.",
+      stats: [
+        {
+          label: 'Institutions',
+          value: '5+',
+        },
+        {
+          label: 'Genes',
+          value: '20K+',
+        },
```
hallieswan, line 88: 20K+ matches the genes stats displayed for mouse models in the current site, but the [design](https://www.figma.com/design/GtUYqZBxoOnPtjV1hLcTgw/Model-AD-Explorer-Designs?node-id=4239-14758&m=dev) shows 30K+ genes for mouse models -- should we double check with Jess whether the value should be updated?

hallieswan (commented):
This looks great! Not blocking, but I was wondering whether it was possible to vertically center the stats in mobile as they are in desktop? The stats are shifted towards the bottom of the arc in mobile -- 

desktop (e.g. 980px) | mobile (e.g. 800px)
:---: | :---: 
Image | Image

libs/model-ad/home/src/lib/home.component.ts
sagely1: @JessterB confirmed it to be 20K+

sagely1: @hallieswan In reference to the alignment on mobile, we would need a different asset for mobile as it cannot be shifted vertically with its current shape.  Seems like a followup PR can address this after a design pass (if Design wants to change the shape).

hallieswan: Sounds good!

hallieswan (approved):

PR #4146, opened 2026-07-24
feat(model-ad): switch model details between mouse and marmoset via modelOrganism query param (MG-987)

hallieswan:
## Description

The model details page needs to serve both mouse and marmoset data from the same route. This introduces a `modelOrganism` query parameter that selects which organism's data the page fetches and renders, defaulting to mouse when the parameter is absent so existing links keep working. The page title and meta description are derived from the organism and model name, and a route guard normalizes legacy or malformed URLs into the canonical form so the parameter survives reloads and sharing.

## Related Issue

- [MG-987](https://sagebionetworks.jira.com/browse/MG-987)

## Changelog

- Add a `modelOrganism` query parameter to the model details route that switches between mouse and marmoset data, falling back to mouse when absent
- Rename the API `Organism` schema and path param to `ModelOrganism`, deriving the TypeScript type from the generated api-client and regenerating all clients
- Derive the model details `<title>` and `<meta>` description from the resolved organism and model name
- Add `modelOrganismGuard` to normalize missing, unknown, or wrong-case organism values into a canonical redirect (preserving path, other query params, and hash fragment), applied to every model details route including the wildcard catch-all
- Preserve the `modelOrganism` param across the component's `location.replaceState` URL rewrites so it isn't silently dropped
- Move `resolveModelOrganism`/`isModelOrganism` helpers into `model-ad/util`
- Consolidate first-letter capitalization into a single `capitalizeFirstLetter` helper in `shared/util`, replacing the duplicated logic in Agora's `HelperService` and the explorers `CapitalizeFirstLetterPipe`

## Testing

- Load `/models/APOE4` with no query param and confirm it redirects to `?modelOrganism=mouse` and renders mouse data
- Load `/models/Presenilin 1?modelOrganism=marmoset` and confirm marmoset data is fetched and rendered
- Confirm the browser tab title and page meta description reflect the organism and model name (e.g. "Mouse Model Details | APOE4 AD model")
- Load a URL with a bogus organism (`?modelOrganism=rat`) or wrong case (`?modelOrganism=Marmoset`) and confirm it redirects to the mouse canonical form
- Switch tabs on a mouse model, then reload the page, and confirm the organism is retained
- Load a share link with query parameters and a hash fragment with no model organism param (e.g. `/models/3xTg-AD/biomarkers?tissue=Hippocampus#insoluble-abeta42`) and confirm the existing query parameters and hash fragment survive the redirect to include the model organism query param

### Preview

Loading `/models/APOE4` with no query param redirects to `?modelOrganism=mouse` and renders mouse data:

https://github.com/user-attachments/assets/c9a80fdc-36ba-45da-af2b-e0670c7956af

Title and description are mouse-specific: 
MG-987_mouse

Marmoset data is loaded and rendered at `/models/Presenilin 1?modelOrganism=marmoset` and title and description are marmoset-specific: 
MG-987_marmoset

Loading a URL with a bogus organism (`?modelOrganism=rat`) redirects to the mouse canonical form:

https://github.com/user-attachments/assets/64cbab82-c8f5-4235-a1c1-36f753cb54a9

Loading a share link with query parameters and a hash fragment (`/models/3xTg-AD/biomarkers?tissue=Hippocampus#insoluble-abeta42`) redirects to include the model organism, maintains the existing query param and fragment, scrolls to the correct section (Insoluble A Beta 42), and selects the correct tissue dropdown (Hippocampus): 

https://github.com/user-attachments/assets/c45c1c5d-610c-4f0e-8f05-211c61be1d16

[MG-987]: https://sagebionetworks.jira.com/browse/MG-987?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/api/ModelApi.java
```diff
@@ -42,10 +42,10 @@ default ModelApiDelegate getDelegate() {
     }
 
     /**
-     * GET /models/{organism}/{name} : Get details for a specific model
-     * Retrieve detailed information for a specific model by its name and organism type.
+     * GET /models/{modelOrganism}/{name} : Get details for a specific model
```
hallieswan, line 45: Rename route parameter from `organism` to `modelOrganism` to match frontend terminology

apps/model-ad/app/e2e/differential-expression-comparison-tool.spec.ts
```diff
@@ -334,7 +334,9 @@ test.describe('differential expression', () => {
     await link.click();
     const popup = await popupPromise;
 
-    await popup.waitForURL((url) => url.toString().endsWith(`/models/${specialModelEncoded}`));
+    await popup.waitForURL((url) =>
+      url.toString().endsWith(`/models/${specialModelEncoded}?modelOrganism=mouse`),
+    );
```
hallieswan, line 339: Model details pages should always have a `modelOrganism` query param now

apps/model-ad/app/src/app/app.routes.ts
```diff
@@ -1,6 +1,20 @@
...
+const modelDetailsData = {
+  title: (route: ActivatedRouteSnapshot) => {
+    const organism = resolveModelOrganism(route.queryParams['modelOrganism']);
+    return `${capitalizeFirstLetter(organism)} Model Details | ${route.params['name']} AD model`;
+  },
+  description: (route: ActivatedRouteSnapshot) => {
+    const organism = resolveModelOrganism(route.queryParams['modelOrganism']);
+    return `Explore information and results for the ${route.params['name']} Alzheimer's Disease ${organism} model.`;
+  },
+};
```
hallieswan, line 17: Add model organism specific route info

apps/model-ad/app/src/app/app.routes.ts
```diff
@@ -84,27 +98,29 @@ export const routes: Route[] = [
   },
   {
     path: `${ROUTE_PATHS.MODELS}/:name`,
+    canActivate: [modelOrganismGuard],
```
hallieswan, line 101: Guard ensures that the model organism query param is added to legacy paths and invalid values (i.e. not mouse or marmoset) fallback to mouse

libs/agora/services/src/lib/helper.service.ts
```diff
@@ -122,9 +122,4 @@ export class HelperService {
     const param = new HttpParams().set(paramName, paramValue);
     return `${url.split('?')[0]}?${param.toString()}`;
   }
-
-  capitalizeFirstLetterOfString(s: string) {
-    if (s.length === 0) return '';
-    return s[0].toUpperCase() + s.slice(1);
-  }
```
hallieswan, line 129: Logic was duplicated in the `CapitalizeFirstLetterPipe` and would have been duplicated in the new model details route data method -- move into the shared typescript helper instead

libs/model-ad/config/src/lib/constants.ts
```diff
@@ -6,13 +6,6 @@ export const HELP_URL =
 
 export const SUPPORT_EMAIL = 'modeladexplorer@sagebionetworks.org';
 
-export const MODEL_ORGANISMS = ['mouse', 'marmoset'] as const;
-export type ModelOrganism = (typeof MODEL_ORGANISMS)[number];
-
-export function isModelOrganism(value: unknown): value is ModelOrganism {
-  return typeof value === 'string' && (MODEL_ORGANISMS as readonly string[]).includes(value);
-}
-
```
hallieswan, line 15: Moved into `model-ad-util`

libs/model-ad/model-details/src/lib/model-details.component.html
```diff
@@ -4,41 +4,50 @@
...
-            [title]="'Pathology'"
-            [modelName]="model.name"
-            [modelControls]="model.matched_controls"
-            [modelDataList]="model.pathology"
-            [wikiParams]="pathologyWikiParams"
+    @switch (model.type) {
+      @case ('marmoset') {
+        <h1>{{ model.name }}</h1>
+      }
+      @case ('mouse') {
```
hallieswan, line 11: This block is added. Otherwise the rest is unchanged other than updating the mouse block to use `mouseModel` instead of `model`

libs/model-ad/model-details/src/lib/model-details.component.spec.ts
```diff
@@ -44,42 +47,90 @@ async function setup(
...
 }
 
 describe('ModelDetailsComponent', () => {
   afterAll(() => jest.restoreAllMocks());
 
-  it('should display the model name', async () => {
-    await setup();
-    expect(screen.getByRole('heading', { level: 1 })).toHaveTextContent(mouseModelMock.name);
-  });
+  describe('model organism resolution', () => {
```
hallieswan, line 56: Tests are reorganized to add a describe block for model organism resolution, move mouse-specific tests into a mouse model describe block, and add a placeholder marmoset model describe block for future marmoset-specific tests

libs/model-ad/model-details/src/lib/model-details.component.ts
```diff
@@ -44,7 +51,12 @@ export class ModelDetailsComponent implements OnInit, AfterViewInit {
 
   isLoading = true;
 
-  model: MouseModel | undefined;
+  model: Model | undefined;
+  modelOrganism: ModelOrganism = ModelOrganism.Mouse;
+
+  get mouseModel(): MouseModel | undefined {
+    return this.model?.type === 'mouse' ? (this.model as MouseModel) : undefined;
+  }
```
hallieswan, line 59: Narrow the type so the mouse-specific components get a `MouseModel`

libs/model-ad/model-details/src/lib/model-details.component.ts
```diff
@@ -80,37 +92,47 @@ export class ModelDetailsComponent implements OnInit, AfterViewInit {
...
+    combineLatest([this.route.paramMap, this.route.queryParamMap])
+      .pipe(
+        map(([params, queryParams]): [ParamMap, ModelOrganism] => [
+          params,
+          resolveModelOrganism(queryParams.get('modelOrganism')),
+        ]),
+        distinctUntilChanged(
+          ([prevParams, prevModelOrganism], [params, modelOrganism]) =>
+            prevParams.get('name') === params.get('name') && prevModelOrganism === modelOrganism,
+        ),
```
hallieswan, line 111: Fetch a model only when the name or the resolved organism (i.e. missing value, invalid values, missing query param all resolve to "mouse", so don't need to re-request because the resolved value is the same even though the actual query param changed) truly changes

libs/model-ad/util/src/lib/model-organism.ts
```diff
@@ -0,0 +1,11 @@
+import { ModelOrganism } from '@sagebionetworks/model-ad/api-client';
+
+export const MODEL_ORGANISMS = Object.values(ModelOrganism);
```
hallieswan, line 3: Derive `MODEL_ORGANISMS` used on the homepage (and now model details pages) from the generated API enum, so there is a single source of truth for the values

libs/model-ad/util/src/lib/model-organism.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export function isModelOrganism(value: unknown): value is ModelOrganism {
+  return typeof value === 'string' && (MODEL_ORGANISMS as string[]).includes(value);
+}
+
+export function resolveModelOrganism(raw: unknown): ModelOrganism {
+  return isModelOrganism(raw) ? raw : ModelOrganism.Mouse;
+}
```
hallieswan, line 12: Moved `isModelOrganism` into `model-ad-util` since `model-ad-config` otherwise contains constants, and we needed to add another related function (`resolveModelOrganism`)

libs/model-ad/model-details/src/lib/model-details.component.ts
hallieswan, line 65: These variables are mouse specific, but the names don't make that clear. Planning to refactor in future PR that adds marmoset specific model details content -- this PR was already a bit larger than intended due to the `Organism` -> `ModelOrganism` rename, so didn't want to expand the scope further here

libs/model-ad/util/src/lib/model-organism.ts
```diff
@@ -0,0 +1,11 @@
+import { ModelOrganism } from '@sagebionetworks/model-ad/api-client';
+
+export const MODEL_ORGANISMS = Object.values(ModelOrganism);
+
+export function isModelOrganism(value: unknown): value is ModelOrganism {
+  return typeof value === 'string' && (MODEL_ORGANISMS as string[]).includes(value);
+}
+
+export function resolveModelOrganism(raw: unknown): ModelOrganism {
+  return isModelOrganism(raw) ? raw : ModelOrganism.Mouse;
```
rrchai: I see this is intentional that wrong casing redirects to the mouse fallback. I wonder if you'd consider lowercasing first before matching? Just thinking about a shared link with accidental capitalization silently landing on the wrong organism's data.

rrchai (approved):
LGTM overall - just one minor question

libs/model-ad/util/src/lib/model-organism.ts
hallieswan: Good idea! Done in the next commit

apps/model-ad/app/src/app/app.routes.ts
```diff
@@ -84,27 +98,29 @@ export const routes: Route[] = [
...
     loadChildren: () =>
       import('@sagebionetworks/model-ad/model-details').then((routes) => routes.routes),
+    data: modelDetailsData,
   },
   // ensure that all models match a route, so the custom url serializer can encode special characters
   {
     path: `${ROUTE_PATHS.MODELS}/**`,
+    canActivate: [modelOrganismGuard],
     loadChildren: () =>
```
sagely1, line 124: shouldn't this route also include a data parameter like above?
`data: modelDetailsData`

sagely1 (changes requested):
Looks good! Have a question on routes.

apps/model-ad/app/src/app/app.routes.ts
hallieswan: `modelDetailsData` uses `route.params['name']`, but a wildcard route has no `:name` param to fill in, so I left off `data` intentionally here -- it falls back to the default 'Model AD' title instead

sagely1: Ah ok.  That makes sense since this is the CT route so we don't need details info.

libs/model-ad/util/src/lib/model-organism.ts
```diff
@@ -0,0 +1,11 @@
+import { ModelOrganism } from '@sagebionetworks/model-ad/api-client';
+
+export const MODEL_ORGANISMS = Object.values(ModelOrganism);
+
+export function isModelOrganism(value: unknown): value is ModelOrganism {
+  return typeof value === 'string' && (MODEL_ORGANISMS as string[]).includes(value);
+}
+
+export function resolveModelOrganism(raw: unknown): ModelOrganism {
```
sagely1: what do you think about standardizing on `value` instead of `raw` here?

hallieswan: Sure, done in the next commit

sagely1 (approved):
LGTM

PR #4159, opened 2026-08-14
feat(model-ad): add marmoset model overview comparison tool (MG-940)

hallieswan:
## Description

Adds the Marmoset Model Overview comparison tool frontend as a new library that wraps `ComparisonToolComponent` and reads from the existing marmoset endpoint, wires it up as a lazy-loaded route at `/comparison/model/marmoset`, and restructures the header nav so "Model Overview" becomes a dropdown containing both Mouse Models and Marmoset Models.

## Related Issue

- [MG-940](https://sagebionetworks.jira.com/browse/MG-940)

## Changelog

- Add the `model-ad-marmoset-model-overview-comparison-tool` library and register its path in `tsconfig.base.json`
- Add the `MARMOSET_MODEL_OVERVIEW` route path and lazy-loaded app route with page title and description metadata
- Nest Mouse Models and Marmoset Models under a "Model Overview" header dropdown
- Make the explorers header dropdown trigger a real `<button>` with `aria-haspopup`, `aria-expanded`, and `aria-controls` so e2e tests can find it by role, and so it is keyboard-operable and announced as a menu
- Add e2e coverage for marmoset model overview CT
- Add a `ComparisonToolConfigDocument` unit test pinning Spring Data's coercion of a bare-string `ui_config` filter `values` into a single-element list — the marmoset config is the first to store `values` as a single string rather than an array, which the API already handles, so the test flags it if that default behavior ever changes
- Update the `libs/model-ad/CLAUDE.md` library table and comparison tool notes

## Testing

- Serve the Model-AD stack, hover/click "Model Overview" in the header, and confirm the dropdown lists Mouse Models and Marmoset Models, both navigating to the correct routes
- Tab to the "Model Overview" nav item with the keyboard and confirm it receives focus and opens with Enter/Space, then close it with Escape
- Shrink the viewport to mobile width and confirm the Model Overview children still render inline in the hamburger menu, with no separate trigger to click
- Load `/comparison/model/marmoset`, confirm the table populates, then verify the filter panel (Model Type, Modified Gene, Available Data) narrows results and that selections round-trip through the URL on reload
- Confirm the Modified Gene filter panel lists `PSEN1` — this is the marmoset config entry whose `values` is stored as a bare string rather than an array
- Click a row's View Details button and a Results/View Data link, and confirm each opens in a new tab at the marmoset model details page (with `modelOrganism=marmoset`) or the Synapse study page

### Preview

MG-940

New header menu and opening marmoset details page: 

https://github.com/user-attachments/assets/a3275320-4814-4ca1-afa1-1384b4143e59

Pinning: 

https://github.com/user-attachments/assets/4db75bb8-deb5-4c87-b7ce-11db558c83a8

[MG-940]: https://sagebionetworks.jira.com/browse/MG-940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/model-ad/api-next/src/test/java/org/sagebionetworks/model/ad/api/next/model/document/ComparisonToolConfigDocumentTest.java
+++ b/apps/model-ad/api-next/src/test/java/org/sagebionetworks/model/ad/api/next/model/document/ComparisonToolConfigDocumentTest.java
@@ -1,0 +1,84 @@
+package org.sagebionetworks.model.ad.api.next.model.document;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+import org.bson.Document;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.springframework.data.mongodb.core.convert.MappingMongoConverter;
+import org.springframework.data.mongodb.core.convert.NoOpDbRefResolver;
+import org.springframework.data.mongodb.core.mapping.MongoMappingContext;
+
+/**
+ * Pins the Spring Data conversion behavior that reads of the `ui_config` collection depend on.
+ * Some filters store `values` as a bare string instead of an array (e.g. `"values": "PSEN1"` in
+ * the marmoset config), while both this document class and the API contract declare it as a list.
+ * Spring's ObjectToCollectionConverter wraps the lone string during the read; no mapping code
+ * performs that normalization, so losing it would silently empty a filter panel.
+ */
+class ComparisonToolConfigDocumentTest {
+
+  private MappingMongoConverter converter;
+
+  @BeforeEach
+  void setUp() {
+    MongoMappingContext mappingContext = new MongoMappingContext();
+    mappingContext.afterPropertiesSet();
+
+    converter = new MappingMongoConverter(NoOpDbRefResolver.INSTANCE, mappingContext);
+    converter.afterPropertiesSet();
+  }
+
+  @Test
+  @DisplayName("should read bare string filter values as a single-element list")
+  void shouldReadBareStringFilterValuesAsSingleElementList() {
+    // given
+    Document source = configWithFilterValues("PSEN1");
+
+    // when
+    ComparisonToolConfigDocument result = converter.read(
+      ComparisonToolConfigDocument.class,
+      source
+    );
+
+    // then
+    assertThat(filterValuesOf(result)).containsExactly("PSEN1");
+  }
+
+  @Test
+  @DisplayName("should read array filter values as a list")
+  void shouldReadArrayFilterValuesAsList() {
+    // given
+    Document source = configWithFilterValues(List.of("Familial AD", "Late Onset AD"));
+
+    // when
+    ComparisonToolConfigDocument result = converter.read(
+      ComparisonToolConfigDocument.class,
+      source
+    );
+
+    // then
+    assertThat(filterValuesOf(result)).containsExactly("Familial AD", "Late Onset AD");
+  }
+
+  private static Document configWithFilterValues(Object values) {
+    Document filter = new Document()
+      .append("name", "Modified Gene")
+      .append("data_key", "modified_genes")
+      .append("query_param_key", "modifiedGenes")
+      .append("values", values);
+
+    return new Document()
+      .append("page", "Marmoset Model Overview")
+      .append("dropdowns", List.of())
+      .append("columns", List.of())
+      .append("filters", List.of(filter));
+  }
+
+  private static List<String> filterValuesOf(ComparisonToolConfigDocument document) {
+    assertThat(document.getFilters()).hasSize(1);
+    return document.getFilters().get(0).getValues();
+  }
+}
--- a/apps/model-ad/app/e2e/constants.ts
+++ b/apps/model-ad/app/e2e/constants.ts
@@ -10,6 +10,13 @@ export const COMPARISON_TOOL_API_PATHS: Record<string, string> = {
   'Disease Correlation': '/comparison-tools/disease-correlation',
 };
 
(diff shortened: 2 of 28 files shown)


apps/model-ad/api-next/src/test/java/org/sagebionetworks/model/ad/api/next/model/document/ComparisonToolConfigDocumentTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.springframework.data.mongodb.core.mapping.MongoMappingContext;
+
+/**
+ * Pins the Spring Data conversion behavior that reads of the `ui_config` collection depend on.
+ * Some filters store `values` as a bare string instead of an array (e.g. `"values": "PSEN1"` in
+ * the marmoset config), while both this document class and the API contract declare it as a list.
+ * Spring's ObjectToCollectionConverter wraps the lone string during the read; no mapping code
+ * performs that normalization, so losing it would silently empty a filter panel.
+ */
+class ComparisonToolConfigDocumentTest {
```
hallieswan, line 21: Marmoset Model Overview CT's ui_config uses a string for a filter with a single value, rather than an array. The API already handles converting the string to an array by default (to match the expected type), but add a test to ensure we're alerted if this default behavior changes

apps/model-ad/app/e2e/helpers/comparison-tool.ts
```diff
@@ -34,7 +37,17 @@ export const navigateToComparison = async (
...
+    const navTrail = COMPARISON_TOOL_NAV_TRAILS[name];
+    if (navTrail.length > 1) {
+      // Desktop renders the parent nav item as a dropdown trigger that must be opened first;
+      // mobile renders the children directly, so there is no trigger to click.
+      const dropdownTrigger = page.getByRole('button', { name: navTrail[0] });
+      if (await dropdownTrigger.isVisible().catch(() => false)) {
+        await dropdownTrigger.click();
+      }
+    }
+    await page.getByRole('link', { name: navTrail[navTrail.length - 1] }).click();
```
hallieswan, line 50: Handle header dropdown menu links when navigating to a CT

libs/explorers/ui/src/lib/components/header/header.component.html
```diff
@@ -22,17 +22,22 @@
               @for (link of links; track link.label) {
                 <li [class.mobile-dropdown-item]="link.children && isMobile">
                   @if (link.children && !isMobile) {
-                    <a
+                    <button
+                      type="button"
```
hallieswan, line 26: Convert dropdown to a button, so e2e tests can find it by role, and so it is keyboard-operable and announced as a menu

apps/model-ad/app/e2e/constants.ts
```diff
@@ -1,19 +1,30 @@
...
   'Differential Expression': '/comparison-tools/transcriptomics',
   'Disease Correlation': '/comparison-tools/disease-correlation',
 };
 
+// Header navigation path to each comparison tool, from the top-level nav item to the link itself
+export const COMPARISON_TOOL_NAV_TRAILS: Record<string, string[]> = {
+  'Marmoset Model Overview': ['Model Overview', 'Marmoset Models'],
+  'Model Overview': ['Model Overview', 'Mouse Models'],
+  'Differential Expression': ['Differential Expression'],
+  'Disease Correlation': ['Disease Correlation'],
```
sagely1, line 20: For consistency, what do you think about refactoring the records for 'Model Overview' to 'Mouse Model Overview'?

apps/model-ad/app/e2e/constants.ts
```diff
@@ -1,19 +1,30 @@
...
+};
+
 export const COMPARISON_TOOL_CONFIG_PATH = 'comparison-tools/config';
 
 // Default sort configurations for each comparison tool (required by API)
 export const COMPARISON_TOOL_DEFAULT_SORTS: Record<string, { field: string; order: 1 | -1 }[]> = {
+  'Marmoset Model Overview': [{ field: 'name', order: 1 }],
   'Model Overview': [
```
sagely1, line 28: We could consider `Mouse Model Overview` here too.

apps/model-ad/app/e2e/constants.ts
hallieswan: That work was already marked as in progress in https://sagebionetworks.jira.com/browse/MG-944, so I was planning to leave the change to that ticket. Happy to make the change here, if that's preferred though!

apps/model-ad/app/e2e/constants.ts
hallieswan: See above!

apps/model-ad/app/e2e/marmoset-model-overview-comparison-tool.spec.ts
```diff
@@ -0,0 +1,215 @@
+import { expect, test } from '@playwright/test';
+import {
+  clickViewDetailsButtonByName,
+  ColumnConfig,
+  expectPinnedParams,
+  getPinnedTable,
+  getQueryParamFromValues,
```
sagely1: you can drop `getQueryParamFromValues` as it is unused

sagely1 (changes requested):

apps/model-ad/app/e2e/constants.ts
sagely1: Ah that makes sense.  Let's defer that change until later.

apps/model-ad/app/e2e/marmoset-model-overview-comparison-tool.spec.ts
hallieswan: Done in the next commit

sagely1 (approved):
LGTM

PR #4163, opened 2026-08-17
feat(explorers): add _id tiebreaker to CT sort for deterministic pagination (MG-1034, AG-2181)

rrchai:
## Description

When multiple rows share the same sort value, MongoDB doesn't guarantee their order across paginated requests — a row can appear on two pages or be missing entirely. This appends `_id` as a final tiebreaker to the CT sort document, ensuring deterministic ordering across all paginated queries in both apps.

## Related Issue

- [MG-1034](https://sagebionetworks.jira.com/browse/MG-1034)
- [AG-2181](https://sagebionetworks.jira.com/browse/AG-2181)

## Changelog

- Append `_id: 1` to the sort document in `ComparisonToolRepositorySupport.buildSortOperation()` after all user-requested sort fields
- Add unit tests verifying `_id` tiebreaker is present and positioned after user sort fields

## Testing

- Navigate to any CT page (e.g. Disease Correlation), sort by a column with duplicate values (e.g. `model_type`), paginate back and forth — rows should never duplicate or disappear across pages

[MG-1034]: https://sagebionetworks.jira.com/browse/MG-1034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AG-2181]: https://sagebionetworks.jira.com/browse/AG-2181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
+++ b/libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
@@ -413,6 +413,7 @@ public abstract class ComparisonToolRepositorySupport<T> {
       sortDoc.append(ApiHelper.isEmptyFlagKey(field), 1);
       sortDoc.append(resolved, order.isAscending() ? 1 : -1);
     }
+    sortDoc.append("_id", 1);
 
     return context -> new Document("$sort", sortDoc);
   }
--- a/libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
+++ b/libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
@@ -413,7 +413,11 @@ public abstract class ComparisonToolRepositorySupport<T> {
       sortDoc.append(ApiHelper.isEmptyFlagKey(field), 1);
       sortDoc.append(resolved, order.isAscending() ? 1 : -1);
     }
-    sortDoc.append("_id", 1);
+    // Break ties on _id so equal sort values keep a stable order across pages. Skip when the
+    // caller already sorts by _id -- re-appending would overwrite their direction.
+    if (!sortDoc.containsKey("_id")) {
+      sortDoc.append("_id", 1);
+    }
 
     return context -> new Document("$sort", sortDoc);
   }
--- a/libs/explorers/api-helper/src/test/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupportTest.java
+++ b/libs/explorers/api-helper/src/test/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupportTest.java
@@ -429,6 +429,35 @@ class ComparisonToolRepositorySupportTest {
   }
 
   @Test
+  @DisplayName("should append _id once after every sort key when multiple sort orders are requested")
+  void shouldAppendIdTiebreakerOnceAfterAllSortKeysForMultiSort() {
+    ComputedRepo repo = new ComputedRepo(mongoTemplate);
+    stubMongoTemplate(0L);
+
+    Pageable pageable = PageRequest.of(
+      0,
+      10,
+      Sort.by(Sort.Order.asc("name"), Sort.Order.desc("hgnc_symbol"))
+    );
+    repo.run(new Criteria(), pageable);
+
+    String pipeline = capturePipeline().toString();
+    assertThat(pipeline)
+      .as("_id must be appended exactly once, not per sort order")
+      .containsOnlyOnce("\"_id\" : 1");
+
+    int computedIdx = pipeline.indexOf("\"name_sort\" : 1");
+    int plainIdx = pipeline.indexOf("\"hgnc_symbol\" : -1");
+    int idIdx = pipeline.indexOf("\"_id\" : 1");
+    assertThat(idIdx)
+      .as("_id must rank below the computed first sort key, not between the sort keys")
+      .isGreaterThan(computedIdx);
+    assertThat(idIdx)
+      .as("_id must rank below the second sort key so it only breaks full ties")
+      .isGreaterThan(plainIdx);
+  }
+
+  @Test
   @DisplayName("should not include _id tiebreaker when sort is unsorted")
   void shouldNotIncludeIdTiebreakerWhenUnsorted() {
     BareRepo repo = new BareRepo(mongoTemplate);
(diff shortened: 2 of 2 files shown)


libs/explorers/api-helper/src/test/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupportTest.java
```diff
@@ -411,4 +411,33 @@ void shouldUseOriginalFieldNameForIsEmptyWhenSortResolvesToComputedAlias() {
       assertThat(pipeline).contains("\"name_sort\" : 1");
     }
   }
+
+  @Test
```
sagely1, line 415: Good test.  Can you write an additional test for multi-sort too and ensure the _id is also appended correctly?

sagely1 (changes requested):
Looks good.  I'm suggesting a further test.

libs/explorers/api-helper/src/test/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupportTest.java
rrchai: Good call - I will add an additional test for mutli-sort in next commit.

libs/explorers/api-helper/src/test/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupportTest.java
```diff
@@ -411,4 +411,62 @@ void shouldUseOriginalFieldNameForIsEmptyWhenSortResolvesToComputedAlias() {
...
+
+    String pipeline = capturePipeline().toString();
+    assertThat(pipeline).contains("\"_id\" : 1");
+    int nameIdx = pipeline.indexOf("\"name\" : 1");
+    int idIdx = pipeline.indexOf("\"_id\" : 1");
+    assertThat(idIdx).as("_id should appear after the user-requested sort field").isGreaterThan(nameIdx);
+  }
+
+  @Test
+  @DisplayName("should append _id once after every sort key when multiple sort orders are requested")
```
hallieswan: Nit: could consider rewording this to be a bit clearer: 
```suggestion
  @DisplayName("should append _id once, after all sort keys, when multiple sort orders are requested")
```

libs/explorers/api-helper/src/test/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupportTest.java
```diff
@@ -411,4 +411,62 @@ void shouldUseOriginalFieldNameForIsEmptyWhenSortResolvesToComputedAlias() {
...
+    assertThat(pipeline)
+      .as("_id must be appended exactly once, not per sort order")
+      .containsOnlyOnce("\"_id\" : 1");
+
+    int computedIdx = pipeline.indexOf("\"name_sort\" : 1");
+    int plainIdx = pipeline.indexOf("\"hgnc_symbol\" : -1");
+    int idIdx = pipeline.indexOf("\"_id\" : 1");
+    assertThat(idIdx)
+      .as("_id must rank below the computed first sort key, not between the sort keys")
+      .isGreaterThan(computedIdx);
```
hallieswan, line 454: Nit: could consider updating the description to clarify that this doesn't assert anything about whether `_id` is not between keys (the next assert does that): 

```suggestion
    assertThat(idIdx)
      .as("_id must rank below the computed first sort key")
      .isGreaterThan(computedIdx);
```

libs/explorers/api-helper/src/test/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupportTest.java
```diff
@@ -411,4 +411,62 @@ void shouldUseOriginalFieldNameForIsEmptyWhenSortResolvesToComputedAlias() {
...
+    assertThat(idIdx)
+      .as("_id must rank below the second sort key so it only breaks full ties")
+      .isGreaterThan(plainIdx);
+  }
+
+  @Test
+  @DisplayName("should not include _id tiebreaker when sort is unsorted")
```
hallieswan: Nit: could consider rewording for clarity


```suggestion
  @DisplayName("should not include _id tiebreaker when no sort is requested")
```

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
```diff
@@ -413,6 +413,7 @@ private AggregationOperation buildSortOperation(
       sortDoc.append(ApiHelper.isEmptyFlagKey(field), 1);
       sortDoc.append(resolved, order.isAscending() ? 1 : -1);
     }
+    sortDoc.append("_id", 1);
```
hallieswan: I think this will overwrite a case where a caller has already sorted by `_id` descending -- should we add a guard to prevent this? e.g. something like: 

```suggestion
    if (!sortDoc.containsKey("_id")) sortDoc.append("_id", 1);
```

hallieswan: Could we also add a comment here explaining why we add the `_id` sort, for future devs? e.g. something like "// break ties with _id to ensure consistent results across pages"

hallieswan (commented):
Looks good! Just a couple small comments for you

libs/explorers/api-helper/src/test/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupportTest.java
rrchai: done in next commit(s)

libs/explorers/api-helper/src/test/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupportTest.java
rrchai: done in next commit(s)

libs/explorers/api-helper/src/test/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupportTest.java
rrchai: done in next commit(s)

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
rrchai: Good catch - guard and comment added. I also added a test to cover it in next commit(s).

hallieswan (approved):
Looks good to me!

sagely1 (approved):
LGTM

PR #4165, opened 2026-08-18
feat(model-ad): port search to Java with marmoset support (MG-537, MG-636, MG-937, MG-938)

rrchai:
## Description

Search was the last endpoint still served by the legacy Express API, and it had to move to the Java api-next server before it could be extended to marmoset models. The port also resolves the routing ambiguity between `GET /models/search` and `GET /models/{modelOrganism}/{name}` by adopting the `GET /search/models` convention, which supports a backend merge across collections and extends to future organisms such as iPSC. The header search now returns mouse and marmoset results merged and ranked by match quality, while the home page tiles scope to a single organism through the new `modelOrganisms` parameter.

## Related Issue

- [MG-537](https://sagebionetworks.jira.com/browse/MG-537) — API: Port search route
- [MG-636](https://sagebionetworks.jira.com/browse/MG-636) — API: Rename /model/search route
- [MG-937](https://sagebionetworks.jira.com/browse/MG-937) — Update search route to support marmoset search
- [MG-938](https://sagebionetworks.jira.com/browse/MG-938) — Update model-ad search input to display correct icon based on mouse vs marmoset search result
## Changelog

- Add `GET /search/models` to the api-next spec under a new `Search` tag, so the generated client exposes the operation as `SearchService` rather than `ModelService`, with a required `q` (1–100 characters) and an optional `modelOrganisms` array, add the required `model_organism` field to `SearchResult`, and delete the legacy `/models/search` path definition
- Add `ModelSearchRepositoryImpl`, which queries `model_details` and `marmo_details` separately and merges the results in Java, ranking by match precedence then name — mouse matches on name, aliases, jax_id, and rrid, marmoset on name — so name matches from both organisms interleave alphabetically
- Avoid the `$unionWith` aggregation stage, which DocumentDB does not support, and sort in Java once both collections are merged; escape the query with `Pattern.quote` so model names containing regex metacharacters such as `Abca7*V1599M` and `LOAD1.Abca7A1527G` match literally
- Add `ModelSearchService` with response caching, plus a `ConstraintViolationException` handler so an out-of-range `q` returns `400 application/problem+json` instead of a 500
- Wire the model-ad search input to the generated `SearchService`, show mouse and marmoset icons based on `model_organism`, navigate with the `modelOrganism` query param, and enable the marmoset home card search
- Route `/api/v1/search/*` to api-next in the Caddyfile, replacing the handler that forwarded `/api/v1/models/search*` to Express
- Add a shared `mockSearchResults` fixture to `model-ad/testing`, and update the search input stories to provide `SearchService` with a mock that filters by `modelOrganisms`, renaming `HomeSearchInput` to `MouseHomeSearchInput` and adding `MarmosetHomeSearchInput`

## Testing

- Start the full stack (`model-ad-build-images && model-ad-docker-start`) and navigate to http://localhost:8000
- On the home page mouse tab, search "apoe" in the Model Search card and confirm only mouse models return, each with a mouse icon; switch to the marmoset tab, search "presenilin" in the Marmoset Model Search card and confirm "Presenilin 1" returns with a marmoset icon
- In the header search input, search "il" and confirm results include both organisms — `LOAD1.Il1rapExon2KO` (mouse) and `Presenilin 1` (marmoset)
- In the header search input, search "trem" and confirm name matches (`Trem2-R47H_NSS`, `Trem2R47H`) are listed before alias matches (the 8 `LOAD1` models)
- Click a search result and confirm navigation includes the organism, e.g. `/models/APOE4?modelOrganism=mouse` or `/models/Presenilin%201?modelOrganism=marmoset`
- Confirm the endpoint moved: `curl "http://localhost:8000/api/v1/search/models?q=apoe"` returns results, and `curl -i "http://localhost:8000/api/v1/models/search?q=apoe"` returns 404
- Confirm invalid input returns 400 rather than 500: `curl -i "http://localhost:8000/api/v1/search/models?q="` and `curl -i "http://localhost:8000/api/v1/search/models?q=apoe&modelOrganisms=rat"`
- Confirm regex metacharacters are treated literally: `curl "http://localhost:8000/api/v1/search/models?q=%5ELOAD1"` returns `[]` because `^` is matched as a character rather than an anchor, while `curl "http://localhost:8000/api/v1/search/models?q=LOAD1"` returns the 8 LOAD1 models

#### Header search — cross-organism results with marmoset interleaved among mouse models
Screenshot 2026-08-20 at 11 58 30 AM

#### Marmoset home card search — filters to marmoset models only
Screenshot 2026-08-20 at 11 58 49 AM

#### Storybook marmoset home card search story

Screenshot 2026-08-28 at 12 53 46 PM

[MG-537]: https://sagebionetworks.jira.com/browse/MG-537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MG-636]: https://sagebionetworks.jira.com/browse/MG-636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MG-937]: https://sagebionetworks.jira.com/browse/MG-937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MG-938]: https://sagebionetworks.jira.com/browse/MG-938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MG-936]: https://sagebionetworks.jira.com/browse/MG-936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/model-ad/apex/Caddyfile
```diff
@@ -21,10 +21,10 @@
    reverse_proxy {env.API_NEXT_HOST}:{env.API_NEXT_PORT}
```
rrchai, line 21: The legacy api server still exists but no longer handles any active routes after this change. Full removal and replacement by api-next server will be done in a separate cleanup PR.

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+    return List.of();
+  }
+
+  private List<SearchResultDocument> searchBothOrganisms(String escapedQuery) {
+    List<AggregationOperation> operations = new ArrayList<>();
+    operations.add(
+        addFieldsStage(mouseMatchBranches(escapedQuery), escapedQuery, ModelOrganismDto.MOUSE));
+    operations.add(matchNonNull());
+    operations.add(unionWithStage(escapedQuery));
```
rrchai: Query `marmo_details`or name matches, tag "marmoset", and merge into the mouse results via `$unionWith`

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
```diff
@@ -0,0 +1,185 @@
...
+    operations.add(sortStage());
```
rrchai: Sort combined results by match quality then alphabetical — mouse and marmoset interleaved

libs/model-ad/ui/src/lib/components/search-input/search-input.component.ts
```diff
@@ -14,27 +15,37 @@ import { Observable } from 'rxjs';
...
+    return this.resultIconByOrganism[(result as ModelAdSearchResult).model_organism];
+  };
 
   navigateToResult = (result: SearchResult): void => {
-    this.router.navigate([ROUTE_PATHS.MODELS, result.id]);
+    const modelOrganism = (result as ModelAdSearchResult).model_organism ?? ModelOrganism.Mouse;
+    this.router.navigate([ROUTE_PATHS.MODELS, result.id], {
+      queryParams: { modelOrganism },
+    });
   };
```
hallieswan, line 38: Since explorers-search-input is generic now, we can type these callbacks to Model AD's `SearchResult` directly and remove the casts across this file. Since `model_organism` is also required in the Model AD type, we can also drop the `?? ModelOrganism.Mouse` fallback

libs/model-ad/ui/src/lib/components/search-input/search-input.component.ts
```diff
@@ -14,27 +15,37 @@ import { Observable } from 'rxjs';
...
 
   getSearchResults = (query: string): Observable<SearchResult[]> => {
-    return this.modelService.searchModels(query);
+    return this.searchService.searchModels(query, this.modelOrganisms());
   };
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   checkQueryForErrors = (query: string): string => {
-    return ''; // empty string if no error
```
hallieswan, line 37: Can we leave this comment? It may not be obvious that we return an empty string to ensure that no error is shown

libs/model-ad/api-description/src/paths/search/models.yaml
hallieswan, line 1: Nit: now that the path is `/search/models`, what do you think about renaming this file to `paths/search/models.yaml`? Other file paths mirror the route path (e.g. `comparison-tools/disease-correlation.yaml`), so the change would align this file with that convention

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/service/ModelSearchService.java
```diff
@@ -0,0 +1,44 @@
...
+  private final ModelSearchRepository modelSearchRepository;
+  private final SearchResultMapper searchResultMapper;
+
+  @Cacheable(key = "#query + '-' + #modelOrganisms")
+  public List<SearchResultDto> searchModels(String query, List<ModelOrganismDto> modelOrganisms) {
+    if (query.length() < MIN_QUERY_LENGTH || query.length() > MAX_QUERY_LENGTH) {
+      throw new IllegalArgumentException(
+          "Query must be between " + MIN_QUERY_LENGTH + " and " + MAX_QUERY_LENGTH
+              + " characters");
+    }
```
hallieswan: Would it be possible to move this constraint into the `q` schema instead (`libs/model-ad/api-description/src/paths/model-search.yaml`)? I'm wondering if that would get us automatic enforcement of this limit, rather than having to roll our own.

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/api/SearchApiDelegateImpl.java
```diff
@@ -0,0 +1,36 @@
...
+@Slf4j
+public class SearchApiDelegateImpl implements SearchApiDelegate {
+
+  private final ModelSearchService modelSearchService;
+
+  @Override
+  public ResponseEntity<List<SearchResultDto>> searchModels(String q,
+      List<ModelOrganismDto> modelOrganisms) {
+    List<ModelOrganismDto> sortedOrganisms = modelOrganisms == null ? null
+        : modelOrganisms.stream().sorted().collect(Collectors.toList());
```
hallieswan: It looks like we split processing the `modelOrganisms` query param across `SearchApiDelegateImpl` and `ModelSearchService` -- what do you think about consolidating it all here? Then we can minimize cache entries (e.g. `null`, `[]`, and `["mouse", "marmoset"]` would all map to the same cache key) and colocate all the processing in one place, so the service can be simplified. We could also drop the duplicated log.debug entry in the service

hallieswan: Also small nit -- could change `Collections.toList()` to just `.toList()`, then could drop the `Collectors` import (as we do in the `ModelSearchService`)

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/configuration/CacheConfiguration.java
```diff
@@ -40,6 +40,7 @@ public CacheManager cacheManager() {
         CacheNames.TRANSCRIPTOMICS,
         CacheNames.TRANSCRIPTOMICS_INDIVIDUAL,
         CacheNames.MODEL,
+        CacheNames.MODEL_SEARCH,
```
hallieswan, line 43: Should we also add this to the `CacheConfigurationTest` for consistency?

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
hallieswan: Unfortunately, it looks like `$unionWith` is _not_ supported by any DocDB version -- see AWS's supported APIs table [here](https://docs.aws.amazon.com/documentdb/latest/devguide/mongo-apis.html). I'm sorry I missed this in the original ticket write-up. We'll have to come up with another approach -- I think we can rework this to 1) run an aggregation per collection that adds the relevant fields (e.g. match_info, model_organism), drops the non-matching entries, then projects to the final SearchResult shape (plus precedence), then 2) combine the results in Java and sort by precedence and id. The mapper will drop precedence, so it won't be returned in the actual response. I think this will also simplify the code

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
```diff
@@ -0,0 +1,185 @@
...
+public class ModelSearchRepositoryImpl implements ModelSearchRepository {
+
+  private static final String MOUSE_COLLECTION = "model_details";
+  private static final String MARMOSET_COLLECTION = "marmo_details";
+
+  private final MongoTemplate mongoTemplate;
+
+  @Override
+  public List<SearchResultDocument> searchModels(String query, List<ModelOrganismDto> organisms) {
+    String escapedQuery = Pattern.quote(query);
```
hallieswan, line 32: The previous implementation manually escaped each meta character, whereas this wraps the entire string with `\Q` and `\E` -- have we checked whether this behavior is supported in DocDB? Just curious because AWS's [`$regexMatch`](https://docs.aws.amazon.com/documentdb/latest/devguide/regexMatch.html) page documents `input`/`regex`/`options`, but doesn't mention `\Q..\E`

hallieswan (changes requested):
This is looking great! Some comments for you. Would you also be willing to add tests for the new endpoint (e.g. service, delegate, etc...)? It will be a big improvement over the test coverage for the old route (i.e. none)

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
rrchai: Ah, that's a bummer. Bit of context on why I didn't catch it either: AWS's table isn't always reliable in both directions: `$getField` is listed as unsupported but works fine on DocDB 8. I hadn't tested `$unionWith` against DocDB until now though, and you're right: it fails with an error saying it's not supported. I will rework with one aggregation per collection merged in Java as you suggested - it does look simpler.

rrchai: Reworked in next commit(s) as you suggested: one aggregation per collection that projects precedence, then merged and sorted in backend.

libs/model-ad/ui/src/lib/components/search-input/search-input.component.ts
rrchai: Good point, this will be address in next commit(s).

libs/model-ad/ui/src/lib/components/search-input/search-input.component.ts
rrchai: Good point, this will be address in next commit(s).

libs/model-ad/api-description/src/paths/search/models.yaml
rrchai: Good point, this will be address in next commit(s).

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/service/ModelSearchService.java
rrchai: Good point, the constraints will be moved to schema and regenerated clients, as well as  `ConstraintViolationException` handler so the violation returns 400 of a 500 in next commit(s).

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/api/SearchApiDelegateImpl.java
rrchai: Valid points for normalizing them to share one cache key. Will refactor them in next commit(s).

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
rrchai: Good point that the AWS $regexMatch page doesn't mention the pattern. [Pattern.quote()](https://github.com/openjdk/jdk/blob/jdk-21-ga/src/java.base/share/classes/java/util/regex/Pattern.java#L1501), which wraps the string in `\Q…\E`, is already used by comparison-tool search in `GET /comparison-tools/model-overview`, just on a different operator ([$regex](https://docs.aws.amazon.com/documentdb/latest/devguide/regex.html)). I expect `$regexMatch` to behave the same way - but we could not verify until it's in DocDb?

libs/model-ad/ui/src/lib/components/search-input/search-input.component.stories.ts
```diff
@@ -29,11 +39,22 @@ export const HeaderSearchInput: Story = {
...
   args: {
     searchPlaceholder: 'Find mouse model by name or ID...',
     searchImagePath: 'model-ad-assets/images/mouse-head.svg',
     searchImageAltText: 'mouse head icon',
     hasThickBorder: true,
+    modelOrganisms: [ModelOrganism.Mouse],
+  },
+};
+
+export const MarmosetHomeSearchInput: Story = {
```
rrchai, line 52: Split the story into `MouseHomeSearchInput` and `MarmosetHomeSearchInput`, one per organism. Let me know if you'd prefer a single story with an organism toggle instead.

libs/model-ad/home/src/lib/home.component.html
```diff
@@ -72,13 +73,20 @@ <h1 class="home-heading">Find the right model for your research</h1>
...
             <explorers-home-card
               [title]="'Marmoset Model Search'"
               [description]="
                 'Search marmoset models to explore <strong>Plasma Biomarker</strong> results and detailed model information.'
               "
-            />
+            >
+              <model-ad-search-input
+                [searchPlaceholder]="'Find marmoset model by name...'"
+                [searchImagePath]="'model-ad-assets/images/marmoset-model.svg'"
```
hallieswan: I missed this in my initial review, but it looks like [the design](https://www.figma.com/design/GtUYqZBxoOnPtjV1hLcTgw/Model-AD-Explorer-Designs?node-id=4531-3615&m=dev) uses the search result image here, rather than the toggle header image: 


```suggestion
                [searchImagePath]="'model-ad-assets/images/marmoset-head.svg'"
```

libs/model-ad/ui/src/lib/components/search-input/search-input.component.stories.ts
```diff
@@ -29,11 +39,22 @@ export const HeaderSearchInput: Story = {
...
+  args: {
+    searchPlaceholder: 'Find marmoset model by name...',
+    searchImagePath: 'model-ad-assets/images/marmoset-model.svg',
```
hallieswan: To align with the design, let's use the search result image here too: 


```suggestion
    searchImagePath: 'model-ad-assets/images/marmoset-head.svg',
```

libs/model-ad/ui/src/lib/components/search-input/search-input.component.spec.ts
```diff
@@ -72,6 +65,7 @@ describe('SearchInputComponent', () => {
       id: 'model5',
       match_field: 'unknown_field',
       match_value: 'some value',
+      model_organism: 'mouse',
```
hallieswan: Nit: should we use the `ModelOrganism` enum throughout this file instead of the string values?

apps/model-ad/app/e2e/search.spec.ts
```diff
@@ -85,6 +85,33 @@ test.describe('search', () => {
...
+    );
+    await expect(searchListItems.first()).toHaveText(model);
+  });
+
+  test('can search across organisms and navigate to marmoset model', async ({ page }) => {
+    const modelName = 'Presenilin 1';
+
+    await page.goto('/');
+
+    const { searchListItems } = await searchAndGetSearchListItems('presenilin', page);
```
hallieswan: Nit: what do you think about making this search term match mouse and marmoset models (e.g. `p`), so that this e2e test validates "search across organisms"?

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
hallieswan: Sounds good - can you add this as acceptance criteria to the ticket with an expected working example, so the validator makes sure to check that this behavior is working as expected in the deployed stack?

libs/model-ad/ui/src/lib/components/search-input/search-input.component.ts
```diff
@@ -45,7 +43,7 @@ export class SearchInputComponent {
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   checkQueryForErrors = (query: string): string => {
-    return '';
+    return ''; // empty string if no error
   };
```
hallieswan, line 47: Currently the character limit is only enforced on the backend, so we'd get an unhelpful error message for a too-long search term from the explorers-search-input (i.e. "An unknown error occurred, please try again."). We could consider adding validation of the query length here (or surface the actual server error message in explorers, so Agora benefits too)? Can also be deferred to a future ticket, if that's preferred

libs/model-ad/home/src/lib/home.component.html
rrchai: Good catch! Will fix in next commit(s).

libs/model-ad/ui/src/lib/components/search-input/search-input.component.stories.ts
rrchai: yes, will fix in next commit(s).

libs/model-ad/ui/src/lib/components/search-input/search-input.component.spec.ts
rrchai: you're right that it's better to use typed `ModelOrganism` here. Will fix in next commit(s).

apps/model-ad/app/e2e/search.spec.ts
rrchai: Yea, this is a good point to have a better example to validate "search across organisms". I will use `il` with fewer outputs example in next commit(s).

```console
 curl -s "http://localhost:8000/api/v1/search/models?q=il" | jq
[
  {
    "id": "LOAD1.Il1rapExon2KO",
    "match_field": "name",
    "match_value": "LOAD1.Il1rapExon2KO",
    "model_organism": "mouse"
  },
  {
    "id": "Presenilin 1",
    "match_field": "name",
    "match_value": "Presenilin 1",
    "model_organism": "marmoset"
  }
]
```

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
rrchai: Good point, added https://sagebionetworks.jira.com/browse/MG-537

libs/model-ad/ui/src/lib/components/search-input/search-input.component.ts
rrchai: Good point. This is pre-existing behavior on the error message for long search term. I'd like to defer it to a ticket to decide whether we show a frontend message with a frontend constraint, or surface the backend message.

rrchai: [ticket here](https://sagebionetworks.jira.com/browse/MG-1077)

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/exception/GlobalExceptionHandler.java
```diff
@@ -138,6 +140,45 @@ protected ResponseEntity<BasicErrorDto> handleIllegalArgument(
...
+        String propertyPath = violation.getPropertyPath().toString();
+        String parameter = propertyPath.substring(propertyPath.lastIndexOf('.') + 1);
+        if (violation.getConstraintDescriptor().getAnnotation() instanceof Size size) {
+          return String.format(
+            "Query parameter %s must be between %d and %d characters",
+            parameter,
+            size.min(),
+            size.max()
+          );
+        }
```
hallieswan: `size` could apply to more than just string length, e.g. `categories` in the [DiseaseCorrelationSearchQuery](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/model-ad/api-description/src/components/schemas/DiseaseCorrelationSearchQuery.yaml#L19-L29) has an array size of 2. What do you think about adding a `violation.getInvalidValue() instanceof CharSequence` here to distinguish them? We can add extra constraint error handling here for arrays etc.. in the future if desired

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/exception/GlobalExceptionHandler.java
```diff
@@ -138,6 +140,45 @@ protected ResponseEntity<BasicErrorDto> handleIllegalArgument(
...
+  @ExceptionHandler(ConstraintViolationException.class)
+  protected ResponseEntity<BasicErrorDto> handleConstraintViolation(
+    ConstraintViolationException ex,
+    NativeWebRequest request,
+    Locale locale
+  ) {
+    String detail = ex
+      .getConstraintViolations()
+      .stream()
+      .findFirst()
```
hallieswan: Nit: consider joining all violations on the request, rather than choosing the first reported or sorting the violations so the first reported is consistent (since the sort order is arbitrary by default)

apps/model-ad/api-next/src/test/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImplTest.java
```diff
@@ -0,0 +1,211 @@
...
+    // Abca7 matched on jax_id (precedence 3), Trem2R47H matched on name (precedence 1)
+    stubAggregation(
+      MOUSE_COLLECTION,
+      mouseResults,
+      List.of(document("Abca7", 3), document("Trem2R47H", 1))
+    );
+    stubAggregation(
+      MARMOSET_COLLECTION,
+      marmosetResults,
+      List.of(document("Marmoset2", 1), document("Marmoset1", 1))
```
hallieswan: Nit: should we use a model name that will sort after Trem2R47H, so it's clear that marmoset and mouse results are fully interleaved? e.g. output would be Marmoset1, Trem2R47h, ZZMarmoset, Abca7

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
```diff
@@ -0,0 +1,144 @@
...
+  private List<Document> mouseMatchBranches(String escapedQuery) {
+    return Arrays.asList(
+        new Document()
+            .append("case", new Document("$regexMatch", new Document()
+                .append("input", "$name").append("regex", escapedQuery)
+                .append("options", "i")))
+            .append("then", new Document()
+                .append("precedence", 1)
+                .append("match_field", "name")
+                .append("match_value", "$name")),
```
hallieswan: Nit: the match branches are pretty similar -- could consider extracting a helper to build them, so it's easier to pick up differences (e.g. field, precedence, etc...)

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
```diff
@@ -0,0 +1,144 @@
...
+    return context -> new Document("$project", new Document()
+        .append("_id", "$name")
+        .append("match_field", "$match_info.match_field")
+        .append("match_value", "$match_info.match_value")
+        .append("model_organism", "$model_organism")
+        .append("precedence", "$match_info.precedence"));
+  }
+
+  private Document buildVars(String escapedQuery) {
+    return new Document("filtered_aliases", new Document("$filter", new Document()
```
hallieswan, line 96: Nit: `buildVars` runs for both collections, but the marmoset branch doesn't use `filtered_aliases` -- could consider making`$let` conditional per organism

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
```diff
@@ -0,0 +1,144 @@
...
+@Slf4j
+public class ModelSearchRepositoryImpl implements ModelSearchRepository {
+
+  private static final String MOUSE_COLLECTION = "model_details";
+  private static final String MARMOSET_COLLECTION = "marmo_details";
+
+  private final MongoTemplate mongoTemplate;
+
+  @Override
+  public List<SearchResultDocument> searchModels(String query, List<ModelOrganismDto> organisms) {
```
hallieswan, line 30: What do you think about adding a couple comments to this method to document decisions? e.g. 

1. `organisms` is expected to not be null or an empty list? `SearchApiDelegateImpl` handles normalizing these values, so this method works as expected, but may be worth explicitly stating here too
2. `$unionWith` is not used because it isn't support in DocDB

apps/model-ad/api-next/src/test/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImplTest.java
```diff
@@ -0,0 +1,211 @@
...
+      anyString(),
+      eq(SearchResultDocument.class)
+    );
+
+    assertThat(aggregationCaptor.getAllValues()).allSatisfy(aggregation ->
+      assertThat(aggregation.toString())
+        .as("$unionWith is unsupported on DocumentDB, so the collections are merged in Java")
+        .doesNotContain("$unionWith")
+    );
+  }
```
hallieswan: What do you think about removing this test in favor of documenting this constraint with a comment in the `searchModels` method?

hallieswan (commented):

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/exception/GlobalExceptionHandler.java
rrchai: sounds good! will add it in next commit(s)

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/exception/GlobalExceptionHandler.java
rrchai: That makes sense - will apply in next commit(s)

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
rrchai: Sure, I was assuming marmoset would get aliases like mouse, although it's omitted currently.  Makes sense to only apply the $let when the field is actually added to the marmoset schema. I'll add a comment so it's clear the $let needs to come back together if we add alias in the future for marmoset.

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/ModelSearchRepositoryImpl.java
rrchai: good point, I will refactor to simplify in next commit(s).

hallieswan (approved):
Looks good to me!

PR #4168, opened 2026-08-20
fix(explorers): size header nav underline to the widest line of text (MG-1029)

sagely1:
## Description

The active-link underline in the shared explorers header spanned the full width of the link box. Once the viewport narrows enough to force a nav label to wrap onto two lines, that box is wider than any line of text, so the underline visibly overhangs the label. This sizes the underline to the widest rendered line instead, which is the behavior MG-1029 asks for. The header is shared across Agora, Model-AD, and QTL, so the fix lands in `libs/explorers` and benefits all three.  Also, the 'Model Overview' nav link is center justified but the other nav links are left-justified so this PR also fixes this inconsistency.

A CSS-only fix isn't possible here: shrink-to-fit width (`fit-content`, `inline-block`, `display: table`, `float`) is resolved from intrinsic sizes *before* line breaking, so a wrapped box stays as wide as the space available and never collapses to its longest line. The widest line box only exists after layout, hence the small measuring directive.

## Related Issue

- [MG-1029](https://sagebionetworks.jira.com/browse/MG-1029) — Header nav element underline is too wide when text wraps

## Changelog

- Add `WidestLineWidthDirective` to `explorers-util`, exposing the host's widest rendered text line as the `--widest-line-width` custom property
- Size the header's active-link underline to `--widest-line-width` instead of the full link box
- Apply the directive to header nav links, external links, and dropdown triggers
- Left-align the dropdown trigger's label, which the `reset-button` mixin left centered by the UA stylesheet
- Export the new directive from the `explorers-util` public API

## Testing

- Open any explorer (Model-AD at `/comparison/model` puts "Model Overview" in the active state) and narrow the browser to roughly 1300–1400px so nav labels wrap onto two lines.
- Confirm the active link's underline matches the widest line of its label rather than overhanging it, and that it starts flush with the left edge of the text.
- Confirm the "Model Overview" dropdown trigger wraps left-aligned, matching the plain `<a>` nav links beside it.
- Widen the browser until labels fit on one line; the underline should still span the full label.
- Resize slowly across the wrap threshold and confirm the underline re-measures rather than sticking at a stale width.
- Narrow below 1300px into the mobile menu and confirm the stacked nav links still show their full-width underline.
- Spot-check the Agora and QTL headers, since the component is shared.

### Preview

| Before                                                                                                                        | After                                                                                                                       |
| ----------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------- |
| image | image |

[MG-1029]: https://sagebionetworks.jira.com/browse/MG-1029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MG-945]: https://sagebionetworks.jira.com/browse/MG-945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/util/src/lib/widest-line-width/widest-line-width.directive.spec.ts
+++ b/libs/explorers/util/src/lib/widest-line-width/widest-line-width.directive.spec.ts
@@ -1,0 +1,71 @@
+import { ApplicationRef, Component } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { render, screen } from '@testing-library/angular';
+import {
+  WIDEST_LINE_WIDTH_PROPERTY,
+  WidestLineWidthDirective,
+} from './widest-line-width.directive';
+
+const LINE_WIDTHS = [98.4, 100.2];
+const WIDEST_LINE_WIDTH = '101px';
+
+@Component({
+  imports: [WidestLineWidthDirective],
+  template: `<a explorersWidestLineWidth data-testid="link"><span>Disease Correlation</span></a>`,
+})
+class TestHostComponent {}
+
+let notifyResize: ResizeObserverCallback | undefined;
+const disconnect = jest.fn();
+
+class ResizeObserverStub {
+  constructor(callback: ResizeObserverCallback) {
+    notifyResize = callback;
+  }
+  observe = jest.fn();
+  unobserve = jest.fn();
+  disconnect = disconnect;
+}
+
+async function setup() {
+  const { fixture } = await render(TestHostComponent);
+  // afterNextRender callbacks, which install the resize observer, only run on tick
+  TestBed.inject(ApplicationRef).tick();
+  return { fixture, link: screen.getByTestId('link') };
+}
+
+describe('WidestLineWidthDirective', () => {
+  beforeEach(() => {
+    notifyResize = undefined;
+    disconnect.mockClear();
+    global.ResizeObserver = ResizeObserverStub as unknown as typeof ResizeObserver;
+    Range.prototype.getClientRects = jest.fn(
+      () => LINE_WIDTHS.map((width) => ({ width })) as unknown as DOMRectList,
+    );
+  });
+
+  it('should set the widest line width as a custom property when the host resizes', async () => {
+    const { link } = await setup();
+
+    notifyResize?.([], {} as ResizeObserver);
+
+    expect(link.style.getPropertyValue(WIDEST_LINE_WIDTH_PROPERTY)).toBe(WIDEST_LINE_WIDTH);
+  });
+
+  it('should leave the custom property unset when the host renders no text lines', async () => {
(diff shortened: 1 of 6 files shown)


libs/explorers/util/src/lib/widest-line-width/widest-line-width.directive.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    global.ResizeObserver = ResizeObserverStub as unknown as typeof ResizeObserver;
+    Range.prototype.getClientRects = jest.fn(
+      () => LINE_WIDTHS.map((width) => ({ width })) as unknown as DOMRectList,
+    );
+  });
+
+  it('should set the widest line width as a custom property when the host resizes', async () => {
+    const { link } = await setup();
+
+    notifyResize?.([], {} as ResizeObserver);
```
rrchai: This passes even if notifyResize was never captured. Assert it's defined first?


```suggestion
    expect(notifyResize).toBeDefined();
    notifyResize?.([], {} as ResizeObserver);
```

rrchai (approved):
LGTM, just one suggestion

libs/explorers/util/src/lib/widest-line-width/widest-line-width.directive.ts
```diff
@@ -0,0 +1,54 @@
+import { afterNextRender, Directive, ElementRef, inject, OnDestroy } from '@angular/core';
+
+export const WIDEST_LINE_WIDTH_PROPERTY = '--widest-line-width';
```
hallieswan, line 5: Should we add a comment noting that we need to keep this value in sync with `libs/explorers/ui/src/lib/components/header/header.component.scss`, e.g. something similar to what we do for the [comparison tool](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.constants.ts#L11-L14)?

sagely1: Here the directive is the sole writer of a property that any stylesheet may read, so naming header.component.scss would go stale the moment a second consumer exists.  That said, I think I can make it clearer in the comment that this is a contract variable.  I'll add a comment to that effect.

libs/explorers/util/src/lib/widest-line-width/widest-line-width.directive.spec.ts
sagely1: I'll refactor this test.  Done in the next commit.

rrchai (approved):

hallieswan (approved):

PR #4169, opened 2026-08-20
feat(model-ad): match full ensembl gene ids in the differential expression filterbox search (MG-1048)

hallieswan:
## Description

Searching the Model-AD differential expression comparison tool for a full Ensembl gene ID returned nothing whenever the matching rows also had a `gene_symbol`, because the filterbox only fell back to `ensembl_gene_id` for rows where `gene_symbol` was blank. Users who paste an Ensembl ID (the identifier shown in the table for genes with no symbol, and the one they carry over from other tools) had no way to find those rows. This routes each search term to the field it actually identifies: a term matching `^ENSMUSG\d{11}$` searches `ensembl_gene_id` only, while every other term keeps the existing `gene_symbol`-with-`ensembl_gene_id`-fallback behavior. Routing happens per term, so a comma-separated list can mix gene symbols and Ensembl IDs.

## Related Issue

- [MG-1048](https://sagebionetworks.jira.com/browse/MG-1048)

## Changelog

- Route full Ensembl mouse gene ID search terms to `ensembl_gene_id` in the Model-AD transcriptomics filterbox search, case-insensitively
- Route each term of a comma-separated search independently, so gene symbols and Ensembl IDs can be mixed in one query
- Preserve the existing fallback behavior for every other term, including partial Ensembl IDs
- Add `ApiHelper.splitSearchTerms` and a `List<String>` overload of `createCaseInsensitiveFullMatchPatterns` so callers can inspect terms before building patterns
- Rename the shared e2e helper `testPartialCaseInsensitiveSearch` to `expectSearchResults` and update its call sites across the Agora and Model-AD specs, since the new tests use it for full-match searches and the old name no longer described what it asserts
- Add unit tests for the new search routing and helper methods, plus 6 e2e tests covering the ticket's cases

## Testing

Navigate to the Model-AD differential expression comparison tool at `/comparison/expression?categories=RNA%2520-%2520DIFFERENTIAL%2520EXPRESSION,Tissue%2520-%2520Hemibrain`, then:

- Search `ENSMUSG00000038619` and confirm the `Ensa` rows are returned (this returned no rows before)
- Search `ensmusg00000038619` and confirm the same rows are returned, verifying case-insensitivity
- Search `ENSMUSG00000038619,plec` and confirm both the `Ensa` and `Plec` rows are returned
- Search `ENSMUSG00000037982` (a gene with no symbol) and confirm its rows are returned
- Search `ens` and confirm the partial-match fallback still holds: rows with a gene symbol match on the symbol (`Ensa`) and rows without one match on the Ensembl ID (`ENSMUSG00000037982`)
- Search a plain gene symbol such as `plec` and confirm nothing about the pre-existing partial, case-insensitive search changed

### Preview

Expectations for each search are described in the testing section.

search | dev | this pr 
:---: | :---: | :---: 
`ENSMUSG00000038619` | image | image
`ensmusg00000038619` | image | image
`ENSMUSG00000038619,plec` | image | image
`ENSMUSG00000037982` | image | image
`ens` | image | image
`plec` | image | image

[MG-1048]: https://sagebionetworks.jira.com/browse/MG-1048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/CustomTranscriptomicsRepositoryImpl.java
+++ b/apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/CustomTranscriptomicsRepositoryImpl.java
@@ -46,7 +46,7 @@ public class CustomTranscriptomicsRepositoryImpl
    * this filterbox is case-insensitive; without it a lower-cased ID would fall through to the
    * gene_symbol path and return nothing.
    */
-  private static final Pattern FULL_ENSEMBL_GENE_ID = Pattern.compile(
+  private static final Pattern FULL_MOUSE_ENSEMBL_GENE_ID = Pattern.compile(
     "^ENSMUSG\\d{11}$",
     Pattern.CASE_INSENSITIVE
   );
@@ -257,7 +257,7 @@ public class CustomTranscriptomicsRepositoryImpl
   }
 
   private static boolean isFullEnsemblGeneId(String term) {
-    return FULL_ENSEMBL_GENE_ID.matcher(term).matches();
+    return FULL_MOUSE_ENSEMBL_GENE_ID.matcher(term).matches();
   }
 
   private static Criteria equalsAnyIgnoringCase(String field, List<String> terms) {
--- a/apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/CustomTranscriptomicsRepositoryImpl.java
+++ b/apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/CustomTranscriptomicsRepositoryImpl.java
@@ -38,7 +38,18 @@ public class CustomTranscriptomicsRepositoryImpl
 
   private static final String COLLECTION_NAME = "rna_de_aggregate";
   private static final String DISPLAY_GENE_SYMBOL_FIELD = "display_gene_symbol";
+  private static final String ENSEMBL_GENE_ID_FIELD = "ensembl_gene_id";
   private static final String GENE_SYMBOL_FIELD = "gene_symbol";
+
+  /**
+   * A complete Ensembl mouse gene ID. Matched case-insensitively because every other matcher in
+   * this filterbox is case-insensitive; without it a lower-cased ID would fall through to the
+   * gene_symbol path and return nothing.
+   */
+  private static final Pattern FULL_ENSEMBL_GENE_ID = Pattern.compile(
+    "^ENSMUSG\\d{11}$",
+    Pattern.CASE_INSENSITIVE
+  );
 
   public CustomTranscriptomicsRepositoryImpl(MongoTemplate mongoTemplate) {
     super(mongoTemplate);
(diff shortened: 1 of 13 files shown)


apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/CustomTranscriptomicsRepositoryImpl.java
```diff
@@ -38,8 +38,19 @@ public class CustomTranscriptomicsRepositoryImpl
...
   private static final String DISPLAY_GENE_SYMBOL_FIELD = "display_gene_symbol";
+  private static final String ENSEMBL_GENE_ID_FIELD = "ensembl_gene_id";
   private static final String GENE_SYMBOL_FIELD = "gene_symbol";
 
+  /**
+   * A complete Ensembl mouse gene ID. Matched case-insensitively because every other matcher in
+   * this filterbox is case-insensitive; without it a lower-cased ID would fall through to the
+   * gene_symbol path and return nothing.
+   */
+  private static final Pattern FULL_ENSEMBL_GENE_ID = Pattern.compile(
```
sagely1: would it make sense to call out the mouse pattern in the variable name?

`FULL_MOUSE_ENSEMBL_GENE_ID` instead of `FULL_ENSEMBL_GENE_ID` since we have marmoset as well?

apps/model-ad/api-next/src/test/java/org/sagebionetworks/model/ad/api/next/model/repository/CustomTranscriptomicsRepositoryImplTest.java
```diff
@@ -39,6 +40,13 @@
 class CustomTranscriptomicsRepositoryImplTest {
 
   private static final String COLLECTION_NAME = "rna_de_aggregate";
+  private static final String ENSEMBL_GENE_ID_FIELD = "ensembl_gene_id";
+  private static final String GENE_SYMBOL_FIELD = "gene_symbol";
+  private static final String TISSUE_FIELD = "tissue";
+
+  private static final String ENSA_ENSEMBL_GENE_ID = "ENSMUSG00000038619";
+  private static final String OTHER_ENSEMBL_GENE_ID = "ENSMUSG00000000001";
+  private static final String PLEC_GENE_SYMBOL = "plec";
```
sagely1, line 50: I'm not sure if this is the PR to do this in but would it be helpful to add a marmoset test id as well?

sagely1 (changes requested):
Looks great!  Just a few suggestions.

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/model/repository/CustomTranscriptomicsRepositoryImpl.java
hallieswan: Done in the next commit

apps/model-ad/api-next/src/test/java/org/sagebionetworks/model/ad/api/next/model/repository/CustomTranscriptomicsRepositoryImplTest.java
hallieswan: Done in the next commit

sagely1 (approved):
LGTM

PR #4170, opened 2026-08-21
feat(model-ad): update route info and header to specify mouse (MG-944)

sagely1:
## Description

Now that Model-AD hosts a Marmoset Model Overview comparison tool alongside the original mouse one, the generic "Model Overview" name and `comparison/model` route no longer make it clear which organism a user is looking at. This renames the mouse tool to "Mouse Model Overview" across the home page tile, the CT header, and the page metadata, and moves it to `comparison/model/mouse` to mirror the marmoset route. The old `comparison/model` URL still works — it redirects to the new path while preserving query params, so previously shared comparison tool links (with their filters, pins, and sorts) keep resolving. The mouse tool's "view details" links now also carry the `modelOrganism=mouse` query param so model details pages open in the correct organism context.

## Related Issue

- [MG-944](https://sagebionetworks.jira.com/browse/MG-944) — Update Mouse Model Overview CT frontend route info and header title

`ui_config` still names this page "Model Overview", so the header title is hardcoded and the config is still fetched with `ComparisonToolPage.ModelOverview`. Both spots carry a `TODO(MG-1057)` to switch over once the config is renamed.

## Changelog

- Move the mouse model overview route from `comparison/model` to `comparison/model/mouse`
- Add a `comparison/model` redirect that preserves query params and fragment via a `UrlTree` redirect function
- Rename the tool to "Mouse Model Overview" in the CT header, home page tile, and page title metadata
- Pass `modelOrganism=mouse` when opening a model details page from the mouse model overview table
- Add e2e coverage for the legacy URL redirect (with and without filter params) and for the view details link
- Update e2e constants and helpers to map the ui_config page name to the displayed header title

## Testing

- Visit `/comparison/model` and confirm the browser lands on `/comparison/model/mouse` with the page title "Mouse Model Overview | Overview of mouse models of Alzheimer's Disease".
- Apply a few filters on the mouse model overview, copy the URL, then replace `/comparison/model/mouse` with `/comparison/model` and reload — the redirect should keep every filter query param intact.
- From the home page with the mouse organism selected, confirm the first card reads "Mouse Model Overview" and navigates to `/comparison/model/mouse`.
- Confirm the CT page header on `/comparison/model/mouse` reads "Mouse Model Overview" and that the marmoset tool at `/comparison/model/marmoset` is unaffected.
- Click "view details" on a row in the mouse model overview table and confirm the new tab opens the model details page with `?modelOrganism=mouse` in the URL and mouse-specific content.

### Preview

Mouse Comparison Tool header (shows "Mouse")
| Before                                                                                                | After                                                                                                       |
| ----------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------- |
| image | image |

Hovering over the Mouse Comparison Tool tab
| Before                                                                                                | After                                                                                                       |
| ----------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------- |
| image | image | 

[MG-944]: https://sagebionetworks.jira.com/browse/MG-944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/model-ad/app/src/app/app.routes.ts
```diff
@@ -1,10 +1,13 @@
-import { ActivatedRouteSnapshot, Route } from '@angular/router';
+import { inject } from '@angular/core';
+import { ActivatedRouteSnapshot, Route, Router } from '@angular/router';
 import { SynapseWikiParams } from '@sagebionetworks/explorers/models';
 import { ROUTE_PATHS, SUPPORT_EMAIL } from '@sagebionetworks/model-ad/config';
 import { resolveModelOrganism } from '@sagebionetworks/model-ad/util';
 import { capitalizeFirstLetter } from '@sagebionetworks/shared/util';
 import { modelOrganismGuard } from './model-organism.guard';
 
+const LEGACY_MOUSE_MODEL_OVERVIEW_PATH = 'comparison/model';
```
hallieswan: Should this be colocated with the other routes in `libs/model-ad/config/src/lib/constants.ts`?

hallieswan: Alternately, we could just specify this path directly, instead of creating a constant for it -- that's how we've handled legacy paths in agora's [app.routes.ts](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/apps/agora/app/src/app/app.routes.ts#L107-L121)

libs/model-ad/home/src/lib/home.component.html
hallieswan: Should we update this to match[ the design](https://www.figma.com/design/GtUYqZBxoOnPtjV1hLcTgw/Model-AD-Explorer-Designs?node-id=4514-17032&m=dev) as well? 




```suggestion
              [title]="'Mouse Model Search'"
```

hallieswan: And [here](https://www.figma.com/design/GtUYqZBxoOnPtjV1hLcTgw/Model-AD-Explorer-Designs?node-id=4239-14737&m=dev) as well? 


```suggestion
              [title]="'Mouse Differential Expression'"
```

libs/model-ad/home/src/lib/home.component.spec.ts
```diff
@@ -51,7 +51,7 @@ describe('HomeComponent', () => {
   describe('species toggle', () => {
     it('should show mouse tiles by default', async () => {
       await setup();
-      expect(screen.getByText('Model Overview')).toBeInTheDocument();
+      expect(screen.getByText('Mouse Model Overview')).toBeInTheDocument();
       expect(screen.getByText('Differential Expression')).toBeInTheDocument();
```
hallieswan: If we update the card title, we'll need to update this test as well -- 


```suggestion
      expect(screen.getByText('Mouse Differential Expression')).toBeInTheDocument();
```

hallieswan (commented):

apps/model-ad/app/src/app/app.routes.ts
sagely1: I could hard code it and add a comment to ensure we capture that this is a legacy path.

hallieswan (approved):
Looks good to me!

PR #4172, opened 2026-08-24
feat(model-ad): highlight the targeted TOC link and section heading (MG-995)

hallieswan:
## Description

Clicking a table of contents link or landing on a model details page with a hash fragment scrolls to the right section, but nothing confirms which section was targeted. On pages with many similarly shaped boxplot headings, a reader who arrives mid-page has to re-read the heading to work out where they landed. This adds a transient emphasis to the targeted TOC link and section heading that holds for three seconds and then drops, so the destination is unmistakable without leaving permanent styling on the page.

## Related Issue

- [MG-995](https://sagebionetworks.jira.com/browse/MG-995)

## Changelog

- Highlight the targeted TOC link and section heading for three seconds on TOC clicks, hash fragments on load, and popstate navigation, covering both mouse section headings and marmoset age-group headings
- Restart the hold when a new anchor is targeted, and clear the pending timer when the component is destroyed
- Switch the TOC link's emphasis trigger from `:focus` to `:focus-visible` so a mouse click no longer leaves the link colored after the highlight drops

## Testing

- Open a mouse model details page such as `/models/3xTg-AD/biomarkers`, expand the table of contents, click a link, and confirm both the link and its section heading emphasize for about three seconds and then return to their normal styling
- Load an anchored URL directly, e.g. `/models/3xTg-AD/biomarkers#nfl`, and confirm the section heading highlights on arrival with no TOC interaction
- Repeat on a marmoset page, e.g. `/models/Presenilin%201/biomarkers?modelOrganism=marmoset#0-1-year`, and confirm the age-group heading highlights
- Click several TOC links in quick succession and confirm only the most recently clicked section is highlighted, and that it holds for a full three seconds measured from the last click rather than expiring early
- Tab to a TOC link and press Enter, then confirm the keyboard focus indicator still reads as a sensible focus affordance once the highlight has dropped

### Mouse model details page

On load: 

https://github.com/user-attachments/assets/0040ca9f-be06-4223-b478-38572364c44b

On clicking TOC link:

https://github.com/user-attachments/assets/bb54952b-78f0-43e3-bd12-3166974e00dd

### Marmoset model details page

On load: 

https://github.com/user-attachments/assets/4a8c93b8-5642-43b9-a3dd-e4ee6591c238

On clicking TOC link: 

https://github.com/user-attachments/assets/188a9bb8-ef73-439d-8541-ca4f0d52b1b3

[MG-995]: https://sagebionetworks.jira.com/browse/MG-995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/model-ad/app/e2e/helpers/model-details.ts
+++ b/apps/model-ad/app/e2e/helpers/model-details.ts
@@ -20,13 +20,16 @@ export function getTocLinks(page: Page) {
   return getTocContainer(page).getByTestId('toc-item-link');
 }
 
+// The highlight is applied after hydration and first paint, which can lag on a cold CI worker.
+const HIGHLIGHT_APPEAR_TIMEOUT_MS = 10000;
+
 // Must exceed ANCHOR_HIGHLIGHT_HOLD_MS in model-details-boxplots-selector.component.ts.
 const HIGHLIGHT_CLEAR_TIMEOUT_MS = 10000;
 
 // All the highlights are asserted before any is waited out, since they share a single hold timer.
 export async function expectHighlightClears(...locators: Locator[]) {
   for (const locator of locators) {
-    await expect(locator).toHaveClass(/highlighted/);
+    await expect(locator).toHaveClass(/highlighted/, { timeout: HIGHLIGHT_APPEAR_TIMEOUT_MS });
   }
   for (const locator of locators) {
     await expect(locator).not.toHaveClass(/highlighted/, { timeout: HIGHLIGHT_CLEAR_TIMEOUT_MS });
--- a/apps/model-ad/app/e2e/helpers/model-details.ts
+++ b/apps/model-ad/app/e2e/helpers/model-details.ts
@@ -1,4 +1,4 @@
-import { expect, Page, test } from '@playwright/test';
+import { expect, Locator, Page, test } from '@playwright/test';
 
 export function getTocContainer(page: Page) {
   return page.locator('.table-of-contents-container');
@@ -18,6 +18,19 @@ export function getTocList(page: Page) {
 
 export function getTocLinks(page: Page) {
   return getTocContainer(page).getByTestId('toc-item-link');
+}
+
+// Must exceed ANCHOR_HIGHLIGHT_HOLD_MS in model-details-boxplots-selector.component.ts.
+const HIGHLIGHT_CLEAR_TIMEOUT_MS = 10000;
+
+// All the highlights are asserted before any is waited out, since they share a single hold timer.
+export async function expectHighlightClears(...locators: Locator[]) {
+  for (const locator of locators) {
+    await expect(locator).toHaveClass(/highlighted/);
+  }
+  for (const locator of locators) {
+    await expect(locator).not.toHaveClass(/highlighted/, { timeout: HIGHLIGHT_CLEAR_TIMEOUT_MS });
+  }
 }
 
 // Expands the table of contents, then clicks every link and asserts it scrolls to its section.
--- a/libs/model-ad/model-details/src/lib/components/marmoset-model-details-boxplots-selector/marmoset-model-details-boxplots-selector.component.html
+++ b/libs/model-ad/model-details/src/lib/components/marmoset-model-details-boxplots-selector/marmoset-model-details-boxplots-selector.component.html
@@ -7,6 +7,7 @@
   let-setActiveShareLink="setActiveShareLink"
   let-clearActiveShareLink="clearActiveShareLink"
   let-isShareLinkActive="isShareLinkActive"
+  let-isHighlighted="isHighlighted"
 >
   <div
     class="marmoset-legend"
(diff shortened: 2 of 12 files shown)


libs/model-ad/model-details/src/lib/components/marmoset-model-details-boxplots-selector/marmoset-model-details-boxplots-selector.component.spec.ts
```diff
@@ -8,20 +9,26 @@ import {
...
+
 async function setup() {
   const { fixture } = await render(MarmosetModelDetailsBoxplotsSelectorComponent, {
     imports: [MockWikiComponent],
     componentInputs: {
       title: 'Plasma Biomarkers',
       modelName: 'Presenilin 1',
       modelDataList: marmosetModelDataMock,
-      wikiParams: validWikiParams,
+      wikiParams: validWikiParams[0],
```
hallieswan, line 27: Fix type error -- `wikiParams` expects a single `SynapseWikiParams` but `validWikiParams` is an array, so pass `validWikiParams[0]` instead

libs/model-ad/model-details/src/lib/components/marmoset-model-details-boxplots-selector/marmoset-model-details-boxplots-selector.component.html
```diff
@@ -24,7 +25,7 @@
         (mouseleave)="clearActiveShareLink()"
       >
         <div class="age-group-title">
-          <h3>{{ ageGroup.age }}</h3>
+          <h3 [class.highlighted]="isHighlighted(anchorId)">{{ ageGroup.age }}</h3>
```
sagely1, line 28: what do you think about adding a tabindex here?  currently, tabbing after clicking on any item on the TOC sends the focus back up to the TOC instead of to the content.

apps/model-ad/app/e2e/helpers/model-details.ts
```diff
@@ -20,6 +20,19 @@ export function getTocLinks(page: Page) {
   return getTocContainer(page).getByTestId('toc-item-link');
 }
 
+// Must exceed ANCHOR_HIGHLIGHT_HOLD_MS in model-details-boxplots-selector.component.ts.
+const HIGHLIGHT_CLEAR_TIMEOUT_MS = 10000;
+
+// All the highlights are asserted before any is waited out, since they share a single hold timer.
+export async function expectHighlightClears(...locators: Locator[]) {
+  for (const locator of locators) {
+    await expect(locator).toHaveClass(/highlighted/);
```
sagely1: To make this potentially less flaky, we may want to specify another constant like below (suggesting 30s) as the highlight would occur after hydration and first paint.

We may want to do something like this:
```
const HIGHLIGHT_APPEAR_TIMEOUT_MS = 30000; // allow time for first paint and hydration
await expect(locator).toHaveClass(/highlighted/, { timeout: HIGHLIGHT_APPEAR_TIMEOUT_MS }); 
```

sagely1 (changes requested):

apps/model-ad/app/e2e/helpers/model-details.ts
hallieswan: Added in the next commit -- used 10s instead of 30s to match similar waits in other e2e tests, and so that real failures are reported more quickly

libs/model-ad/model-details/src/lib/components/marmoset-model-details-boxplots-selector/marmoset-model-details-boxplots-selector.component.html
hallieswan: Good question! The TOC is sticky, so the focused button stays visible and tabbing on through the TOC list seems reasonable for a persistent nav. Moving focus into the plots would mean tabbing back to reach the TOC list that's still on screen. Happy to explore in a follow-up ticket if you'd like to file one!

sagely1: With html anchor links, the focus would move to the anchored element and that would also be ideal for accessibility and screen readers but I'll defer to you on final judgment.  We could consider adding a note in the description of the tabbing behavior.

sagely1 (approved):
LGTM

libs/model-ad/model-details/src/lib/components/marmoset-model-details-boxplots-selector/marmoset-model-details-boxplots-selector.component.html
hallieswan: Good call! I opened a new ticket and will address in a follow-up PR (along with two bugs I found) -- https://sagebionetworks.jira.com/browse/MG-1076

PR #4175, opened 2026-08-28
feat(model-ad): refactor routing (MG-934)

sagely1:
## Description

The Differential Expression comparison tool serves several datasets from a single route, with the
active one determined by the `categories` query param (one value per dropdown level: modality, then
tissue). The header previously offered a single flat "Differential Expression" link, so users always
landed on whatever selection the CT defaulted to and had to discover and change the modality dropdown
themselves. This PR splits that entry into explicit "RNA - Differential Expression" and
"Protein - Differential Expression" dropdown links that pre-seed the CT selection, which required refactoring
the shared explorers header to carry query params on navigation links.

## Related Issue

- [MG-934](https://sagebionetworks.jira.com/browse/MG-934)
  - Related: [MG-1046](https://sagebionetworks.jira.com/browse/MG-1046) — protein differential
    expression support; the Protein header link errors until it ships

## Changelog

- Add `queryParams` to `NavigationLink` and forward it through the explorers header, for desktop
  dropdown menu items and mobile inline links alike
- Add query-param-aware active-link highlighting to the explorers header (`isLinkActive`,
  `urlValuesStartWith`, `resolvePathMatchOptions`) so a link stays active when its declared param
  values are a prefix of the URL's — e.g. `categories=RNA` matches `categories=RNA,Tissue`
- Extract comma-separated query-param serialization into shared `stringifyCommaSeparatedQueryParam`
  and `parseCommaSeparatedQueryParam` helpers (with unit tests) in `@sagebionetworks/shared/util`, and
  refactor `ComparisonToolUrlService` to use them instead of its private, inlined logic
- Replace the Model-AD "Differential Expression" header link with RNA and Protein entries nested
  under a "Mouse Models" subheader; each link pins only its `categories` value while the CT resolves
  the remaining levels (tissue, …) from `ui_config`
- Add `DIFFERENTIAL_EXPRESSION_CATEGORIES` (and its `DifferentialExpressionCategory` type) to
  `model-ad/config` as the single source for the category values used by the nav; delete the old
  `differential-expression-categories.ts` module and update the DE component to consume the config
  constants
- Extract a shared `navigateViaHeaderNav` e2e helper and `HeaderNavTrail` type, replacing the
  duplicated hamburger/dropdown click logic in the Agora and Model-AD comparison tool helpers
- Split the e2e sort helpers into their own `comparison-tool-sort.ts` module, add a shared
  `ComparisonToolSort` type, and type the e2e constant maps by the generated `ComparisonToolPage` /
  `ComparisonToolConfigPage` unions
- Add e2e coverage for header-driven navigation to the RNA and Protein views, and for reaching
  Agora's nominated targets and drugs through the "Nominations" dropdown; update the header spec,
  story, and navigation link mocks to cover query-param links

## Testing

- Serve the Model-AD app and open the "Differential Expression" header dropdown; confirm it shows a
  "Mouse Models" subheader with "RNA - Differential Expression" and "Protein - Differential
  Expression" beneath it
- Click the RNA link and confirm the URL carries
  `categories=RNA - DIFFERENTIAL EXPRESSION,Tissue - Hemibrain` (URL-encoded) and that both dropdown
  selectors on the page reflect that selection
- Navigate to `/comparison/expression?categories=RNA%20-%20DIFFERENTIAL%20EXPRESSION,Tissue%20-%20Hippocampus`
  first, then use the RNA header link; the tissue must switch to Hemibrain, proving the link's query
  params are applied rather than the CT keeping or defaulting its own selection
- Click the Protein link and confirm the URL and category selector switch to
  `PROTEIN - DIFFERENTIAL EXPRESSION`; the table failing to load is expected until MG-1046
- Shrink to a mobile breakpoint, open the hamburger menu, and confirm both links appear under the
  subheader and still navigate with their query params
- In Agora, confirm the "Nominations" dropdown still navigates to Nominated Targets and Nominated
  Drugs, since its e2e navigation helper was rewritten on the shared implementation

### Preview
Desktop:
| Before                                                                                                                | After                                                                                                                                        |
| --------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
| image        | image |

Mobile:
| Before                                                                                                                | After                                                                                                                                        |
| --------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
| image | image  |

[MG-934]: https://sagebionetworks.jira.com/browse/MG-934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MG-1046]: https://sagebionetworks.jira.com/browse/MG-1046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/model-ad/util/src/lib/navigation-links.ts
```diff
@@ -22,7 +32,27 @@ export const headerLinks: NavigationLink[] = [
...
+        children: [
+          {
+            label: 'RNA - Differential Expression',
+            routerLink: [ROUTE_PATHS.DIFFERENTIAL_EXPRESSION],
+            queryParams: differentialExpressionQueryParams(DIFFERENTIAL_EXPRESSION_CATEGORIES.RNA),
+          },
+          {
+            label: 'Protein - Differential Expression',
+            routerLink: [ROUTE_PATHS.DIFFERENTIAL_EXPRESSION],
+            // TODO(MG-1046): this route will error until MG-1046 is implemented.
```
hallieswan: Let's use MG-1042 here instead -- MG-1046 just adds the backend route, MG-1042 wires it up

apps/agora/app/e2e/helpers/comparison-tool.ts
```diff
@@ -26,20 +31,15 @@ export const navigateToComparison = async (
     const urlPath = queryParameters ? `${path}?${queryParameters}` : path;
     await page.goto(urlPath);
   } else {
-    // Open the hamburger menu if the button is visible (mobile breakpoint)
-    const menuButton = page.locator('.hamburger-menu-button');
-    if (await menuButton.isVisible().catch(() => false)) {
-      await menuButton.click();
-    }
-    await page.getByRole('link', { name: name }).click();
+    await navigateViaHeaderNav(page, COMPARISON_TOOL_NAV_TRAILS[name]);
```
hallieswan, line 34: Nice work aligning agora and model-ad's navigation for CTs!

apps/agora/app/e2e/nominated-drugs.spec.ts
```diff
@@ -46,6 +46,13 @@ test.describe('nominated drugs - comparison tool', () => {
 
   runFilterPanelTests(async (page) => navigateToComparison(page, CT_PAGE, true, 'url'));
 
+  // The header nests both comparison tools under a "Nominations" dropdown, which renders as a
+  // lazily-populated popup on desktop, so the link is only reachable once the trigger is opened.
```
hallieswan: Nit: could consider dropping this comment -- the dropdown behavior is already documented in `navigateViaHeaderNav`, so this comment would become stale if that behavior changes. The test title also covers this particular test's goal

apps/agora/app/e2e/nominated-targets.spec.ts
```diff
@@ -55,6 +55,13 @@ test.describe('nominated targets - comparison tool', () => {
 
   runFilterPanelTests(async (page) => navigateToComparison(page, CT_PAGE, true, 'url'));
 
+  // The header nests both comparison tools under a "Nominations" dropdown, which renders as a
+  // lazily-populated popup on desktop, so the link is only reachable once the trigger is opened.
```
hallieswan: Nit: consider dropping this comment too

apps/model-ad/app/e2e/constants.ts
```diff
@@ -10,29 +13,39 @@ export const LEGACY_MOUSE_MODEL_OVERVIEW_PATH = '/comparison/model';
...
+    link: 'Marmoset Models',
+  },
+  'Model Overview': {
+    dropdown: 'Model Overview',
+    link: 'Mouse Models',
+  },
+  'Differential Expression': {
+    dropdown: 'Differential Expression',
+    link: 'RNA - Differential Expression',
+  },
```
hallieswan, line 39: Nit: there is a comment explaining why the protein route is missing from this map, but I was a bit confused until I got to that file in the diff. Would you mind adding another comment here to add the protein DE HeaderNavTrail as part of MG-1042?

libs/model-ad/config/src/lib/constants.ts
```diff
@@ -28,6 +28,21 @@ export const ROUTE_PATHS = {
...
+  RNA: 'RNA - DIFFERENTIAL EXPRESSION',
+  PROTEIN: 'PROTEIN - DIFFERENTIAL EXPRESSION',
+} as const;
+
+export type DifferentialExpressionCategory =
+  (typeof DIFFERENTIAL_EXPRESSION_CATEGORIES)[keyof typeof DIFFERENTIAL_EXPRESSION_CATEGORIES];
+
+// Available tissues differ per category; Hemibrain is the only one offered for both, so the header
+// nav pins it regardless of which category is linked. Switching tissue is done from the CT itself.
+export const DIFFERENTIAL_EXPRESSION_DEFAULT_TISSUE = 'Tissue - Hemibrain';
```
hallieswan: Do we need to pin the tissue, or could the link specify just the category (i.e. RNA vs Protein)? `normalizeSelection` prefix-matches a partial `categories` array and fills the remaining levels from `ui_config`, so dropping the tissue would keep `ui_config` in control of the default tissue for each category, rather than duplicating (or drifting from) that choice in the header links

libs/model-ad/config/src/lib/constants.ts
```diff
@@ -28,6 +28,21 @@ export const ROUTE_PATHS = {
   ERROR: 'error',
 } as const;
 
+// Dropdown values served by the comparison tool config endpoint for the Differential Expression
+// page. A `categories` selection is one value per level, ordered outermost first: category, then
+// tissue.
+export const DIFFERENTIAL_EXPRESSION_CATEGORIES = {
+  RNA: 'RNA - DIFFERENTIAL EXPRESSION',
+  PROTEIN: 'PROTEIN - DIFFERENTIAL EXPRESSION',
```
hallieswan, line 36: Should we use the CT's query param encoded values here instead? Otherwise, the value won't match the current URL, so the "active" link styling will never apply. 

Even with that fixed, the link would only highlight on this exact category + tissue pair -- switching tissues within the CT changes the `categories` and stops the match. Properly highlighting these would need category-aware logic in the header, alongside `isDropdownActive`.

For example, when we're currently on the "Marmoset Model Overview" CT, the mobile header link is highlighted: 

Image

 but when we're on the RNA view of the DE CT, the mobile link is not highlighted: 

Image

libs/explorers/ui/src/lib/components/header/header.component.spec.ts
```diff
@@ -248,6 +248,34 @@ describe('HeaderComponent', () => {
...
+    changeWindowSize(DESKTOP_WIDTH);
+    const { component } = await setup();
+
+    const flatItems = component.dropdownMenuItems.get('DropdownLink');
+    expect(flatItems?.[1].routerLink).toEqual(['/child-link-2']);
+    expect(flatItems?.[1].queryParams).toEqual({ tab: 'overview' });
+
+    const grandchildren = component.dropdownMenuItems.get('DropdownWithSubheader')?.[0].items;
+    expect(grandchildren?.[1].routerLink).toEqual(['/sub-child-link-2']);
+    expect(grandchildren?.[1].queryParams).toEqual({ tab: 'details' });
```
hallieswan, line 262: What do you think about making the routes in the new tests identical, so they are only distinguished by their query params (to match the RNA vs Protein header links added in this PR)?

libs/explorers/ui/src/lib/components/header/header.component.spec.ts
```diff
@@ -248,6 +248,34 @@ describe('HeaderComponent', () => {
...
+      'href',
+      '/child-link-2?tab=overview',
+    );
+    expect(screen.getByRole('link', { name: 'SubChildLink2' })).toHaveAttribute(
+      'href',
+      '/sub-child-link-2?tab=details',
+    );
+  });
+
   it('should mark dropdown as active when a subheader grandchild route is active', async () => {
```
hallieswan, line 306: Should we add a similar test for a link with query parameters?

libs/explorers/ui/src/lib/components/header/header.component.html
```diff
@@ -79,6 +81,7 @@
                   } @else if (link.routerLink) {
                     <a
                       [routerLink]="link.routerLink"
+                      [queryParams]="link.queryParams"
```
hallieswan, line 84: Should we add a test that validates that query params are added for a top-level link?

hallieswan (changes requested):
This is looking good! A few questions for you

libs/explorers/testing/src/lib/e2e/comparison-tool.ts
```diff
@@ -445,6 +445,39 @@ export const closeVisualizationOverviewDialog = async (page: Page) => {
   });
 };
 
+/**
```
rrchai: nit: do you consider to move these in a new file like `header-nav.ts` instead. Nothing here is CT specific.

libs/model-ad/config/src/lib/constants.ts
sagely1: Good catch.  Fixed in next commit.

libs/explorers/ui/src/lib/components/header/header.component.spec.ts
sagely1: This is covered by this test:  
`describe('HeaderComponent active dropdown links in mobile mode', () => {
`

libs/explorers/testing/src/lib/e2e/comparison-tool.ts
sagely1: Good idea.  Done in next commit.

libs/model-ad/config/src/lib/constants.ts
```diff
@@ -28,6 +28,17 @@ export const ROUTE_PATHS = {
   ERROR: 'error',
 } as const;
 
+// Dropdown values served by the comparison tool config endpoint for the Differential Expression
+// page. A `categories` selection is one value per level, ordered outermost first: category, then
+// tissue.
```
hallieswan: Nit: should this comment be updated to indicate that this object only contains the main category values?

libs/explorers/services/src/lib/comparison-tool-url.service.ts
```diff
@@ -77,14 +78,18 @@ export class ComparisonToolUrlService {
...
     } else if (value !== undefined) {
       params[key] = null;
     }
   }
 
+  /** Inverse of `parseCommaSeparatedQueryParam`. Encoding each value before joining on commas is
+   * what preserves commas within individual values. */
+  private toCommaSeparatedQueryParam(values: string[]): string {
+    return values.map((value) => encodeURIComponent(value)).join(',');
+  }
```
hallieswan: What do you think about colocating this method with `parseCommaSeparatedQueryParam` in `libs/shared/typescript/util/src/lib/helpers.ts`, so a future dev doesn't miss that the "inverse" method already exists and rewrites it instead?

libs/shared/typescript/util/src/lib/helpers.ts
```diff
@@ -73,3 +73,37 @@ export function capitalizeFirstLetter(value: string | null | undefined): string
   if (!value) return '';
   return value.charAt(0).toUpperCase() + value.slice(1);
 }
+
+/**
+ * Parses a multi-value query param written as a comma-separated list of individually
+ * percent-encoded values, the inverse of `values.map(encodeURIComponent).join(',')`. Encoding each
```
hallieswan: If we move `toCommaSeparatedQueryParam` into this file, then we can reference that function by name instead of by its implementation

apps/model-ad/app/e2e/constants.ts
```diff
@@ -10,29 +13,46 @@ export const LEGACY_MOUSE_MODEL_OVERVIEW_PATH = '/comparison/model';
...
+  'Disease Correlation': {
+    link: 'Disease Correlation',
+  },
+};
+
+// 'Differential Expression' exposes two header sub-links; the map above uses RNA as the default.
+// This trail reaches the Protein sub-link.
+export const DIFFERENTIAL_EXPRESSION_PROTEIN_NAV_TRAIL: HeaderNavTrail = {
+  dropdown: 'Differential Expression',
+  link: 'Protein - Differential Expression',
```
hallieswan: Nit: not necessary, but could consider creating an object for both differential expression nav trails, then referencing the RNA path in the `COMPARISON_TOOL_NAV_TRAILS` object, so this object feels less offhand, e.g. something like: 

```
export const DIFFERENTIAL_EXPRESSION_NAV_TRAILS: Record<'RNA' | 'PROTEIN', HeaderNavTrail> = {
  RNA: { dropdown: 'Differential Expression', link: 'RNA - Differential Expression' },
  PROTEIN: { dropdown: 'Differential Expression', link: 'Protein - Differential Expression' },
};

// and COMPARISON_TOOL_NAV_TRAILS['Differential Expression'] = DIFFERENTIAL_EXPRESSION_NAV_TRAILS.RNA
```

apps/model-ad/app/e2e/differential-expression-comparison-tool.spec.ts
```diff
@@ -79,6 +83,50 @@ test.describe('differential expression', () => {
...
+
+    // The visualization overview dialog is only shown on the first visit to a comparison tool
+    await navigateToComparison(page, CT_PAGE, false, 'link');
+
+    await expectCategoriesParams(page, [rnaCategory]);
+    // Asserted via the category selectors rather than expectCategories, whose getByText matching
+    // also picks up the near-identically named header menu items.
+    for (const category of [rnaCategory, rnaDefaultTissue]) {
+      await expect(page.getByRole('combobox', { name: category, exact: true })).toBeVisible();
+    }
```
hallieswan: What do you think about making `expectCategories` more specific instead? It uses `getByText` to handle dropdowns with multiple options (rendered as combobox) and dropdowns with a single option (rendered as text). If we scope it to `explorers-comparison-tool-category-selectors` before calling `getByText`, I think it should avoid the header menu item conflict at the source. Both this test and the Protein view test can then just use `expectCategories`, and the Protein test can assert its tissue too

libs/explorers/ui/src/lib/components/header/header.component.html
```diff
@@ -79,6 +81,7 @@
...
                       routerLinkActive="active"
```
hallieswan: Just wanted to confirm that it's intended that we're still using `routerLinkActive` here instead of the new `isLinkActive` method?

libs/explorers/ui/src/lib/components/header/header.component.ts
```diff
@@ -70,18 +78,37 @@ export class HeaderComponent implements OnInit {
...
+    if (!link.routerLink) return false;
+    if (!this.router.isActive(link.routerLink.join('/'), PATH_MATCH_OPTIONS)) return false;
+
+    const linkParams = link.queryParams;
+    if (!linkParams) return true;
+
+    const urlParams = this.router.parseUrl(this.router.url).queryParams;
+    return Object.keys(linkParams).every((key) =>
+      this.isValuePrefix(
+        parseCommaSeparatedQueryParam(linkParams[key]),
```
hallieswan: I think `linkParams` is already an unencoded value (e.g. we use actual spaces when defining the RNA/protein query params in `navigation-links.ts`), so should we treat it as a single value here instead of running it through `parseCommaSeparatedQueryParam`? I think the split would cause a link value containing a comma to be inadvertently broken into multiple parts -- e.g. a Disease Correlation cluster like "Cell Cluster, NMD Consensus Cluster D" would become two values on the link side, but stay one on the URL side since it's encoded there, so I don't think it would ever match. We don't have a link like that today, but it seems like it would silently fail to highlight if we added one?

libs/explorers/ui/src/lib/components/header/header.component.ts
```diff
@@ -70,18 +78,37 @@ export class HeaderComponent implements OnInit {
...
+   * the same param in the URL. `routerLinkActive` can't express this: it compares whole raw param
+   * strings, so a link declaring one value never matches a URL carrying that value plus the ones
+   * the page filled in, and its comparison is also blind to percent-encoding differences.
+   */
+  isLinkActive(link: NavigationLink): boolean {
+    if (!link.routerLink) return false;
+    if (!this.router.isActive(link.routerLink.join('/'), PATH_MATCH_OPTIONS)) return false;
+
+    const linkParams = link.queryParams;
+    if (!linkParams) return true;
```
hallieswan, line 112: Just wanted to double check the behavior here -- a link with no query param returns true whenever the path matches, so a link without query params and a link with a query param on the same route would both come back active at the same time (e.g. the example set up by "NoParams" and "Overview" in the spec -- if the user lands on `/comparison-tool?tab=overview`, then both "Overview" and "NoParams" will be highlighted as active). 

If that's right, would you mind noting it in the method comment? Our current header links don't have that shape, so it doesn't affect anything, but it'd be good for future devs to see it was a deliberate decision

libs/explorers/ui/src/lib/components/header/header.component.spec.ts
hallieswan: Sorry for not being clear -- the test you mentioned does cover the mobile case, but this test is at the desktop width. Since mobile and desktop take different paths in the template, I think the desktop case here is still untested

hallieswan (commented):
Looking great! A few questions for you

libs/explorers/ui/src/lib/components/header/header.component.spec.ts
sagely1: Ah, my mistake.  I see what you mean.  I'll add a test to the next commit.

libs/shared/typescript/util/src/lib/helpers.ts
sagely1: Done in the next commit!  As part of this refactor, I've renamed this to stringifyCommaSeparatedQueryParam as this naming is more obvious that it is the inverse of `parseCommaSeparatedQueryParam`.

apps/model-ad/app/e2e/constants.ts
sagely1: Good idea.  I'll also reference the RNA entry to prevent drift.

libs/explorers/ui/src/lib/components/header/header.component.html
sagely1: It is a good time to refactor this.  Done in the next commit.

libs/explorers/ui/src/lib/components/header/header.component.ts
```diff
@@ -84,29 +89,66 @@ export class HeaderComponent implements OnInit {
...
+    );
+
+    // Every value the link declares must equal the URL value at the same position:
+    //   link ['RNA']          vs url ['RNA', 'Tissue'] -> true  (matches, extra url value ignored)
+    //   link ['RNA']          vs url ['Protein']       -> false (position 0 differs)
+    //   link ['RNA', 'Tissue'] vs url ['RNA']          -> false (url[1] is undefined, so it fails)
+    return linkValues.every((value, index) => value === urlValues[index]);
+  }
+
+  // Query params are matched manually via the prefix logic below, so path matching always ignores
```
hallieswan, line 142: Nit: looks like this comment might be stale -- can we refer to the method by name instead of by position? e.g. something like this: 

```
// Query params are matched manually via urlValuesStartWith, so path matching always ignores
```

hallieswan (approved):
Looks good to me! Just one small optional suggestion

hallieswan (approved):

PR #4176, opened 2026-08-28
feat(model-ad): make model details toc entries behave like links (MG-1073, MG-1075, MG-1076)

hallieswan:
## Description

Clicking a table of content entry on the model details pages scrolled the page to a section but since entries weren't built from real links, none of the behavior users expect of in-page navigation was available: no URL to open in a new tab or copy, no destination preview on hover, and no focus change for keyboard and screen reader users, who stayed behind in the TOC while the page scrolled away from them. This PR turns each entry into a genuine anchor pointing at the current page plus the section fragment, and moves focus to the target section heading on activation.

Real links are only correct if the current URL is, so this branch also fixes URL fidelity on model details URLs with no tab segment. Previously a URL such as `/models/Presenilin%201?modelOrganism=marmoset` dropped its query parameters and hash fragment while falling back to the default panel, which would have made every table of contents link on that page point at a stripped-down URL.

## Related Issue

- [MG-1073](https://sagebionetworks.jira.com/browse/MG-1073)
- [MG-1075](https://sagebionetworks.jira.com/browse/MG-1075)
- [MG-1076](https://sagebionetworks.jira.com/browse/MG-1076)

## Changelog

- Render table of contents entries as anchors whose `href` is the current page URL plus the section fragment, so they can be hovered, copied, and opened in a new tab
- Move focus to the target section heading whenever the page scrolls to a section -- on click and on load with a section fragment in the URL -- without cancelling the smooth scroll
- Leave modified clicks (cmd, ctrl, shift, alt) to the browser so the link can be opened in a new tab or window, or downloaded
- Preserve query parameters and hash fragments when a model details URL has no tab segment and falls back to the default panel
- Add e2e coverage for link `href` values, new-tab opening, and focus movement across the mouse and marmoset pages

## Testing

- On a mouse model, open `/models/Abca7*V1599M/biomarkers?modelOrganism=mouse`, expand the table of contents, and hover an entry -- the browser status bar should show the full current URL plus the section fragment, not the home page
- Cmd/ctrl-click an entry and confirm the new tab loads the same page scrolled to that section, with the original tab left where it was
- Click an entry normally and confirm the page scrolls smoothly (not a hard jump), the URL fragment updates, and the section heading and table of contents entry both highlight and then clear after a few seconds
- Immediately after that click, press Tab and confirm focus continues from inside the target section rather than back in the table of contents; with a screen reader on, confirm the section title is announced
- Paste a section URL such as `/models/Abca7*V1599M/biomarkers?modelOrganism=mouse#nfl` into a fresh tab, then press Tab -- focus should continue from inside that section rather than from the top of the page
- Set the Tissue and Sex filters away from their defaults, then copy an entry's link and open it in a new tab -- the filters and the section should both be restored
- Repeat the hover and new-tab checks on a marmoset model at `/models/Presenilin%201?modelOrganism=marmoset` (no tab segment) and confirm the age group links keep `?modelOrganism=marmoset` and land on the right age group

### Preview

TOC link can be opened in new tab, which scrolls to appropriate section: 

https://github.com/user-attachments/assets/20f98d57-c0e4-416a-96f2-c3426962929c

TOC link can be previewed and scrolls to appropriate section on click: 

https://github.com/user-attachments/assets/c86f1269-f2f7-43bf-958c-c818012fb156

TOC link focus moves to appropriate section after scroll: 

https://github.com/user-attachments/assets/d471d876-769f-4992-b8b7-1ee9a74c9aa7

Model details page without a tab segment respects the hash fragment: 

https://github.com/user-attachments/assets/ac504cb3-08d5-45cf-aaa3-765594995a63

Model details page without a tab segment respects the filter query parameter: 

https://github.com/user-attachments/assets/358eac9f-8dbb-4437-afd0-2814f9e2492c

[MG-1073]: https://sagebionetworks.jira.com/browse/MG-1073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/model-ad/model-details/src/lib/components/model-details-boxplots-selector/model-details-boxplots-selector.component.scss
+++ b/libs/model-ad/model-details/src/lib/components/model-details-boxplots-selector/model-details-boxplots-selector.component.scss
@@ -105,19 +105,13 @@
     li {
       padding-bottom: 10px;
 
-      button {
-        background: none;
-        border: none;
-        padding: 0;
+      a {
         color: var(--color-text);
         font-weight: 700;
         font-size: 18px;
         line-height: 20px;
-        cursor: pointer;
         text-decoration: underline;
 
-        // focus-visible rather than focus: a mouse click leaves the button focused, which
-        // would hold the emphasis color past the end of the highlight
         &:hover,
         &:focus-visible {
           color: var(--color-action-primary);
--- a/apps/model-ad/app/e2e/differential-expression-comparison-tool.spec.ts
+++ b/apps/model-ad/app/e2e/differential-expression-comparison-tool.spec.ts
@@ -399,7 +399,7 @@ test.describe('differential expression', () => {
 
   test('link for model with special characters opens model details page', async ({ page }) => {
     const specialModel = '5xFAD (IU/Jax/Pitt)';
-    const specialModelEncoded = '5xFAD%20%28IU%2FJax%2FPitt%29';
+    const specialModelEncoded = '5xFAD%20%2528IU%252FJax%252FPitt%2529';
 
     await navigateToComparison(page, CT_PAGE, true);
 
--- a/apps/model-ad/app/e2e/helpers/model-details.ts
+++ b/apps/model-ad/app/e2e/helpers/model-details.ts
@@ -17,7 +17,11 @@ export function getTocList(page: Page) {
 }
 
 export function getTocLinks(page: Page) {
-  return getTocContainer(page).getByTestId('toc-item-link');
+  return getTocContainer(page).getByRole('link');
+}
+
+export function getTocLink(page: Page, name: string) {
+  return getTocContainer(page).getByRole('link', { name, exact: true });
 }
 
 // The highlight is applied after hydration and first paint, which can lag on a cold CI worker.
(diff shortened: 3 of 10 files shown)


libs/model-ad/model-details/src/lib/components/model-details-boxplots-selector/model-details-boxplots-selector.component.scss
```diff
@@ -105,19 +105,13 @@
...
+      a {
         color: var(--color-text);
         font-weight: 700;
         font-size: 18px;
         line-height: 20px;
-        cursor: pointer;
         text-decoration: underline;
 
-        // focus-visible rather than focus: a mouse click leaves the button focused, which
-        // would hold the emphasis color past the end of the highlight
```
rrchai, line 120: should we keep this comment?

rrchai (approved):
LGTM

libs/model-ad/model-details/src/lib/components/model-details-boxplots-selector/model-details-boxplots-selector.component.scss
hallieswan: I don't think so! The focus now moves to the selected section on click, so the emphasis color won't extend past the end of the highlight

PR #4185, opened 2026-09-10
fix(model-ad): read model details route state from one snapshot (MG-1099)

rrchai:
## Description

Clicking a search result for a model of the opposite organism landed on the 404 page, even though the address bar showed the correct URL. The model details page read the model name and the modelOrganism query parameter as two separate router streams, which Angular advances one after the other within a single navigation — so the page paired the new organism with the previous model name and requested a model that does not exist. This fix reads both values from the router snapshot instead, and makes model requests cancellable so a superseded request can no longer redirect the user away from the page they asked for.

## Related Issue

- [MG-1099](https://sagebionetworks.jira.com/browse/MG-1099)

## Changelog

- Add `routeSnapshotChanges` to `@sagebionetworks/explorers/util`, which emits the router snapshot once
  per navigation so path parameters and query parameters are always read as one consistent unit
- Read model details route state through that helper instead of pairing `paramMap` and `queryParamMap`
  with `combineLatest`, eliminating the mismatched organism/name request
- Cancel in-flight model requests when a newer navigation starts, replacing an uncancelled nested
  `subscribe` with `switchMap`
- Handle a not-found model with `catchError` so the redirect no longer terminates the route
  subscription

## Testing

- On the marmoset **Presenilin 1** details page, type `LOAD1` in the header search and click the
  typeahead match. You should land on the mouse LOAD1 details page, and the network tab should show the correct LOAD1 model detail page (See Preview) with no request for `/api/v1/models/mouse/Presenilin%201`.
- Repeat in the other direction: from a mouse model details page, search a marmoset model and click the
  result.

### Preview

Cross-organism navigation works in both directions now 

#### From a mouse model page → marmoset search result

| Dev (before) | This PR (after) |
| --- | --- |
| dev-mouse-to-marmoset | pr-mouse-to-marmoset |

#### From a marmoset model page → mouse search result

| Dev (before) | This PR (after) |
| --- | --- |
| dev-marmoset-to-mouse | pr-marmoset-to-mouse |

[MG-1099]: https://sagebionetworks.jira.com/browse/MG-1099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/explorers/util/src/lib/route-state/route-state.ts
+++ b/libs/explorers/util/src/lib/route-state/route-state.ts
@@ -1,0 +1,16 @@
+import { ActivatedRoute, ActivatedRouteSnapshot } from '@angular/router';
+import { distinctUntilChanged, map, merge, Observable } from 'rxjs';
+
+/**
+ * Params and query params come in on two separate streams, and the router pushes one right after
+ * the other, so combining them can give you the new query param with the old path param. The
+ * snapshot is already updated before either push, so read from that and use the streams only as
+ * a trigger. Both pushes hand back the same snapshot object, so distinctUntilChanged drops the
+ * second one and we get a single emission per navigation.
+ */
+export function routeSnapshotChanges(route: ActivatedRoute): Observable<ActivatedRouteSnapshot> {
+  return merge(route.paramMap, route.queryParamMap).pipe(
+    map(() => route.snapshot),
+    distinctUntilChanged(),
+  );
+}
--- a/libs/model-ad/model-details/src/lib/model-details.component.spec.ts
+++ b/libs/model-ad/model-details/src/lib/model-details.component.spec.ts
@@ -1,14 +1,22 @@
-import { ActivatedRoute, convertToParamMap } from '@angular/router';
+import { Component } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { ActivatedRoute, convertToParamMap, Router, RouterOutlet } from '@angular/router';
 import { PlatformService } from '@sagebionetworks/explorers/services';
 import { provideLoadingIconColors } from '@sagebionetworks/explorers/testing';
 import { LoadingIconComponent } from '@sagebionetworks/explorers/util';
 import { Model, ModelOrganism, ModelService } from '@sagebionetworks/model-ad/api-client';
-import { MODEL_AD_LOADING_ICON_COLORS } from '@sagebionetworks/model-ad/config';
+import { MODEL_AD_LOADING_ICON_COLORS, ROUTE_PATHS } from '@sagebionetworks/model-ad/config';
 import { marmosetModelMock, mouseModelMock } from '@sagebionetworks/model-ad/testing';
 import { render, screen } from '@testing-library/angular';
 import userEvent from '@testing-library/user-event';
-import { of } from 'rxjs';
+import { of, throwError } from 'rxjs';
 import { ModelDetailsComponent } from './model-details.component';
+
+@Component({
+  selector: 'model-ad-not-found-stub',
+  template: 'Not found',
+})
+class NotFoundStubComponent {}
 
 async function setup(
   model: Model = mouseModelMock,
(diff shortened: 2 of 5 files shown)


libs/explorers/util/src/lib/route-state/route-state.ts
```diff
@@ -0,0 +1,16 @@
+import { ActivatedRoute, ActivatedRouteSnapshot } from '@angular/router';
```
rrchai, line 1: Placed in explorers so any page reading both a path param and a query param can reuse it, if needed

libs/model-ad/model-details/src/lib/model-details.component.ts
```diff
@@ -84,56 +84,56 @@ export class ModelDetailsComponent implements OnInit, AfterViewInit {
...
           ([prevParams, prevModelOrganism], [params, modelOrganism]) =>
             prevParams.get('name') === params.get('name') && prevModelOrganism === modelOrganism,
         ),
+        tap(() => this.reset()),
+        switchMap(([params, modelOrganism]) =>
+          // only fetch data during client hydration
+          this.platformService.isBrowser
+            ? this.loadPanelData(params, modelOrganism).pipe(map((model) => ({ model, params })))
+            : EMPTY,
+        ),
```
rrchai, line 103: `loadPanelData` is updated to return an observable now, so `switchMap` cancels an in-flight request when a newer navigation starts instead of letting the old one land late

libs/model-ad/model-details/src/lib/model-details.component.ts
```diff
@@ -84,56 +84,56 @@ export class ModelDetailsComponent implements OnInit, AfterViewInit {
...
-    combineLatest([this.route.paramMap, this.route.queryParamMap])
+    routeSnapshotChanges(this.route)
       .pipe(
-        map(([params, queryParams]): [ParamMap, ModelOrganism] => [
-          params,
-          resolveModelOrganism(queryParams.get(MODEL_ORGANISM_QUERY_KEY)),
+        map((snapshot): [ParamMap, ModelOrganism] => [
+          snapshot.paramMap,
+          resolveModelOrganism(snapshot.queryParamMap.get(MODEL_ORGANISM_QUERY_KEY)),
         ]),
```
rrchai, line 92: Reads the model name and organism from a single snapshot, so both values always belong to the same navigation

libs/model-ad/model-details/src/lib/model-details.component.ts
```diff
@@ -84,56 +84,56 @@ export class ModelDetailsComponent implements OnInit, AfterViewInit {
...
         distinctUntilChanged(
           ([prevParams, prevModelOrganism], [params, modelOrganism]) =>
             prevParams.get('name') === params.get('name') && prevModelOrganism === modelOrganism,
         ),
+        tap(() => this.reset()),
+        switchMap(([params, modelOrganism]) =>
+          // only fetch data during client hydration
+          this.platformService.isBrowser
+            ? this.loadPanelData(params, modelOrganism).pipe(map((model) => ({ model, params })))
+            : EMPTY,
```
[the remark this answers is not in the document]
rrchai: Valid, but out of scope for this ticket. it needs a click landing in a narrow window on a slow connection, for a model name that doesn't exist. if it ever shows up in practice, we can fix this preexisting gap then~

I will add a test for the not-found redirect though in next commit.

libs/explorers/util/src/lib/route-state/route-state.ts
hallieswan: Would you mind creating a follow-up ticket to track migrating model-ad-gene-details and model-ad-protein-details to use this as well? They have the same potential for this bug, although neither can actually hit it right now

libs/model-ad/model-details/src/lib/model-details.component.spec.ts
```diff
@@ -122,6 +134,80 @@ describe('ModelDetailsComponent', () => {
...
+        return notFoundDelayMs ? timer(notFoundDelayMs).pipe(switchMap(() => notFound)) : notFound;
+      });
+
+      const { navigate } = await render('<router-outlet></router-outlet>', {
+        imports: [RouterOutlet, LoadingIconComponent],
+        routes: [
+          { path: 'models/:name', component: ModelDetailsComponent },
+          { path: ROUTE_PATHS.NOT_FOUND, component: NotFoundStubComponent },
+        ],
+        initialRoute: `/models/${marmosetModelMock.name}?modelOrganism=marmoset`,
```
hallieswan: Nit: consider using `ROUTE_PATHS.MODELS` to construct this path (as we do for the not found path above)

hallieswan (approved):
Looks good to me! Just a couple small suggestions. Could also consider adding e2e tests to show that the routing works as expected on a real browser, but up to you

libs/model-ad/model-details/src/lib/model-details.component.spec.ts
rrchai: good catch, will switch to `ROUTE_PATHS.MODELS` in next commit.

hallieswan (approved):

PR #4187, opened 2026-09-10
feat(explorers): pin all matching rows across pages (MG-564, MG-565)

hallieswan:
## Description

The explorers CT only pinned the current page of matching rows when a user clicked "Pin All", so a search matching 40 rows across four pages pinned 10. This was different than Agora's GCT, which pins all matching rows across pages in a single click. This PR aligns the explorers CT with Agora's GCT, so that clicking "Pin All" pins every matching row across pages.

Pinning across pages means the client has to ask the server for matches it is not currently displaying, so this PR adds a `remainingBudget` query parameter to the comparison tool search queries and MG-564 uses it to implement Pin All. Because the budget makes it possible to reach the pin limit in one click, the limit is now enforced consistently wherever pins come from -- including a shared or hand-edited URL, which previously could restore more pins than a user is allowed to have.

## Related Issue

- [MG-564](https://sagebionetworks.jira.com/browse/MG-564)
- [MG-565](https://sagebionetworks.jira.com/browse/MG-565)

## Changelog

- Add `remainingBudget` to all seven explorers-based CT search query schemas (Agora nominated targets and drugs, Model-AD mouse and marmoset model overview, disease correlation, transcriptomics, proteomics), capped at each CT's pin limit, and regenerate the Java DTOs and Angular clients
- Emit a bare `$limit` instead of `$skip` / `$limit` in `ComparisonToolRepositorySupport` when a budget applies, return the result over `Pageable.unpaged` so the page shape does not misstate it, and keep the full match count so callers can detect truncation
- Wire `remainingBudget` through every delegate's `VALID_QUERY_PARAMS`, every service's `@Cacheable` key, and each repository's `findAll`
- Replace `ComparisonToolTableComponent.pinAll`'s current-page-only loop with `ComparisonToolService.pinAll()`, which fetches the matching ids server-side against the pins the user has left, warns when the budget truncated the result, toasts on failure, and balances its loading counter in `finalize` so an observable that completes without emitting cannot leave the table stuck loading
- Add `pinAllFetch` to `ComparisonToolService.connect()` and implement it in all six comparison tools by extracting each one's query builder into a reusable `buildUnpinnedQuery`, with the shared pagination-or-budget rule living in `ComparisonToolService.buildPaginationOrBudget`
- Make the pin limit an invariant of `pinnedData` no matter where pins come from: cap it in `setPinnedData` so URL-restored pins are trimmed to the first N by the user's current sort, derive `pinnedResultsCount` from it, clamp `setMaxPinnedItems` to `MAX_PINNED_ITEMS`, and make `setPinnedItems` skip a write matching the current pins so the pinned/unpinned refetch loop terminates
- Add e2e coverage for pin-all across pages, pin-all hitting the limit, and URL pins exceeding the limit on both Agora CTs and Model-AD
- Document the query-parameter wiring and the budgeted pipeline tail in `.claude/rules/ct-repository.md`

## Testing

- **Pin All reaches rows the browser never held.** On Model-AD Differential Expression (RNA, Hippocampus, a couple of models selected), search a term whose matches span several pages but stay under 50 -- e.g. `cacul1,ensa,plec`. Confirm the paginator reports more matches than one page holds, then click Pin All: every match should end up in the pinned table in the current sort order, the matching results table should read "No results found", and the `pinned` URL param should carry every id, not just the first page's.
- **Pin All stops at the limit and says so.** With the same categories, search a broad term (`a`). Click Pin All and confirm exactly 50 rows pin, a warning toast reads "Only 50 rows were pinned, because you reached the maximum of 50 pinned items.", and the Pin All button is disabled afterwards. Then unpin a few and click Pin All again -- it should top back up to 50 rather than re-pinning from scratch. Repeat with a different sort column selected first: which 50 rows survive should follow the sort showing in the table, since that is the only case where the sort changes the outcome.
- **Agora's CTs build their own query.** Each comparison tool supplies its own `pinAllFetch`, so a mis-wired query builder would only show up on that product. Verify both Agora CTs separately: on Nominated Targets search `in` and on Nominated Drugs search `ib`, confirming in each case that Pin All pins matches from beyond the current page. Then search `a` on Nominated Targets and confirm Pin All stops at 50 with the same warning toast.
- **URL pins over the limit are capped.** Hand-build a Differential Expression URL carrying 60 `pinned` ids. On load, confirm only the first 50 remain pinned, the same warning toast fires, and the URL is rewritten to the 50 that survived. Reload the trimmed URL and confirm no second toast.
- **The budget is honored and bounded server-side.** Confirm a budgeted request returns rows from beyond the first page, and that a budget above the pin limit is rejected with a 400 rather than silently clamped:

```bash
# 50 rows in one response, with the full match count intact
curl -s "http://localhost:8000/api/v1/comparison-tools/transcriptomics?categories=RNA%20-%20DIFFERENTIAL%20EXPRESSION&categories=Tissue%20-%20Hippocampus&itemFilterType=exclude&sortFields=gene_symbol&sortOrders=1&remainingBudget=50" \
  | jq '.page.totalElements, (.transcriptomics | length)'

# over the schema maximum -- expect 400
curl -s "http://localhost:8000/api/v1/comparison-tools/transcriptomics?categories=RNA%20-%20DIFFERENTIAL%20EXPRESSION&categories=Tissue%20-%20Hippocampus&itemFilterType=exclude&sortFields=gene_symbol&sortOrders=1&remainingBudget=51" | jq
```

### Preview

When search has matching results across pages, but total results are less than pin limit, every match is pinned: 

https://github.com/user-attachments/assets/7d612dff-e423-4387-9572-d0d266001136

When search has matching results across pages, and total results are more than pin limit, matches up to pin limit are pinned:

https://github.com/user-attachments/assets/1c518647-bf94-4dd0-8f8d-7380f2a91d36

When URL contains more pins than the limit, pinned items are truncated to pin limit: 

https://github.com/user-attachments/assets/54f6c122-47db-42f6-b13b-8c48d140f173

[MG-564]: https://sagebionetworks.jira.com/browse/MG-564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MG-565]: https://sagebionetworks.jira.com/browse/MG-565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
```diff
@@ -157,9 +186,14 @@ protected final Page<T> executePagedAggregation(Criteria matchCriteria, Pageable
...
-      operations.add(Aggregation.skip(skipCount));
-      operations.add(Aggregation.limit(pageable.getPageSize()));
+      boolean budgetApplies = !isInclude && remainingBudget != null;
+      if (budgetApplies) {
+        operations.add(Aggregation.limit(remainingBudget));
+      } else {
+        long skipCount = (long) pageable.getPageNumber() * pageable.getPageSize();
+        operations.add(Aggregation.skip(skipCount));
+        operations.add(Aggregation.limit(pageable.getPageSize()));
+      }
```
hallieswan, line 196: Do not apply budget for pinned requests (`itemFilterType = INCLUDE`) -- these are already asking for a known set of items. 

When we do apply the budget, only use `$limit` instead of both `$skip/$limit` so the request returns all matching rows in sort order from across all pages

libs/explorers/api-helper/src/main/java/org/sagebionetworks/explorers/ComparisonToolRepositorySupport.java
```diff
@@ -177,7 +211,8 @@ protected final Page<T> executePagedAggregation(Criteria matchCriteria, Pageable
         new Query(matchCriteria).collation(CASE_INSENSITIVE),
         getCollectionName()
       );
-      return new PageImpl<>(results.getMappedResults(), pageable, total);
+      Pageable resultPageable = budgetApplies ? Pageable.unpaged(pageable.getSort()) : pageable;
+      return new PageImpl<>(results.getMappedResults(), resultPageable, total);
```
hallieswan, line 215: Use `Pageable.unpaged` when the budget applies so the page number is 0 and size is number of rows returned, rather than the page values that weren't applied. It also stops `PageImpl` from rewriting the total, which the truncation warning on the frontend depends on

libs/explorers/constants/src/lib/toasts.ts
```diff
@@ -1 +1,6 @@
 export const TOAST_DURATION_MS = 5000; // milliseconds
+
+export const getMaxPinnedItemsWarning = (pinnedCount: number, maxPinnedItems: number): string => {
```
hallieswan, line 3: This toast copy is used both when pin-all has too many results and hits the pin limit and when the URL has too many pins that are capped to the pin limit. Move into this file so that both toasts (and their e2e tests) can share the copy

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -31,6 +32,17 @@ import { ToastNotificationService } from './toast-notification.service';
...
+ * Fetches the rows matching the current query across all pages, capped at `remainingBudget`, so
+ * "pin all" can pin matches the browser is not currently holding. `totalElements` is the full match
+ * count, which is what makes truncation detectable: `totalElements > rows.length` means the budget
+ * ran out before every match was returned.
+ */
+export type PinAllFetch<T> = (
+  query: ComparisonToolQuery,
+  remainingBudget: number,
+) => Observable<{ rows: T[]; totalElements: number }>;
+
```
hallieswan, line 45: Allow each CT consumer to provide its own pin all fetch at connect time

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -260,7 +275,7 @@ export class ComparisonToolService<T> {
 
   loadingResultsCount = computed(() => this.currentConfig()?.row_count ?? '');
   totalResultsCount = signal<number>(0);
-  pinnedResultsCount = signal<number>(0);
+  pinnedResultsCount = computed(() => this.pinnedData().length);
```
hallieswan, line 278: Derive `pinnedResultsCount` from `pinnedData`, rather than setting the value in each consumer CT by hand -- makes the pin limit enforceable in one place

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -555,29 +581,61 @@ export class ComparisonToolService<T> {
     this.setPinnedItems(this.visiblePinIds().filter((item) => item !== id));
   }
 
-  pinList(ids: string[]) {
```
hallieswan, line 558: `pinList` was only called by `pinAll`, so rename the method to make its use clear

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -555,29 +581,61 @@ export class ComparisonToolService<T> {
...
+    if (this.isLoadingTableData() || this.hasMaxPinnedItems()) return;
+
+    const currentPinIds = this.visiblePinIds();
+    const remainingBudget = this.maxPinnedItems() - currentPinIds.length;
+
+    this.startFetch();
+    fetch(this.query(), remainingBudget)
+      .pipe(
+        takeUntilDestroyed(this.destroyRef),
+        finalize(() => this.completeFetch()),
```
hallieswan, line 604: Use `finalize` rather than clearing the counter in `next` or `error` so an observable that compeltse without emitting can't leave the table stuck loading

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -620,8 +678,29 @@ export class ComparisonToolService<T> {
...
+   * `setPinnedItems` ignores a write that matches the current pins.
+   */
   setPinnedData(pinnedData: T[]) {
-    this.pinnedDataSignal.set(pinnedData);
+    const maxPinnedItems = this.maxPinnedItems();
+    if (pinnedData.length > maxPinnedItems) {
+      const trimmedData = pinnedData.slice(0, maxPinnedItems);
+      this.pinnedDataSignal.set(trimmedData);
+      this.showMaxPinnedItemsWarning(trimmedData.length);
+      this.setPinnedItems(this.extractRowIds(trimmedData));
```
hallieswan, line 700: Caps pinned data to the pin limit, so a hand-edited URL with more IDs than a user is allowed to pin shows a toast and drops the extra IDs

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.html
```diff
@@ -40,7 +40,7 @@
           <div class="table-divider-controls">
             <button
               class="pin-all-button"
-              [disabled]="hasMaxPinnedItems()"
+              [disabled]="hasMaxPinnedItems() || isLoadingTableData()"
```
hallieswan, line 43: Disable "Pin All" when loading table data, so the user can't click "Pin All" when we're mid-fetch

libs/agora/nominated-drugs-comparison-tool/src/lib/nominated-drugs-comparison-tool.component.ts
```diff
@@ -105,25 +105,36 @@ export class NominatedDrugsComparisonToolComponent implements OnInit, OnDestroy
...
+          ),
     });
   }
 
   ngOnDestroy() {
     this.comparisonToolService.disconnect();
   }
 
-  getUnpinnedData(currentQuery: ComparisonToolQuery) {
+  private buildUnpinnedQuery(
```
hallieswan, line 124: Extract unpinned query building to a helper, so it can be reused in the standard unpinned data fetch and in the new pin all fetch

libs/agora/nominated-drugs-comparison-tool/src/lib/nominated-drugs-comparison-tool.component.ts
```diff
@@ -174,11 +189,9 @@ export class NominatedDrugsComparisonToolComponent implements OnInit, OnDestroy
         next: (response: NominatedDrugsPage) => {
           const data = response.nominatedDrugs;
           this.comparisonToolService.setPinnedData(data);
-          this.comparisonToolService.pinnedResultsCount.set(data.length);
```
hallieswan, line 177: Pinned results count is now derived in the service

libs/explorers/testing/src/lib/e2e/comparison-tool-dropdown.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Selects an option in one of the category dropdowns. Returns once the dropdown has closed, which
+ * is before the table has refetched -- assert on the resulting data to wait for the new rows.
+ * @param page - Playwright Page object
+ * @param dropdownIndex - Position of the dropdown in the selector row, 0 being the leftmost and -1
+ * the rightmost. Levels with a single option render as plain text, so only rendered dropdowns count.
+ * @param option - The option label, matched case-insensitively as a substring, or the option's
+ * position in the listbox for tests that don't know the labels up front
+ * @returns The label of the selected option
+ */
+export const selectCategoryOption = async (
```
hallieswan, line 17: New helper method to replace the repeated logic for selecting a CT dropdown option across e2e tests

apps/model-ad/app/e2e/differential-expression-comparison-tool.spec.ts
```diff
@@ -233,6 +245,84 @@ test.describe('differential expression', () => {
...
+    await navigateToComparison(page, CT_PAGE, true, 'url', categoriesAndModelsQueryParameters);
+    await testPinAllAcrossPages(
+      page,
+      searchTerm,
+      transcriptomics.map((row) => row.composite_id),
+    );
+  });
+
+  test('Pin All pins every matching row for the protein main category', async ({ page }) => {
+    const proteinSearchTerm = 't';
```
rrchai, line 268: Nit: this search will return more rows as the real data grows, and the test will fail:  `testPinAllAcrossPages` below expects `expectedPinnedIds.length <= MAX_PINNED_ITEMS`. Would you consider a search term that will match count stays under the limit?
```suggestion
    const proteinSearchTerm = 'Dbt,Th,Tpd52l1';
```

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.html
```diff
@@ -40,7 +40,7 @@
...
               (click)="pinAll()"
               [pTooltip]="
                 hasMaxPinnedItems() ? disabledPinTooltip() : 'Pin all matching rows to the top.'
```
rrchai: Image

nit: As the button is now also disabled while loading, the tooltip here still shows 'Pin all matching rows to the top.' if you hover it again mid-fetch. It's a small window and I don't think it's critical. Just checking whether you'd noticed, and whether you'd want a different message for that state (something like 'Pinning rows...'), or to just suppress the tooltip while the fetch is in flight. Honestly, keeping it as is is no problem on my end either.

rrchai (approved):
Looking good, just two nitpicks

sagely1 (approved):
LGTM other than @rrchai nits

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.component.html
hallieswan: Good catch! I'll add a simple "Waiting for data to load..." tooltip when `isLoadingTableData()` is true, so we capture the cases where table data is loading but we're not actively pinning rows (e.g. user changes the selected filters, so the unpinned data is refetched)

apps/model-ad/app/e2e/differential-expression-comparison-tool.spec.ts
hallieswan: This is a great point! Unfortunately, we only have mock data for the proteomics collection right now, so I'm not sure what values will be valid once the real data is added. I'll add a comment to update this search term when we have the real proteomics data!

JessterB: There's a preview version of the real proteomics data here: https://www.synapse.org/Synapse:syn76225990

Unfortunately, it just missed the last data release....

rrchai (approved):

PR #4189, opened 2026-09-14
feat(explorers): heatmap circle is not drawn if either numeric result is null or missing (MG-1096)

hallieswan:
## Description

A heatmap cell carries two numeric values -- a color metric (`log2_fc`, `correlation`) and `adj_p_val`, which drives the circle's size. Model-AD's GX suites already validate that both are present and non-null, but if a malformed cell slipped through, the front end still drew something: a missing or null color value fell back to a pale gray circle sized by the surviving p-value, which reads as real data rather than as absent data. This change makes an unusable cell render nothing at all, so a data gap looks like a gap. Valid data is unaffected.

## Related Issue

- [MG-1096](https://sagebionetworks.jira.com/browse/MG-1096)

## Changelog

- Add heatmap circle utilities that treat a cell as drawable only when both the color metric and `adj_p_val` are usable numbers
- Skip rendering the heatmap circle and its surrounding click target for cells that cannot be drawn
- Simplify `HeatmapCircleComponent` around the drawable/undrawable distinction, removing the gray fallback circle
- Add a Storybook story and unit tests covering unusable cell data
- Correct the `HeatmapCircleData` description in `libs/explorers/comparison-tool/CLAUDE.md`

## Testing

- Open Storybook (`nx start explorers-storybook`, port 4402) and compare `Comparison Tool/ComparisonToolTable/HeatmapCircleComponent` -> `Demo` against `NoCircleDrawn`. `Demo` shows a circle; `NoCircleDrawn` is intentionally an empty canvas.
- In Model AD, open both comparison tools (Disease Correlation and Differential Expression) and confirm the heatmaps look unchanged for real data -- same circle sizes, colors, and tooltips as on `main`.
- Click a heatmap circle in each tool and confirm the details panel still opens with the expected values.
- Toggle the significance threshold control and confirm circles above the threshold still disappear and reappear as before.
- Hover an empty heatmap cell and confirm no tooltip or pointer cursor appears, since no circle or button is rendered there.
- Check that the legend's no-circle entry still explains the empty-cell case to users.

### Preview

Storybook demonstrates all data cases described in the ticket: 

size value | color value | storybook
:---: | :---: | :---: 
numeric | null | MG-1096_colorValue_null
null | numeric | MG-1096_sizeValue_null
numeric | missing | MG-1096_colorValue_missing
missing | numeric | MG-1096_sizeValue_missing
missing | missing | MG-1096_bothValues_missing

Model AD differential expression CT is unchanged, but inspecting a "missing" data cell shows that an unusable button is no longer rendered: 

dev: 
MG-1096_dev

this PR: 
MG-1096_thisPR

[MG-1096]: https://sagebionetworks.jira.com/browse/MG-1096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/explorers/comparison-tool/src/lib/comparison-tool-table/base-table/base-table.component.html
```diff
@@ -49,7 +49,7 @@
                 [rowData]="rowData"
               />
             }
-            @if (columnType === 'heat_map') {
+            @if (columnType === 'heat_map' && canDrawHeatmapCircle(cellData)) {
```
hallieswan, line 52: Don't render an unusable button if the heatmap circle can't actually be drawn

libs/explorers/comparison-tool/src/lib/comparison-tool-table/heatmap-circle/heatmap-circle.component.ts
```diff
@@ -175,22 +158,4 @@ export class HeatmapCircleComponent<T extends HeatmapCircleData = HeatmapCircleD
...
-    const size = this.getCircleSize(adjustedPValue);
-    const color = this.getCircleColor(colorValue);
-
-    return {
-      display: size > 0 ? 'block' : 'none',
-      width: size + 'px',
-      height: size + 'px',
-      backgroundColor: color,
-    };
-  }
```
hallieswan, line 189: Collapsed into `circleStyle` computed signal, since we now handle the "undrawable" case upfront

libs/explorers/comparison-tool/src/lib/comparison-tool-table/heatmap-circle/heatmap-circle.component.ts
```diff
@@ -114,15 +104,12 @@ export class HeatmapCircleComponent<T extends HeatmapCircleData = HeatmapCircleD
     }
   }
 
-  private getCircleValueSign(colorValue: number | null | undefined): CircleValueSign {
-    if (colorValue === null || colorValue === undefined) return 'none';
```
hallieswan, line 118: Simplify these methods, since we now handle the "undrawable" case upfront

libs/explorers/comparison-tool/src/lib/comparison-tool-table/heatmap-circle/heatmap-circle.utils.ts
```diff
@@ -0,0 +1,50 @@
...
+export const ADJUSTED_P_VALUE_KEY = 'adj_p_val';
+
+export type HeatmapCircleMetrics<T extends HeatmapCircleData = HeatmapCircleData> =
+  | {
+      isDrawable: true;
+      colorKey: HeatmapCircleColorKey<T>;
+      colorValue: number;
+      adjustedPValue: number;
+    }
+  | { isDrawable: false };
```
hallieswan, line 12: A heatmap circle must have a numeric value for both its color and its size (adjusted p-value), otherwise we will not draw it -- use the type to distinguish between these cases clearly

libs/explorers/comparison-tool/src/lib/comparison-tool-table/heatmap-circle/heatmap-circle.utils.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  it('does not let an unrelated key stand in for the color metric', () => {
+    const metrics = resolveHeatmapCircleMetrics({
+      correlation: null,
+      sample_count: 12,
+      [ADJUSTED_P_VALUE_KEY]: 0.01,
+    });
+
+    expect(metrics.isDrawable).toBe(false);
+  });
```
rrchai: looks like this passes because `correlation` is null and comes first and `sample_count` never gets read, so it isn't really being rejected. Would you like to rename this to what it checks a null color value, or refactor to test an unrelated key?

libs/explorers/comparison-tool/src/lib/comparison-tool-table/heatmap-circle/heatmap-circle.utils.ts
```diff
@@ -0,0 +1,50 @@
+import { HeatmapCircleColorKey, HeatmapCircleData } from '@sagebionetworks/explorers/models';
+
+export const ADJUSTED_P_VALUE_KEY = 'adj_p_val';
```
rrchai: Nit: looks like `comparison-tool.variables.ts` declares the constant for `ADJUSTED_P_VALUE_LABEL`. Do you consider to move this key next to it?

rrchai (approved):
LGTM, just two non-blocking suggestions.

libs/explorers/comparison-tool/src/lib/comparison-tool-table/heatmap-circle/heatmap-circle.component.ts
```diff
@@ -175,22 +158,4 @@ export class HeatmapCircleComponent<T extends HeatmapCircleData = HeatmapCircleD
     // ensure the smallest circles have a min size to be easily hoverable/clickable
     return size < MIN_SIZE ? MIN_SIZE : size;
```
sagely1: This function has a clamp on MIN_SIZE but doesn't do so for MAX_SIZE.

libs/explorers/comparison-tool/src/lib/comparison-tool-table/base-table/base-table.component.html
sagely1: `resolveHeatmapCircleMetrics` is recomputed in `HeatmapCircleComponent` - consider computing once and reusing the output

libs/explorers/comparison-tool/src/lib/comparison-tool-table/heatmap-circle/heatmap-circle.component.ts
```diff
@@ -28,73 +38,53 @@ export class HeatmapCircleComponent<T extends HeatmapCircleData = HeatmapCircleD
...
+      display: size > 0 ? 'block' : 'none',
+      width: size + 'px',
+      height: size + 'px',
+      backgroundColor: color,
+    };
   });
 
   private getDefaultTooltip(data: T | null | undefined): string {
-    if (!data) {
+    const metrics = resolveHeatmapCircleMetrics(data);
```
sagely1: could we use `this.metrics()` here?

libs/explorers/comparison-tool/src/lib/comparison-tool-table/heatmap-circle/heatmap-circle.utils.spec.ts
hallieswan: Good call! Will fix in the next commit

libs/explorers/comparison-tool/src/lib/comparison-tool-table/heatmap-circle/heatmap-circle.utils.ts
hallieswan: Good catch! Will do in the next commit

libs/explorers/comparison-tool/src/lib/comparison-tool-table/heatmap-circle/heatmap-circle.component.ts
hallieswan: Good call, will fix in the next commit

rrchai (approved):
LGTM

PR #4190, opened 2026-09-14
fix(model-ad): rapid CT filter toggling bug fix (MG-1047)

sagely1:
## Description

Removing multiple filters from a comparison tool in quick succession could leave the results table out of sync with the filter UI: a chiclet would disappear but its query never re-fired, so the table kept showing rows for a filter the user had already removed. The root cause was a race — the removal handler mutated a filter option in place and then re-read/cloned the filter signal, so a second rapid click operated on a stale object that was no longer part of the current state. This change makes filter mutations atomic and identity-independent so every removal is reflected in the fired query, regardless of how fast the user clicks.

## Related Issue

- [MG-1047](https://sagebionetworks.jira.com/browse/MG-1047)

## Changelog

- Add `setFilterOptionSelected` and `clearAllFilters` to `ComparisonToolFilterService` that rebuild the filter array immutably from the current signal value, keyed by `data_key` + `label` instead of object identity
- Rewire filter chiclet removal and "Clear all" in `ComparisonToolFilterListComponent` to the new keyed service methods, dropping in-place option mutation
- Switch filter panel checkboxes to one-way `[ngModel]` + `ngModelChange` so toggles route through the keyed service method instead of mutating the option object
- Track the filter list `@for` loops by stable keys (`data_key` / `label`) rather than `$index`
- Add a regression test covering two filter removals in immediate succession

## Testing

Note: This is an intermittent timing bug so this requires a few passes to confirm and can be seen on DEV with a few tries.

- Open the Differential Expression comparison tool in Model-AD
- Add three or more model/filter chiclets (choose a combination unlikely to be cached)
- Quickly click the "x" on two of the chiclets in rapid succession, then confirm both chiclets disappear **and** the table updates to reflect only the remaining filter(s)
- Repeat the rapid toggle using the filter panel checkboxes (select several, then rapidly deselect two) and confirm the table stays consistent with the checkbox state
- Use "Clear all" and confirm every chiclet is removed and the table returns to the unfiltered result set

## Preview

[MG-1047]: https://sagebionetworks.jira.com/browse/MG-1047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/explorers/services/src/lib/comparison-tool-filter.service.ts
```diff
@@ -28,6 +28,34 @@ export class ComparisonToolFilterService {
     });
   }
 
+  setFilterOptionSelected(filterDataKey: string, optionLabel: string, selected: boolean) {
+    const updatedFilters = this.filters().map((filter) =>
+      filter.data_key === filterDataKey
```
hallieswan: What do you think about keying off the `query_param_key` instead? A filter's `data_key` wasn't read anywhere in explorers before this change and existing CT services already treat `query_param_key` as the filter's identity -- `ComparisonToolHelperService` uses `query_param_key` as the map key, it's what the URL sync round-trips through, and it's the key each consumer CT looks up selections by when building its search query. It also has to be unique for the URL sync to work, but nothing enforces that for `data_key`.

libs/explorers/comparison-tool/src/lib/comparison-tool-filter-list/comparison-tool-filter-list.component.html
```diff
@@ -20,13 +20,13 @@
             (cleared)="removeSignificanceThresholdFilter()"
           />
         }
-        @for (filter of filters(); track $index) {
-          @for (option of filter.options; track $index) {
+        @for (filter of filters(); track filter.data_key) {
```
hallieswan: If we switch to using `query_param_key` elsewhere, then would also be great to align the track identity on that as well!

libs/explorers/comparison-tool/src/lib/comparison-tool-filter-panel/comparison-tool-filter-panel.component.html
hallieswan: If we update the `filter-list` track to use `filter.query_param_key`, then consider making the same change here, so the components are aligned on the filter identify

libs/explorers/services/src/lib/comparison-tool-filter.service.ts
```diff
@@ -28,6 +28,34 @@ export class ComparisonToolFilterService {
     });
   }
```
hallieswan: Does `setFilters` still need to live in the service? The only caller is the CT filter provider, which is used by tests. Could consider refactoring, so we don't inadvertently use this method again in the future (and re-introduce this bug)

libs/explorers/services/src/lib/comparison-tool-filter.service.ts
```diff
@@ -28,6 +28,34 @@ export class ComparisonToolFilterService {
...
+          }
+        : filter,
+    );
+    this.comparisonToolService.updateQuery({
+      filters: updatedFilters,
+      pageNumber: this.comparisonToolService.FIRST_PAGE_NUMBER,
+    });
+  }
+
+  clearAllFilters() {
```
hallieswan, line 45: What do you think about adding early returns to both `setFilterOptionSelected` and `clearAllFilters` when no filter option changes? Then we could skip duplicate data fetches. For example, if we toggle the significance filter on, then click "Clear All", we re-fetch the unpinned data, even though the significance filter is a frontend-only filter, so the filters object and the unpinned data are unchanged

hallieswan (commented):
This is a great fix! Just a couple small comments

libs/explorers/services/src/lib/comparison-tool-filter.service.ts
sagely1: Looks like `data_key` is dead code that is confusing.  I'll also remove it from ComparisonToolConfigFilter.

hallieswan (approved):
Looks good to me!

PR #4192, opened 2026-09-15
fix(model-ad): fix out of order api responses bug (MG-1108)

sagely1:
## Description

The Agora and Model-AD comparison tools re-fetch data whenever the query changes, but each change fired an independent HTTP request with no cancellation and wrote its result to the table unconditionally. Responses were therefore applied in resolution order, not issue order, so a slower earlier request (e.g. a broad, uncached filter combination) could resolve after a faster later one and overwrite it — leaving the table showing rows that don't match the current filter UI. This centralizes each data stream through a single `switchMap`-driven subscription in the shared `ComparisonToolService`, so a new query cancels the prior in-flight request and only the latest result renders. This is distinct from MG-1047, which fixed which query gets *fired*; this fixes which fired query's result gets *rendered*.

## Related Issue

- [MG-1108](https://sagebionetworks.jira.com/browse/MG-1108) — CT table state could be out of sync due to requests returning out of order
  - Related: [MG-1047](https://sagebionetworks.jira.com/browse/MG-1047) — fixed the filter-state removal race (which query is fired); this ticket fixes the response race (which result is rendered)

## Changelog

- Add a single `switchMap`-driven fetch stream per data path (pinned/unpinned) in `ComparisonToolService`, so a newer query cancels the prior in-flight request and only the latest response writes the table and results count (latest-wins).
- Add `fetchUnpinned`/`fetchPinned` methods that take a mapped result observable; keep the loading counter balanced on completion, error, or cancellation via `finalize`, and map failures to an empty result via `catchError`.
- Update the four Model-AD comparison tools (differential expression, disease correlation, mouse model overview, marmoset model overview) to hand the service a mapped `{ data, totalCount }` observable instead of subscribing themselves.
- Remove the per-component manual `.subscribe()` blocks and their `takeUntilDestroyed` fetch teardown; error presentation stays centralized in the existing HTTP error interceptor.
- Extend and update unit tests: base-service latest-wins/cancellation/loading-counter/error coverage, and the differential-expression component fetch tests.

## Testing

- Serve Model-AD and open the Differential Expression comparison tool.
- Throttle the network in DevTools (slow/variable latency) and rapidly select/deselect several model filters; confirm the table always settles to match the current filter chiclets, never a superseded query.
- In the network panel, confirm superseded in-flight requests are cancelled when a newer query fires.
image
- Verify both the unpinned (filtered) rows and pinned rows behave correctly, and that the displayed results count matches the final state.
- Repeat with an uncached filter combination and on the Disease Correlation, Mouse Model Overview, and Marmoset Model Overview tools.

[MG-1108]: https://sagebionetworks.jira.com/browse/MG-1108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MG-1047]: https://sagebionetworks.jira.com/browse/MG-1047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -582,9 +613,9 @@ export class ComparisonToolService<T> {
    * Pins every row matching the current query, not just the rows on the current page. The matching
    * ids are discovered server-side, capped at the pins the user has left.
    *
-   * Only starts when no fetch is in flight: mid-fetch, the pinned data on hand is not yet the
-   * server-confirmed pin set, so the remaining budget could not be computed against it. The
-   * `startFetch()` below therefore also serves as the re-entry guard.
+   * The guard below keeps this from running while the table is still being populated: the cap sent to
+   * the server is how many more pins the user may add, counted from the pins on screen, and those are
+   * not final until the in-flight fetch lands.
```
hallieswan: Nit: should we keep the note about `startFetch()` acting as the re-entry guard, since that's the only thing documenting that there can only be one pin-all fetch at a time?

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -248,41 +310,11 @@ export class ComparisonToolService<T> {
...
-    const configColumns = config.columns;
-    if (!configColumns || configColumns.length === 0) return [];
-
-    const savedColumns = this.columnsForDropdownsSignal().get(this.dropdownKey(config.dropdowns));
-    return this.applyColumnPreferences(configColumns, savedColumns);
-  });
-
-  selectedColumns = computed(() => {
-    return this.columns().filter((col) => col.selected);
-  });
```
hallieswan, line 268: These were moved to the "Config-Driven Signals" section above without any changes

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -248,41 +310,11 @@ export class ComparisonToolService<T> {
...
-  totalResultsCount = signal<number>(0);
-  pinnedResultsCount = computed(() => this.pinnedData().length);
-
-  hasMaxPinnedItems = computed(() => {
-    return this.pinnedResultsCount() >= this.maxPinnedItems();
-  });
-
-  disabledPinTooltip = computed(() => {
-    return `You have already pinned the maximum number of items (${this.maxPinnedItems()}). You must unpin some items before you can pin more.`;
-  });
```
hallieswan, line 284: These were moved to the "Results & Pin State Signals" section above without any changes

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -160,12 +181,52 @@ export class ComparisonToolService<T> {
   // Used for the URL serialization
   readonly visiblePinIds = computed(() => this.extractRowIds(this.pinnedData()));
 
-  private extractRowIds(rows: T[]): string[] {
-    const rowIdKey = this.viewConfig().rowIdDataKey;
-    return rows.map((row: T) => String((row as Record<string, unknown>)[rowIdKey]));
-  }
```
hallieswan, line 166: `extractRowIds` was moved to the bottom of the file without any changes

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -117,6 +127,17 @@ export class ComparisonToolService<T> {
...
   private readonly hoveredRowIdSignal = signal<string | null>(null);
 
+  // Fetch streams: each fetch pushes a result observable onto these subjects, and switchMap keeps
+  // only the latest in flight. A newer query cancels (unsubscribes) the prior request, so responses
+  // can never be applied out of order. Unpinned and pinned are independent streams.
+  //
+  // Typed with `unknown` rather than `T`: the coordinator keeps every CT service in one
+  // ComparisonToolService<unknown> list, and a `Subject<...T...>` field would make that assignment
+  // a compile error. The public fetch methods stay strongly typed, and the stream subscription
+  // casts the result rows back to `T[]`.
```
hallieswan: Optional nit on the wording here: 

```suggestion
  // Typed with `unknown` rather than `T`: the coordinator tracks the active CT service as 
  // ComparisonToolService<unknown>, and a `Subject<...T...>` field would make that assignment
  // a compile error. The public fetch methods stay strongly typed, and the stream subscription
  // casts the result object back to `T[]`.
```

libs/explorers/services/src/lib/comparison-tool.service.ts
hallieswan, line 670: Nit: since we moved the comment about this guard preventing an indefinite refetch to `applyPinnedData`, what do you think about adding a short comment here explaining why the `isEqual` is necessary? e.g. something like this: 

```suggestion
  // applyPinnedData can re-send the same ids in a new array, so compare by value to terminate the trim cycle
  if (isEqual(deduplicatedItems, this.pinnedItems())) {
    return;
  }
```

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -698,9 +742,19 @@ export class ComparisonToolService<T> {
...
+
+  /**
+   * Imperatively sets pinned data outside a fetch stream (test/story setup). Runtime fetches go
+   * through the pinned stream, which applies the same cap via `applyPinnedData`. The
+   * `completeFetch()` balances a `startFetch()` a caller may have paired with this write.
+   */
+  setPinnedData(pinnedData: T[]) {
+    this.applyPinnedData(pinnedData);
     this.completeFetch();
   }
```
hallieswan: Nit: should this comment lead with the test/story restriction, so it's clear that we shouldn't use this method in a runtime caller anymore? Alternately, should we have `provideComparisonToolService` and stories set data with `fetchPinned(of(...))`, so `setPinnedData` can go away entirely? We should align `setUnpinnedData` on the same decision (e.g. either add a clear comment that it's only used for tests/stories or update stories/provider to set data directly)

libs/model-ad/mouse-model-overview-comparison-tool/src/lib/mouse-model-overview-comparison-tool.component.ts
```diff
@@ -146,25 +144,18 @@ export class MouseModelOverviewComparisonToolComponent implements OnInit, OnDest
...
-          this.comparisonToolService.totalResultsCount.set(0);
-        },
-      });
+    this.comparisonToolService.fetchUnpinned(
+      this.mouseModelOverviewService.getMouseModelOverviews(query).pipe(
+        map((response: MouseModelOverviewsPage) => ({
+          data: response.mouseModelOverviews,
+          totalCount: response.page.totalElements,
+        })),
+      ),
```
hallieswan, line 157: Could you update the [eQTL comparison tool component comments](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/qtl/eqtl-comparison-tool/src/lib/eqtl-comparison-tool.component.ts) to match this new setup, so we don't try to use the old methods (e.g. `setUnpinnedData`) when we're ready to configure it?

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -743,6 +797,50 @@ export class ComparisonToolService<T> {
...
+   * Fetches pinned rows from the given result observable. Independent of the unpinned stream, so a
+   * new pinned fetch does not cancel an in-flight unpinned fetch (and vice versa).
+   */
+  fetchPinned(source$: Observable<ComparisonToolFetchResult<T>>): void {
+    this.startFetch();
+    this.pinnedFetch$.next(source$);
+  }
+
   /** Call before starting a data fetch to increment loading counter */
   startFetch() {
```
hallieswan: What do you think about making this `private` like `completeFetch`, so only the CT service itself manages the fetch counter? We'd need to update tests etc... that relied on calling `startFetch` to reach the loading state to use something like `fetchUnpinned(new Subject<never>())` instead

hallieswan (commented):
Looking great to me! Mostly small nits to clarify the new comments

libs/explorers/services/src/lib/comparison-tool.service.ts
sagely1: Good idea.  I will opt for updating the stories with `fetchPinned(of(...))`

libs/qtl/eqtl-comparison-tool/src/lib/eqtl-comparison-tool.component.ts
```diff
@@ -62,38 +62,23 @@ export class EqtlComparisonToolComponent {
...
-  //       },
-  //     });
+  //   this.comparisonToolService.fetchUnpinned(
+  //     this.eqtlService.getEqtlData({ ...query, selectedRowId }).pipe(
+  //       map((response) => ({ data: response.data, totalCount: response.page.totalElements })),
+  //       // NOTE: selectedRowId and notifySelectedRowValidity are specific to the selected row
+  //       // feature (rowSelectionEnabled) -- other CTs do not use them. The backend must check
+  //       // whether selectedRowId exists anywhere in the full filtered result set (not just the
+  //       // current page) and return selectedRowInResults: boolean in the response. Call
+  //       // notifySelectedRowValidity in the subscribe below if needed.
```
hallieswan, line 72: I probably missed it, but what is the "subscribe below" that the comment mentions?

hallieswan (approved):
Looks good to me! Just one small non-blocking question

libs/explorers/services/src/lib/comparison-tool.service.ts
```diff
@@ -160,12 +181,52 @@ export class ComparisonToolService<T> {
...
 
   constructor() {
+    // Unpinned rows are paged, so totalCount is the cross-page total used for pagination.
+    this.subscribeToFetchStream(this.unpinnedFetch$, ({ data, totalCount }) => {
+      this.unpinnedDataSignal.set(data);
+      this.totalResultsCount.set(totalCount);
+    });
+
+    // Pinned rows are never paged: applyPinnedData enforces the pin cap and the count comes from
+    // pinnedData(), so totalCount from the response is unused and therefore is ommitted intentionally.
```
rrchai:
```suggestion
    // pinnedData(), so totalCount from the response is unused and therefore is omitted intentionally.
```

rrchai (approved):
LGTM - just one typo

libs/qtl/eqtl-comparison-tool/src/lib/eqtl-comparison-tool.component.ts
sagely1: Not worded well and this section is extraneous.  Removing in next commit.

libs/explorers/services/src/lib/comparison-tool.service.ts
sagely1: Fixed in next commit

rrchai (approved):

libs/qtl/eqtl-comparison-tool/src/lib/eqtl-comparison-tool.component.ts
```diff
@@ -62,38 +62,18 @@ export class EqtlComparisonToolComponent {
...
-  //     .pipe(takeUntilDestroyed(this.destroyRef))
-  //     .subscribe({
-  //       next: (response) => {
-  //         this.comparisonToolService.setUnpinnedData(response.data);
-  //         this.comparisonToolService.totalResultsCount.set(response.page.totalElements);
-  //         // NOTE: selectedRowId and notifySelectedRowValidity are specific to the
-  //         // selected row feature (rowSelectionEnabled) -- other CTs do not use them.
-  //         // The backend must check whether selectedRowId exists anywhere in the full
-  //         // filtered result set (not just the current page) and return
-  //         // selectedRowInResults: boolean in the response.
```
hallieswan: I think we may want to keep this comment -- it's describing a base CT feature (row selection) that isn't used by any other CTs yet, so would have to be rediscovered without this comment.

hallieswan (approved):

libs/qtl/eqtl-comparison-tool/src/lib/eqtl-comparison-tool.component.ts
sagely1: Ah you're right.  I'll rewrite the comment to be clearer since it seemed a bit unclear

hallieswan (approved):

PR #4193, opened 2026-09-17
feat(model-ad): add individual proteomics route and wire it into protein details (MG-1022)

rrchai:
## Description

This PR adds the `/proteomics-individual` route to `model-ad-api-next` and wires it into the Protein Details page, replacing the hardcoded mock data. Users can now view real per-animal protein abundance for a given protein, tissue and model, displayed as boxplots across each age timepoint. 

## Related Issue

- [MG-1022](https://sagebionetworks.jira.com/browse/MG-1022)

## Changelog

- Add `ProteomicsIndividual` and `ProteomicsIndividualFilterQuery` schemas plus the `/proteomics-individual` path to the api-next OpenAPI spec, and regenerate the Java and Angular clients
- Add the Java endpoint — document, repository, mapper, service and API delegate implementation — querying `protein_de_individual` by `unique_id` + tissue + (`name` | `model_group`) and returning results sorted by `age_numeric` ascending
- Register a `proteomicsIndividual` cache in `CacheNames` and `CacheConfiguration`, keyed on all four query parameters
- Route `/api/v1/proteomics-individual` to `model-ad-api-next` in the apex Caddyfile
- Wire `ProteinDetailsComponent` to fetch from the new endpoint, removing all four `TODO(MG-1022)` markers and the hardcoded mock import
- Retype `proteomicsIndividualMocks` from a hand-written interface to the generated `ProteomicsIndividual` type
- Add unit tests for the service, delegate and mapper, including `age_numeric` ordering, null-age handling and rejection of unrecognized query parameters

## Testing

- The `name` lookup returns four records in ascending age order:

  ```console
  curl -s "http://localhost:8000/api/v1/proteomics-individual?tissue=Hemibrain&uniqueId=ENSMUSG00000029781Q9Z247&modelIdentifier=LOAD2&modelIdentifierType=name" | jq -c '[.[].age_numeric]'
  ```

  it returns `[4,12,18,24]`

- The `modelGroup` lookup returns the same set:

  ```console
  curl -s "http://localhost:8000/api/v1/proteomics-individual?tissue=Hemibrain&uniqueId=ENSMUSG00000029781Q9Z247&modelIdentifier=LOAD2&modelIdentifierType=modelGroup" | jq -c '[.[].age_numeric]'
  ```

  it returns `[4,12,18,24]`

- The response carries the three proteomics-only fields:

  ```console
  curl -s "http://localhost:8000/api/v1/proteomics-individual?tissue=Hemibrain&uniqueId=ENSMUSG00000029781Q9Z247&modelIdentifier=LOAD2&modelIdentifierType=name" | jq -c '.[0] | {unique_id, uniprotid, display_symbol, units, result_order}'
  ```

  it returns `{"unique_id":"ENSMUSG00000029781Q9Z247","uniprotid":"Q9Z247","display_symbol":"Fkbp9 (Q9Z247)","units":"Log2 Counts per Million","result_order":["LOAD1","LOAD2"]}`

- Unknown query parameters are rejected:

  ```console
  curl -s "http://localhost:8000/api/v1/proteomics-individual?tissue=Hemibrain&uniqueId=ENSMUSG00000029781Q9Z247&modelIdentifier=LOAD2&modelIdentifierType=name&bogus=x" | jq -c '{status, detail}'
  ```

  it returns `{"status":400,"detail":"Unknown query parameter: bogus"}`

- An invalid `modelIdentifierType` is rejected, as is omitting any required parameter:

  ```console
  curl -s "http://localhost:8000/api/v1/proteomics-individual?tissue=Hemibrain&uniqueId=ENSMUSG00000029781Q9Z247&modelIdentifier=LOAD2&modelIdentifierType=nonsense" | jq -c '{status, detail}'
  ```

  it returns `{"status":400,"detail":"Invalid request content."}`

- A `uniqueId` with no matching documents returns an empty list rather than an error:

  ```console
  curl -s -w " <- HTTP %{http_code}\n" "http://localhost:8000/api/v1/proteomics-individual?tissue=Hemibrain&uniqueId=NOSUCHID&modelIdentifier=LOAD2&modelIdentifierType=name"
  ```

  it returns `[] <- HTTP 200`
 

## Preview

The Protein Details page renders as expected with live data. Same URL on both sides:
`http://localhost:8000/proteins/ENSMUSG00000029781Q9Z247?modelGroup=LOAD2&tissue=Hemibrain`

Screenshot 2026-09-17 at 2 50 40 PM

[MG-1022]: https://sagebionetworks.jira.com/browse/MG-1022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/model-ad/api-next/src/main/resources/openapi.yaml
```diff
@@ -2359,6 +2353,132 @@ components:
...
+          description: Model group
+          nullable: true
+          type: string
+        matched_control:
+          description: Matched control for the model
+          example: C57BL/6J
+          type: string
+        units:
+          description: Units of measurement for proteomics data
+          example: Log2 Counts per Million
```
rrchai, line 2421: @JessterB  does `Log2 Counts per Million` look correct as the `units` example for proteomics?

JessterB: @rrchai I'm still trying to get our collaborators to tell me the correct units for that plot. I think it might actually end up being Log2 Parts Per Million, but that isn't confirmed....

apps/model-ad/api-next/src/test/java/org/sagebionetworks/model/ad/api/next/service/ProteomicsIndividualServiceTest.java
```diff
@@ -0,0 +1,309 @@
...
+
+    // then
+    assertThat(result).hasSize(3);
+    assertThat(result.get(0).getAgeNumeric()).isEqualTo(4);
+    assertThat(result.get(1).getAgeNumeric()).isEqualTo(12);
+    assertThat(result.get(2).getAgeNumeric()).isEqualTo(18);
+  }
+
+  @Test
+  @DisplayName("should sort documents with null age_numeric last")
```
sagely1, line 207: `age_numeric` is defined as required in the OpenAPI spec.  What do you think about aligning to match?

libs/model-ad/api-description/src/components/schemas/ProteomicsIndividual.yaml
```diff
@@ -0,0 +1,75 @@
...
+    description: List of genotypes in the order to display results
+    example: ['LOAD1', 'LOAD2']
+  data:
+    type: array
+    description: List of individual proteomics data points
+    items:
+      $ref: ./IndividualData.yaml
+required:
+  - ensembl_gene_id
+  - gene_symbol
```
sagely1, line 64: `gene_symbol` is `@Nullable` and required.  Consider updating to match.

sagely1 (changes requested):
Looks good! Just a few suggestions.

hallieswan (approved):
Looks good to me!

libs/model-ad/api-description/src/components/schemas/ProteomicsIndividual.yaml
rrchai: Good catch. It looks like the `@Nullable` is the wrong side, neither RNA/protein indivi collection has a null `gene_symbol`. I created a [ticket](https://sagebionetworks.jira.com/browse/MG-1129) to update both together in a separate pr.

apps/model-ad/api-next/src/test/java/org/sagebionetworks/model/ad/api/next/service/ProteomicsIndividualServiceTest.java
rrchai: Good call. To clarify, a field can be `required` + `nullable`, but `age_numeric` does not have any nulls in data as well, so null genuinely can't occur here. I'll drop that test to match RNA individual's in the next commit.

libs/model-ad/api-description/src/components/schemas/ProteomicsIndividual.yaml
rrchai: I included the changes to drop `@Nullable` from `geneSymbol` in both individual expression documents to match the schema in this PR.

apps/model-ad/api-next/src/main/java/org/sagebionetworks/model/ad/api/next/service/ProteomicsIndividualService.java
```diff
@@ -0,0 +1,63 @@
...
+    } else {
+      throw new IllegalArgumentException("Invalid modelIdentifierType: " + modelIdentifierType);
+    }
+
+    return documents
+      .stream()
+      .sorted(
+        Comparator.comparing(
+          ProteomicsIndividualDocument::getAgeNumeric,
+          Comparator.nullsLast(Comparator.naturalOrder())
```
sagely1: We can probably remove the nullsLast() since it is non-nullable now.  I think we can also remove it in TranscriptomicsIndividualService.

rrchai: sounds good to me, will do in next commit.

sagely1 (approved):
LGTM!

PR #4195, opened 2026-09-17
feat(model-ad): added customizable note section, updated labels, tooltips and text to download pin control

sagely1:
## Description

UX interviews found that the comparison tool's pinned-results download control was
unclear. This PR clarifies that control by relabeling its actions, adding explanatory tooltips, and surfacing a documentation link that points users to the study-specific data, metadata, and methods files.

## Related Issue

- [MG-1083](https://sagebionetworks.jira.com/browse/MG-1083) — CT: Adjust labels and text for download control

## Changelog

- Rename the pinned-results actions to "Download Pins" and "Clear All Pins"
- Add tooltips to the Download Pins and Clear All Pins controls
- Change the download dialog heading to "Download pinned results as:"
- Add an optional linked note under the download button, driven by a new `note` input on the shared download-dom-image components and a `DownloadNote` model
- Wire the Model AD Explorer documentation note into all four Model-AD comparison tools via a `DOWNLOAD_PINS_NOTE` config constant and the `downloadPinsNote` view-config field

## Testing

- Open any Model-AD comparison tool (mouse/marmoset model overview, differential expression, disease correlation) and pin at least one row so the pinned-results controls appear
- Confirm the button reads "Download Pins" and the secondary action reads "Clear All Pins"
- Hover each control and confirm the tooltips "Download pinned results" and "Clear all pinned results"
- Open the Download Pins popover and confirm the heading reads "Download pinned results as:"
- Confirm the note appears under the download button and its "Model AD Explorer documentation" link opens the AD Knowledge Portal resources page in a new tab
- Confirm the note is absent for download controls that aren't given one (e.g. plot downloads on details pages)

### Preview

| Before | After |
| ------ | ----- |
| image | image |

[MG-1083]: https://sagebionetworks.jira.com/browse/MG-1083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/model-ad/config/src/lib/constants.ts
```diff
@@ -1,11 +1,20 @@
...
 export const HELP_URL =
   'https://help.adknowledgeportal.org/apd/Model-AD+Explorer+Resources.4077682781.html';
 
 export const SUPPORT_EMAIL = 'modeladexplorer@sagebionetworks.org';
 
+export const DOWNLOAD_PINS_NOTE: DownloadNote = {
+  textBefore: 'See the ',
+  linkText: 'Model AD Explorer documentation',
+  linkUrl:
+    'https://help.adknowledgeportal.org/apd/Model-AD-Explorer-Resources-Resources.4077682781.html#ModelADExplorerResources-GeneExpressionAnalysis',
```
hallieswan: This link is pretty similar to the `HELP_URL` above -- could we construct it using that constant instead? I checked and both versions render the same help page.

```suggestion
  linkUrl:
    `${HELP_URL}#ModelADExplorerResources-GeneExpressionAnalysis`,
```

hallieswan: The section referenced in the link is also specific to the gene expression analysis -- should each CT link to a different section of the help docs? e.g. Disease Correlation CT links to the "Disease Correlation Analysis" section. Alternately, we could drop the fragment and allow the user to identify the relevant section themselves?

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.constants.ts
```diff
@@ -20,6 +20,13 @@ export const COMPARISON_TOOL_BODY_CLASS = 'comparison-tool-body';
 export const PIN_ALL_TOOLTIP = 'Pin all matching rows to the top.';
 export const PIN_ALL_LOADING_TOOLTIP = 'Waiting for data to load...';
 
+// Pinned results download control
+export const DOWNLOAD_PINS_LABEL = 'Download Pins';
+export const DOWNLOAD_PINS_TOOLTIP = 'Download pinned results';
+export const DOWNLOAD_PINS_HEADING = 'Download pinned results as:';
+export const CLEAR_ALL_PINS_LABEL = 'Clear All Pins';
```
hallieswan: Should we update e2e tests to use this label to identify the clear all pins button? e.g. [`testFiltersRemovedFromUrlOnClearAll`](https://github.com/Sage-Bionetworks/sage-monorepo/blob/main/libs/explorers/testing/src/lib/e2e/comparison-tool-filter.ts#L43-L53) and test description across [CT e2e files](https://github.com/search?q=repo%3ASage-Bionetworks%2Fsage-monorepo+filters+are+removed+from+url+when+clear+all+is+clicked&type=code)

libs/explorers/comparison-tool/src/lib/comparison-tool-table/comparison-tool-table.constants.ts
```diff
@@ -20,6 +20,13 @@ export const COMPARISON_TOOL_BODY_CLASS = 'comparison-tool-body';
...
+export const CLEAR_ALL_PINS_TOOLTIP = 'Clear all pinned results';
```
hallieswan: Nit: could consider grouping these into one object, so the five named imports and pass through class fields in comparison-tool-table.component.ts each collapse to one. Up to you!

hallieswan (commented):
Looks great! Just a couple small questions. 

I also noticed that we only added a download pins note for Model AD CTs. Are we planning to add a similar note for Agora and QTL CTs in the future? If so, should we track in a follow-up ticket?

sagely1:
> Looks great! Just a couple small questions.
> 
> I also noticed that we only added a download pins note for Model AD CTs. Are we planning to add a similar note for Agora and QTL CTs in the future? If so, should we track in a follow-up ticket?

I've confirmed with @JessterB we will not be needing a download pins note for Agora and QTL for now.  The placeholders will be blank for those apps but if they are needed in the future, a followup Jira issue will be created for those products in the future.

libs/model-ad/config/src/lib/constants.ts
sagely1: Good point - the CT is going to now just go to the current HELP_URL (with no anchor).  In a future PR an anchor will be added to a new more generic section.

hallieswan (approved):
Looks good to me!

